PR #238, opened 2012-10-12
Implemented channel sorting + stable join order

### LINKED ISSUE #104: Allow ordering channels [$15]
A few people have requested/complained that ZNC doesn't order channels how they wish.

Previously ZNC written its configs differently so that channels were saved in order they were added, I changed it to use CConfig to write the config. This doesn't store them in the same order.

We could implement some kind of index on the channels in the config. This will allow us to read them and place them in the same order as before.

```
/msg *status movechan #channel <new-index>
```

The webadmin could also be updated to allow moving channels up/down.

I have created an issue so we can discuss this further and decide if we should implement this, and how.

Related: #89
## 

<bountysource-plugin>
---
There is a **[$15 open bounty](https://www.bountysource.com/issues/696-allow-ordering-channels?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sebageek:
When connecting to znc the join is unpredictable (or in the same way it was when znc was started the first time). With this patch a sort order can be defined. This is an numeric attribute attached to each channel. On AddChannel() SortChannels() is called. This also happens when setting a SortOrder via the *status bot thingy.

Fix #104

### DIFF
--- a/include/znc/Chan.h
+++ b/include/znc/Chan.h
@@ -48,8 +48,12 @@ public:
 		M_Except     = 'e'
 	} EModes;
 
+	static const unsigned int    m_uDefaultSortOrder;
+
 	CChan(const CString& sName, CIRCNetwork* pNetwork, bool bInConfig, CConfig *pConfig = NULL);
 	~CChan();
+
+	bool operator<(const CChan& other) const;
 
 	void Reset();
 	CConfig ToConfig();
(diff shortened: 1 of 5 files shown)


Mikaela:
It seems that there are commits here after the last comment (GitHub doesn't notify when commits are added to PR, you must comment after that so repo owners get hte information) and the PR has merge conflict.

DarthGandalf:
No new commits are added to it... They are cherry-picks of this commit to
other forks of ZNC.

2016-02-02 10:31 GMT+00:00 Mikaela Suomalainen <notifications@github.com>:

luke-jr:
Surprised this isn't merged yet. Sure, it's not perfect, but ZNC is basically unusable without it IMO...

Rebased on 1.7.1 here:

https://bpaste.net/show/8a0623b000ba
https://bpaste.net/show/79c78abb0d07

PR #257, opened 2012-11-21
systemWideConfig implementation

wired:
This patch adds a systemWideConfig mode to znc.

-S user
	enables systemWideConfig
	switches to [user] if znc was run as root

long version of -S: --system-wide-config-as

When enabled, if running as root, znc will drop privileges to the
specified user.

WriteNewConfig has also been modified to skip the "Launch ZNC now" step
when systemWideConfig is enabled.

Per my initial talk with psychon, the new parameter is not documented
in -h for now.

The patch was improved with the help of DGandalf

### DIFF
--- a/include/znc/znc.h
+++ b/include/znc/znc.h
@@ -169,6 +169,8 @@ public:
 
 	static void DumpConfig(const CConfig* Config);
 
+	void SetSystemWideConfig(bool systemWideConfig);
+
 private:
 	CFile* InitPidFile();
 	bool DoRehash(CString& sError);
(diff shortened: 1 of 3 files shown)


DarthGandalf:
On the other hand, switching user is better done externally, e.g. with start-stop-daemon

DarthGandalf:
9688727f3262c15dee0016c33c8dbf18bb4feb25 allows to skip the "Launch ZNC now" step.
This part obviously needs to be in ZNC, as it couldn't be disabled without changing source.

PR #325, opened 2013-04-20
Web interface for the log module

pepijndevos:
I thought it'd be cool to be able to view logs in the web interface.

It took me forever to write, since it's my first piece of C++.


DarthGandalf:
Nothing happened with this in 2 years...

PR #336, opened 2013-04-29
User registration module

pepijndevos:
This module allows users to register themselves. This plugin is aimed at (commercial) ZNC providers, not at individuals.

Besides simple registration, it contains two non-obvious features that are very important to me and maybe others.

First of all, it takes an optional argument. If this argument is provided, it is used as a salt to hash the username. It will also reveal a verification code field. The use case for this is that you can have your paywall spit out an url like `myawesomeznc.com/mods/global/register/?user=pepijn&code=deadbeef`

The second is that there is an `#ifdef` section that will add a default network if you configure with `CXXFLAGS="-DREGISTER_HOST=irc.example.com" ./configure`. This is because virtually all my users will connect to the same server and some people have trouble adding a network.

Of course, the module can also be used without any of those options, just allowing simple registrations.

### DIFF
--- a/modules/data/register/tmpl/index.tmpl
+++ b/modules/data/register/tmpl/index.tmpl
@@ -10,10 +10,8 @@
 			<div class="sectionbody">
 				<div class="subsection">
 					<div class="inputlabel">Username:</div>
-
-						<input type="text" name="user" value="<? VAR Username ?>" class="half" maxlength="128"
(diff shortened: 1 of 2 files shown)


ssbarnea:
Was there any reason for closing this without a merge? Being able to allow users to register seems like a very useful feature.

teward:
@ssbarnea Probably because between the last comment and the time of the closure almost 2 years went by with no updates or changes here.

PR #395, opened 2013-08-13
Fake server notice

Xe:
Prefer server notices in the form of

```
:*status.znc.in NOTICE * :*** &rest
```


DarthGandalf:
4. Let's have what's written to socket and what's written in debug output the same

PR #503, opened 2014-03-02
Allow znc-only users to join partyline channels

lpsmith:
As it stands,  users that are not connected to an external network cannot join partyline channels.   Here's how I fixed it on my own bouncer.

I openly admit that this patch is a kludge,  but the approach taken didn't require me to delve deeply into the ZNC source.   There are certainly better ways to fix this,  so I'm not particularly expecting this pull request to be accepted,  but rather I'm offering it as documentation of the issue.

### DIFF
--- a/modules/partyline.cpp
+++ b/modules/partyline.cpp
@@ -299,6 +299,28 @@ public:
 				m_pUser->PutUser(":irc.znc.in 442 " + m_pClient->GetNick() + " " + sChannel + " :You're not on that channel");
 			}
 			return HALT;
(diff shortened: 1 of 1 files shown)


DarthGandalf:
@jpnurmi #1024 supersedes this, right?

jpnurmi:
There's some overlap. I like the idea of changing ```CClient::ReadLine()``` to call those hooks even without a network, but I wonder if we might break modules with that change.

jpnurmi:
Superseded by #1059

PR #531, opened 2014-04-14
modules/log: Don't make a new logfile for every day

Xe:
Having a new logfile for every day can cause problems with directory traversal and it does not scale.


DarthGandalf:
And such timestamp format setting was pushed (#978)

Issue #547, opened 2014-05-12
PRIVMSGs sent to prefixed channels by a client are not broadcast to all other clients as ordinary PRIVMSGs are


TheWug:
Many modern ircds (freenode included) will allow PRIVMSGs of the following format:

```
PRIVMSG @#channel :This is an ops-only message.
```

ZNC will not echo PRIVMSGs of this format to all connected clients when one client sends one, as it will with other PRIVMSGs sent to channels.

I didn't actually try it, but if it were happening it would happen here: https://github.com/znc/znc/blob/master/src/Client.cpp#L357
And I don't see anything there that would handle it.

The network sends its list of prefix symbols in the ISUPPORT (005) numeric upon connecting, so basically the chance you'd be looking at is:

```
CChan* pChan = m_pNetwork->FindChan(sTarget);
if (pChan == NULL && GetIRCSock()->GetPerms().find(sTarget[0]) != CString::npos)
{
    pChan = m_pNetwork->FindChan(sTarget.substr(1));
}
```

Or, you could modify _CIRCNetwork::IsChan_ to return true for valid channel names prepended with a valid channel mode prefix.  That wouldn't change the fact that prefixed channel messages wouldn't end up in the scrollback though.

Maybe I'll patch this myself.  Who knows.

Mikaela:
I think this doesn't happen anymore and should be closed.

jpnurmi:
ZNC 1.7.x-git-377-3d95aea - http://znc.in
```
CLI -> ZNC [PRIVMSG @#znc-test :hello ops]
ZNC -> CLI [@time=... :jpnurmi!jpnurmi@qt/jpnurmi PRIVMSG @#znc-test :hello ops]
ZNC -> CLI [@time=... :jpnurmi!jpnurmi@qt/jpnurmi PRIVMSG @#znc-test :hello ops]
ZNC -> IRC [PRIVMSG @#znc-test :hello ops]
```

PR #548, opened 2014-05-12
Fixed https://github.com/znc/znc/issues/547

TheWug:
Fix issue 547, wherein a prefixed channel message would not be propagated to other clients.  If m_pNetwork->FindChan found a channel, then pChan is a channel (or a prefixed channel, since FindChan accounts for that) so we don't need the redundant IsChan check.

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -264,7 +264,7 @@ void CClient::ReadLine(const CString& sData) {
 			}
 
 			// Relay to the rest of the clients that may be connected to this user
-			if (m_pNetwork->IsChan(sTarget)) {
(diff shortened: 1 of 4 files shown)


src/Client.cpp
```diff
@@ -264,7 +264,7 @@ void CClient::ReadLine(const CString& sData) {
 			}
 
 			// Relay to the rest of the clients that may be connected to this user
-			if (m_pNetwork->IsChan(sTarget)) {
+			if (pChan != NULL || m_pNetwork->IsChan(sTarget)) {
```
DarthGandalf: With modified `IsChan()`, what's the point of checking for NULL?

DarthGandalf:
#547 is fixed already

PR #603, opened 2014-07-25
self_message module.

NuclearW:
Add support for the self-message CAP as defined by https://github.com/ircv3/ircv3-specifications/blob/master/extensions/self-message-3.2.md

At the moment znc already sends messages to different clients as expected by self-message for channels, this module adds the same for direct messages.

Modified a module better suited as an external module to make this.  Though I'm not sure if the features here would be better suited to being built into znc itself, rather than as a module.


jpnurmi:
Superceded by echo-message: #950

PR #620, opened 2014-08-10
Add support for --quiet argument in client commands

jpnurmi:
This allows clients to manage buffers etc. by sending commands to *status (eg. clear the buffer when a query is closed) without undesired responses (that would re-open the query).

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -94,6 +94,7 @@ public:
 		m_bUHNames = false;
 		m_bAway = false;
 		m_bServerTime = false;
+		m_bQuiet = false;
 		EnableReadLine();
 		// RFC says a line can have 512 chars max, but we are
 		// a little more gentle ;)
@@ -128,6 +129,7 @@ public:
 	void PutStatusNotice(const CString& sLine);
 	void PutModule(const CString& sModule, const CString& sLine);
 	void PutModNotice(const CString& sModule, const CString& sLine);
+	void PutCommandInfo(const CString& sInfo);
 
 	bool IsCapEnabled(const CString& sCap) { return 1 == m_ssAcceptedCaps.count(sCap); }
 
@@ -153,6 +155,7 @@ public:
 private:
 	void HandleCap(const CString& sLine);
 	void RespondCap(const CString& sResponse);
+	void HandleCommandArgs(CString& sLine);
 
 protected:
 	bool                 m_bGotPass;
@@ -163,6 +166,7 @@ protected:
 	bool                 m_bUHNames;
 	bool                 m_bAway;
 	bool                 m_bServerTime;
+	bool                 m_bQuiet;
 	CUser*               m_pUser;
 	CIRCNetwork*         m_pNetwork;
 	CString              m_sNick;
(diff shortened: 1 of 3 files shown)


OGAWAHirofumi:
Adding alternative command does not make sense, IMO. Really want to add alternative commands to modules only for suppress "success" message?

If so, I'd like to add another one to clearbufferonmsg, because I don't want to compile external module.

jpnurmi:
That particular command makes sense for the playback module that manages buffers, but in general yes, I'm still supporting the idea of adding this functionality to ZNC core in a way or another. We just didn't manage to find a satisfying solution yet, so I put it on hold for now... :)

OGAWAHirofumi:
Ok, one idea. How about to add new command "ClearBuffer2" to core?

This share code with "ClearBuffer", but suppress feedback.

jpnurmi:
The problem with --quiet, -silent, or even clearbuffer2 is that it's tricky for clients to start using them. A clear winner would be a solution that doesn't choke older ZNC versions.

OGAWAHirofumi:
Hm, how do clearbuffer2 choke older znc version?

I meant to support both of
    /ZNC clearbuffer
and 
    /ZNC clearbuffer2
commands in new version. So, old ZNC user can continue use "clearbuffer" without any change.

jpnurmi:
I mean, when can clients start using clearbuffer2? Anyone that runs non-bleeding-edge ZNC gets

    Unknown command [clearbuffer2] try 'Help'

instead of a silent clear as intended. I can't think of any backwards compatible way to provide this functionality, but I'm hoping someone comes up with a smart idea. :)

OGAWAHirofumi:
Client can see help and parse is not enough?

jpnurmi:
I thought about having ZNC status prefix and version injected to RPL_ISUPPORT (#660), but even that's a bit too late for proper interaction with modules. Clients consider themselves fully connected and typically start sending initial commands already after receiving RPL_WELCOME.

OGAWAHirofumi:
Well, IMO, I don't see big difference with using module command. On both cases, client don't know if the command is available or not. So, I'm not sure why we need special care in new znc command case.

jpnurmi:
Maybe I'm over complicating things, but overloading commands just doesn't feel like the right solution. As long as there's no nice way for clients to even know the ZNC version, there's no nice way to take such new commands in use without breaking compatibility with older ZNC versions. Maybe it should be something totally different... for example, request a capability to open a special client<->ZNC communication channel/protocol so clients can also interact with modules without having to know the statusprefix.

OGAWAHirofumi:
I'm still not need to special care for this though. (because no any breakages until client use new commands, even doesn't need to care there are new commands). But clients would want to know znc has what capabilities.

Well, so, I'm not checking RFCs though. It is wrong to add capability string of IRC to tell version like "znc.in/server-time-iso"? For example, tell "znc.in/version-1.6.0" (If clients requested this version capability, znc just ack).

jpnurmi:
IRC v3.2 introduces [capability values](https://github.com/ircv3/ircv3-specifications/blob/master/specification/capability-negotiation-3.2.md), so something like "znc.in/version=1.6.0" becomes possible indeed.

PR #644, opened 2014-08-28
Better buffer playback notification

renatosilva:
Send buffer playback notification as a NOTICE instead of PRIVMSG, so IRC clients lacking the server-time capability can avoid regarding it as a normal user message (and thus highlight it differently for example).


jpnurmi:
Clients can request a ```batch``` capability to get buffer playback in ```znc.in/playback``` batches (https://github.com/znc/znc/commit/73b6936d6d7422eb17d957c90ae5f9ce1ffc0856).

PR #647, opened 2014-09-01
znc.service: Add support for reloading

Lyude:
ZNC has the ability to reread it's configuration when it receives a
SIGHUP, so we can use this to allow reloading of znc's configuration
files from systemctl

### DIFF
--- a/znc.service
+++ b/znc.service
@@ -4,6 +4,7 @@ After=network.target
 
 [Service]
 ExecStart=/usr/bin/znc -f
+ExecReload=/bin/kill -HUP $MAINPID
 User=znc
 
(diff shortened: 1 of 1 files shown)


md-5:
If a service file is gonna be published it may as well be as complete as it can be....

Mikaela:
I would agree with @md-5 and I took that to my version of the service.

* https://github.com/Mikaela/shell-things/blob/master/etc/systemd/system/znc.service

It could probably be changed to user service, but I am not familiar with how those are supposed to work yet as they seem to be broken on fresh Ubuntu Server where systemd has been installed.

PR #685, opened 2014-09-28
Add OnUser*Tags hooks for modules.

NuclearW:
This will allow modules to use new hooks to get access to any tags sent with messages from a client to znc.

At present there is no additional behavior or actions taken by znc when message tags are present.

This pull isn't intended to be ready right now (though if it builds I suppose it could be, I've built it sans python/perl).  It will need to be discussed, tested, and then if it's something that's wanted I'll go ahead and squash it down into one or two commits.


jpnurmi:
Superseded by #1031

PR #695, opened 2014-09-30
log: Add -timestamp option

NuclearW:
Allows users to define a new timestamp as formatted by strftime through CUtils::FormatTime
Additionally `-sanitize` is no longer necessarily the first argument, but `-timestamp` and the actual timestamp must be last as it can contain spaces.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -65,6 +65,7 @@ public:
 
 private:
 	CString                 m_sLogPath;
+	CString                 m_sTimestamp;
 	bool                    m_bSanitize;
 };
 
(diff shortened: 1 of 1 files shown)


markusj:
What needs to be done to get this PR merged? I have implemented some [simpler workaround](https://github.com/markusj/znc/tree/log-with-user-time-format), but in my opinion this approach is the better one.

DarthGandalf:
@markusj 
"Everything after this must be timestamp" needs to be fixed (I commented inline)
"need better parsing for parameters in modules" is meta comment, it's fine...
Also it has some merge conflicts.

anlutro:
Anything being done with this? I could try and finish @NuclearW's work if no one more qualified wants to (my C++ is pretty shabby).

DarthGandalf:
Pushed at #978

PR #791, opened 2014-12-26
Change namespace for <fcntl.h>

James-TR:
* Fixes warnings under musl-libc.
* <fcntl.h> should be universal in the Unix world, it's more likely a system doesn't have <sys/fcntl.h> than if it doesn't have <fcntl.h> (and has <sys/fcntl.h>.)


James-TR:
I've updated the commit to take into account that CSocket is external, and whilst I was at it, rebased onto master

Issue #746, opened 2014-11-11
Add ecdsa-nist256p-challenge into sasl. [$10]


KindOne-:
Since ```dh-blowfish``` and ```dh-aes``` have been killed off, might as well add support for ```ecdsa-nist256p-challenge```.

https://github.com/atheme/ecdsatool

https://github.com/atheme/atheme/blob/master/modules/saslserv/ecdsa-nist256p-challenge.c

Mikaela:
WeeChat git version got support for it last night https://github.com/weechat/weechat/commit/e2be01833f2593b90c4059f712a9a2e2cf01ea43 (quick instructions: https://github.com/weechat/weechat/issues/251#issuecomment-70576923) and Limnoria merged the support to stable version some time ago.

Mikaela:
I think it was decided that this isn't going to happen because of the log quoted below, but I am unable to find source for it at GitHub.

```
[11:28:05] <kaniini> but ECDSA-CHALLENGE is probably dead anyway
[11:28:20] <kaniini> because in mammon we're going to implement a version which does not use NIST curves
[11:28:37] <kaniini> because they appear to be malleable: http://safecurves.cr.yp.to/rigid.html
[11:31:43] <SaberUK> nist standardising things which are dodgy? colour me surprised! ;)
[11:36:23] <kaniini> i doubt they knew
[11:42:25] <sheep> is it going to use Curve25519/Ed25519?
[11:43:04] <kaniini> probably
```

From \#ircv3 on 2015-04-04.

PR #793, opened 2014-12-26
modules/sasl.cpp: update help for "set"

Mikaela:
Preparing for #746 (ECDSA-NIST256P-CHALLENGE) which requires username.
It's also not entirely impossible that in the future there are other
mechanisms that want username and/or password so it doesn't hurt to have
help text open for them.

### DIFF
--- a/modules/sasl.cpp
+++ b/modules/sasl.cpp
@@ -70,7 +70,7 @@ public:
 		AddCommand("Help",        static_cast<CModCommand::ModCmdFunc>(&CSASLMod::PrintHelp),
 			"search", "Generate this output");
 		AddCommand("Set",         static_cast<CModCommand::ModCmdFunc>(&CSASLMod::Set),
-			"username password", "Set username and password for the PLAIN mechanism");
+			"<username> [<password>]", "Set username and password for the mechanisms that need them. You can also omit the password to only set username");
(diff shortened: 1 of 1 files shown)


modules/sasl.cpp
```diff
@@ -70,7 +70,7 @@ class CSASLMod : public CModule {
 		AddCommand("Help",        static_cast<CModCommand::ModCmdFunc>(&CSASLMod::PrintHelp),
 			"search", "Generate this output");
 		AddCommand("Set",         static_cast<CModCommand::ModCmdFunc>(&CSASLMod::Set),
-			"username password", "Set username and password for the PLAIN mechanism");
+			"username password", "Set username and password for the mechanisms that need them. You can also omit the password to only set username");
```
Mikaela: @DarthGandalf Is this any better?

DarthGandalf: `<username> [<password>]`

Mikaela: Done. Should anything else be changed?

jpnurmi:
Looks like pull requests don't close automatically with "closes/fixes #NNN".

Mikaela:
> Looks like pull requests don't close automatically with "closes/fixes #NNN".

They do as long as the commit is merged to default branch or in this case *master*.

jpnurmi:
Ah, makes sense. Thanks.

PR #797, opened 2015-01-02
Select the TLS ciphers in server-preferred order

kerio92:


kylef:
Changes to Csocket need to go  https://github.com/jimloco/Csocket

PR #799, opened 2015-01-03
Move Csocket to git submodule

### LINKED ISSUE #580: Make Csocket as a submodule in git
Current "Csocket merges" are dumb and doesn't provide any meaningful Csocket changelog. Moreover it breaks Github links.

For example: https://github.com/znc/znc/commit/6012cdee77c4e11879537c973177d138b9e66006
Good luck finding exact change that added charset support here. Also link in commit message is broken.

Yes, I can go to Csocket repository and check there, but that's what submodules actually do.

DarthGandalf:
Fixes #580


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "Csocket"]
+	path = third_party/Csocket
+	url = https://github.com/jimloco/Csocket
```
Mikaela: I am not familiar with submodules, but shouldn't this be `https://github.com/jimloco/Csocket.git`? GitHub works even without the .git, but it might be confusing to people who are used to .git always being there.

DarthGandalf:
For reference: df9e909664904a18ed2a774e0a1fbdf0bdef6327 was previous version of this PR

.gitmodules
DarthGandalf: Done

kylef:
:+1: finally. I remember taking a stab at this and failing a while back. Nice job :+1:.

jpnurmi:
Yeah, a good step forward. :) After the release is out, a future improvement is to eliminate all csocket.h includes from the znc headers, which in turn eliminates the need of copying csocket sources around. Use aggregation instead of inheritance, and forward virtual calls where necessary.

PS. +1 for src/3rdparty for aesthetic reasons.

DarthGandalf:
@jpnurmi what's up with src/3rdparty? third_party is not aesthetic?

Mikaela:
> what's up with src/3rdparty? third_party is not aesthetic?

3rdparty looks more aesthetic to me

DarthGandalf:
@Mikaela We use third_party at work...

jpnurmi:
And we use 3rdparty at work... :)

DarthGandalf:
https://github.com/search?utf8=%E2%9C%93&q=third_party shows more results than https://github.com/search?utf8=%E2%9C%93&q=3rdparty :P

kylef:
We use `vendor` at work, it has over 4 times the results ;)

- https://github.com/search?utf8=✓&q=vendor&type=Repositories&ref=searchresults

PR #803, opened 2015-01-04
savebuff: save & load query buffers - fixes #802

### LINKED ISSUE #802: savebuff: query buffers are not saved and loaded
See @xor-freenet's comments in https://github.com/znc/znc/commit/287a020d21ec5f4798d3df845826e8c103a1225d. Despite the blind changes applied in the commit, query buffers are not actually saved and loaded at all.

- OnIRCConnected() incorrectly assumes that queries would already exist
- SaveBuffersToDisk() is missing handling of queries altogether

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7471264-savebuff-query-buffers-are-not-saved-and-loaded?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:
The old internal storage format was:

    ::__:SAVEBUFF:__::<buffer lines>

The new storage format for channels:

    ::__:CHANBUFF:__::<buffer name>
    <buffer lines>

The new storage format for queries:

    ::__:QUERYBUFF:__::<buffer name>
    <buffer lines>

See #802 for details why the storage format had to be changed. Reading
the old storage format is still supported so old saved buffers load as
appropriate.

### DIFF
--- a/modules/savebuff.cpp
+++ b/modules/savebuff.cpp
@@ -30,9 +30,12 @@
 #include <znc/FileUtils.h>
 #include <znc/Query.h>
 
+using std::set;
 using std::vector;
 
-#define CRYPT_VERIFICATION_TOKEN "::__:SAVEBUFF:__::"
+#define LEGACY_VERIFICATION_TOKEN "::__:SAVEBUFF:__::"
+#define CHAN_VERIFICATION_TOKEN "::__:CHANBUFF:__::"
+#define QUERY_VERIFICATION_TOKEN "::__:QUERYBUFF:__::"
 // this is basically plain text, but so is having the pass in the command line so *shrug*
 // you could at least do something kind of cool like a bunch of unprintable text
 #define CRYPT_LAME_PASS "::__:NOPASS:__::"
@@ -58,7 +61,6 @@ public:
 	MODCONSTRUCTOR(CSaveBuff)
 	{
 		m_bBootError = false;
-		m_bFirstLoad = false;
 
 		AddHelpCommand();
 		AddCommand("SetPass", static_cast<CModCommand::ModCmdFunc>(&CSaveBuff::OnSetPassCommand), "<password>", "Sets the password");
@@ -69,7 +71,7 @@ public:
 	{
 		if (!m_bBootError)
 		{
-			SaveBufferToDisk();
+			SaveBuffersToDisk();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


modules/savebuff.cpp
```diff
@@ -102,107 +104,116 @@ class CSaveBuff : public CModule
...
+					continue;
+
+				CString sName;
+				CString sBuffer;
+				EBufferType eType = DecryptBuffer(pFile->GetLongName(), sBuffer, sName);
+
+				switch (eType) {
+				case InvalidBuffer:
+					m_sPassword = "";
+					CUtils::PrintError("[" + GetModName() + ".so] Failed to Decrypt [" + pFile->GetShortName() + "]");
```
DarthGandalf: Let's use absolute name... Which user saved buffer into file "764efa883dda1e11db47671c4a3bbd9e"?

jpnurmi: Good point, fixed

jpnurmi:
When AutoClearXxxBuffer=true, something goes wrong if a client connects to ZNC meanwhile its connecting to IRC?

jpnurmi:
CIRCNetwork::ParseConfig() loads modules first, then adds configured channels. This is why savebuff loads buffers in OnIRCConnected() instead of OnLoad(). That's too late if a client connected before ZNC got connected to IRC. OnClientLogin() is also too late, because the initial playback has already been sent by then... ideas?

jpnurmi:
OnUnknownUserRaw() would get called early enough, but it's not too reliable because it can be blocked by any other module. OnClientConnect() on the other hand is not usable for savebuff, because it's a global hook.

jpnurmi:
For what it's worth, the problem is not introduced by this commit. The savebuff module works only if you wait until ZNC connects to IRC. If a client connects before that, you won't get the playback from savebuff.

DarthGandalf:
What about OnBoot()?

DarthGandalf:
And it's not related to order in which what connects to what.
Files are simply not removed from disk, and will be read again next time ZNC starts.

jpnurmi:
> And it's not related to order in which what connects to what.

If ZNC is not connected to IRC, channels are not "on", and no playback is sent. Sometime later when ZNC connects IRC, it won't send buffer playback to "newly joined channels".

DarthGandalf:
Yes... So it's a separate issue with similar sympthoms

jpnurmi:
OnBoot() seems better. I had to tweak CQuery::SendBuffer() to use CClient::GetNick() instead of CNetwork::GetIRCNick() since the IRC nick is not yet known.

DarthGandalf:
Oops, sorry, I haven't realized that you updated PR...

modules/savebuff.cpp
```diff
@@ -94,115 +95,119 @@ class CSaveBuff : public CModule
...
-				for (unsigned int uIdx = 0; uIdx < uSize; uIdx++) {
-					const CBufLine& Line = Buffer.GetBufLine(uIdx);
-					timeval ts = Line.GetTime();
-					sFile +=
-						"@" + CString(ts.tv_sec) + "," + CString(ts.tv_usec) + " " +
-						Line.GetFormat() + "\n" +
-						Line.GetText() + "\n";
-				}
+			// cleanup auto-cleared queries
+			CDir::Delete("*", GetSavePath());
```
DarthGandalf: What if ZNC crashes right after this? I think it's better to delete unused files after writing new files.

modules/savebuff.cpp
```diff
@@ -94,115 +95,119 @@ class CSaveBuff : public CModule
...
-				size_t uSize = Buffer.Size();
-				for (unsigned int uIdx = 0; uIdx < uSize; uIdx++) {
-					const CBufLine& Line = Buffer.GetBufLine(uIdx);
-					timeval ts = Line.GetTime();
-					sFile +=
-						"@" + CString(ts.tv_sec) + "," + CString(ts.tv_usec) + " " +
-						Line.GetFormat() + "\n" +
-						Line.GetText() + "\n";
-				}
+			// cleanup auto-cleared queries
```
DarthGandalf: s/queries/buffers/

modules/savebuff.cpp
jpnurmi: Done

modules/savebuff.cpp
jpnurmi: Done

Issue #602, opened 2014-07-24
Q-Module: Clients dont reconnect after restart of znc


fred0r:
The q-module seems broken. when enabled the quakenet-clients dont rejoin the added chans. i disabled the q-module and reconnect was fine.

PR #812, opened 2015-01-11
Q: Ignore JoinAfterCloaked unless UseCloakedHost is set

ahf:
This patch ignores the value of JoinAfterCloaked unless UseCloakedHost
is set. Without this patch, it's possible to enable JoinAfterCloaked,
but not have UseCloakedHost enabled, which means that no channel will
ever be joined automatically by ZNC.

Fixes: #602


fred0r:
thx

PR #813, opened 2015-01-12
Expand crypt module to cover Notices, Actions & Topics

anthonyryan1:
Improves compatibility with mircryption.

Notes:
1. I know the use of `OnRaw` is not suggested, but I cannot find a more appropriate way to intercept the incoming topic message.
2. There's some duplication here, but when I added functions for the individual pieces of functionality needed to handle outgoing messages, it did not seem to significantly improve readability.

I'm open to all comments and suggestions.

### DIFF
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -18,6 +18,9 @@
 //
 // The encryption here was designed to be compatible with mircryption's CBC mode.
 //
+// Latest tested against:
+// MircryptionSuite - Mircryption ver 1.19.01 (dll v1.15.01 , mirc v7.32) CBC loaded and ready.
+//
 // TODO:
 //
 // 1) Encrypt key storage file
@@ -54,10 +57,9 @@ public:
 	virtual ~CCryptMod() {}
 
 	EModRet OnUserMsg(CString& sTarget, CString& sMessage) override {
-		sTarget.TrimLeft(NickPrefix());
+		sTarget.TrimPrefix(NickPrefix());
 
-		if (sMessage.Left(2) == "``") {
-			sMessage.LeftChomp(2);
+		if (sMessage.TrimPrefix("``")) {
 			return CONTINUE;
 		}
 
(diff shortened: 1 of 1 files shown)


DarthGandalf:
`OnTopic`?

modules/crypt.cpp
```diff
@@ -84,18 +87,144 @@ class CCryptMod : public CModule {
...
+			return HALTCORE;
+		}
+
+		return CONTINUE;
+	}
+
+	virtual EModRet OnUserAction(CString& sTarget, CString& sMessage) override {
+		sTarget.TrimLeft(NickPrefix());
+
+		if (sMessage.StartsWith("``")) {
```
DarthGandalf: if (s.TrimPrefix(..))

DarthGandalf:
Looks like you're not open to comments :)

anthonyryan1:
With all due respect, your comment in this issue wasn't exactly meaningful. What about OnTopic?

The other part, was admittedly my fault. I didn't recall seeing your comment about switching the `sTarget.TrimLeft(NickPrefix())` with `sTarget.TrimPrefix()`. If you can elaborate on what you mean I can ensure both of your requests are incorporated.

DarthGandalf:
OnTopic was response for "but I cannot find a more appropriate way to intercept the incoming topic message."

anthonyryan1:
At least when I tested, the raw 332 was not corrected by OnTopic alone. I can chekc if that has changed since my prior testing. `OnTopic` is  used, but without the `OnRaw` it wasn't working fully.

Perhaps a better solution to that would be to ensure that OnTopic does intercept the 332s as well as what it does currently.

DarthGandalf:
Oh, I didn't notice `sTarget.TrimLeft` before, thanks for fixing it. `TrimLeft` does something completely different. Speaking about `sTarget.TrimLeft`, there is a similar one added below in this PR.
My comment was about `sMessage.StartsWith("``")`.

Regarding `OnTopic`, I see... OnTopic handles cases when it is changed, 332 is for case if client requests `/topic` and joins channel.

May you rebase/squash commits?

anthonyryan1:
I hope to get this rebased tomorrow, sorry about the delay.

anthonyryan1:
Rebased & tested.

Let me know if you have any comments of concerns.

modules/crypt.cpp
```diff
@@ -84,22 +86,144 @@ class CCryptMod : public CModule {
 		return CONTINUE;
 	}
 
+	EModRet OnUserNotice(CString& sTarget, CString& sMessage) override {
+		sTarget.TrimLeft(NickPrefix());
```
DarthGandalf: TrimLeft or TrimPrefix?

anthonyryan1:
It should have indeed been TrimPrefix. Somehow that change regressed while rebasing.

Issue #816, opened 2015-01-17
Replace some C++98isms with C++11isms


md-5:
Notably we can now use range based for loops, eg:
```c++
            int count = 0;
            for (auto network : m_pUser->GetNetworks()) {
            	for (auto chan : network->GetChans()) {
            		chan->ClearBuffer();
            		count++;
            	}
            }
```
as opposed to the old style:
```c++
const vector<CIRCNetwork*>& vNetworks = m_pUser->GetNetworks();
            vector<CIRCNetwork*>::const_iterator networkIter;
            for (networkIter = vNetworks.begin(); networkIter != vNetworks.end(); ++networkIter) {
                const vector<CChan*>& vChans = (*networkIter)->GetChans();
                vector<CChan*>::const_iterator chanIter;
                for (chanIter = vChans.begin(); chanIter != vChans.end(); ++chanIter) {
                    (*chanIter)->ClearBuffer();
                    count++;
                }
            }
```

jpnurmi:
We've got a lot of C++11 related changes merged. Anything else on the wish list?
- range-based for loops
- override
- nullptr

Loops were changed by hand, override & nullptr were added mostly with help of clang-modernize. Additionally, some auto's could be sprinkled here and there, but I think I'm not going to start systematically doing that. We can do such tweaks when we touch the respective code. Case closed for now?

md-5:
Looks nice. Thanks!

PR #817, opened 2015-01-17
Work on issue #816 - modernize some for loops.

md-5:


src/Client.cpp
```diff
@@ -199,13 +199,12 @@ void CClient::ReadLine(const CString& sData) {
 		return;                       // Don't forward this msg.  We don't want the client getting us disconnected.
 	} else if (sCommand.Equals("PROTOCTL")) {
 		VCString vsTokens;
-		VCString::const_iterator it;
 		sLine.Token(1, true).Split(" ", vsTokens, false);
 
-		for (it = vsTokens.begin(); it != vsTokens.end(); ++it) {
-			if (*it == "NAMESX") {
+		for (auto it : vsTokens) {
```
DarthGandalf: This makes a copy of the string. Please default to `const auto&`.
If it needs to be changed, and not in copy, `auto&` will do.

DarthGandalf: Btw, "it" may have a better name now, as it's not an iterator anymore ;)

src/Client.cpp
```diff
@@ -402,7 +401,7 @@ void CClient::ReadLine(const CString& sData) {
 		sPatterns.Split(" ", vsChans, false, "", "", true, true);
 
 		set<CChan*> sChans;
-		for (const CString& sChan : vsChans) {
+		for (auto sChan : vsChans) {
```
DarthGandalf: `for (const CString& sChan : vsChans)` was not modern enough? And it does not make a copy

src/ClientCommand.cpp
```diff
@@ -635,20 +635,19 @@ void CClient::UserCommand(CString& sLine) {
 			return;
 		}
 
-		const CModules& vMods = pOldNetwork->GetModules();
-		for (CModules::const_iterator i = vMods.begin(); i != vMods.end(); ++i) {
-			CString sOldModPath = pOldNetwork->GetNetworkPath() + "/moddata/" + (*i)->GetModName();
-			CString sNewModPath = pNewUser->GetUserPath() + "/networks/" + sNewNetwork + "/moddata/" + (*i)->GetModName();
+		for (auto pMod : pOldNetwork->GetModules()) {
```
DarthGandalf: You can use `auto*` or `const auto*` to make it clear that this is a pointer

DarthGandalf: Or just use `CModule*` :)

jpnurmi: +1 for CModule*, overusing auto reduces readability

md-5: @jpnurmi my reasoning was that in a loop it should be really clear what the value is.

DarthGandalf: How CModule* makes it less clear?

jpnurmi: @md5 Things might seem clear to you at the time you refactor some piece of code, but think about people who have to read, maintain and browse the code later. The auto keyword is really neat when used as a replacement for clumsy long template/iterator monster types, but IMHO it shouldn't be used as the default type for every single for-loop, because it makes things ambiguous and less readable.

md5: Wrong @md5! You meant @md-5 :)

DarthGandalf:
#892, #899 supersede this.

PR #823, opened 2015-01-24
Tell user to use PASS <username>/<network> instead of just the username.

Zarthus:
In versions prior to 1.0, ZNC did not allow multiple networks per user. The string telling users to connect has never been changed after ZNC 1.0, and this fixes just that.

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -82,7 +82,10 @@ void CClient::SendRequiredPasswordNotice() {
 	PutClient(":irc.znc.in 464 " + GetNick() + " :Password required");
 	PutClient(":irc.znc.in NOTICE AUTH :*** "
 			  "You need to send your password. "
-			  "Try /quote PASS <username>/<network>:<password>");
+			  "Configure your client to send a server password.");
+	PutClient(":irc.znc.in NOTICE AUTH :*** "
+			  "To connect now, you can use /quote PASS <username>:<password>, "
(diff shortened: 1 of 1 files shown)


DarthGandalf:
`<user>:<pass>` still works. It uses network named "default" by default, if it exists. If such network doesn't exist, it uses just some network.

`<user>/<net>:<pass>` has more variables to substitute == more things for user to do wrong :(

Mikaela:
How about just saying `Set password as <username>/<network>:<password> in your IRC client and reconnect` and forgetting that manual `/pass`?

Zarthus:
We could do both, changing it into two messages.

```
You need to send your password. Configure your client to send a server password.
To connect now, you can use /quote PASS <username>:<password>, or /quote PASS <username>/<network>:<password> to connect to a specific network.
```

md-5:
> it uses just some network.

Is that deterministic or....?

Mikaela:
:+1: for @Zarthus's both messages.

DarthGandalf:
@md-5 https://github.com/znc/znc/blob/600f91f128802ea6225d457b7247aba4a4119362/src/Client.cpp#L690

Cajs:
Definitely agree. This is something that should have been adjusted many versions ago :+1:

DarthGandalf:
@Zarthus may you update the pull request?

Zarthus:
@DarthGandalf Done, was there anything else that needed doing?

PR #824, opened 2015-01-25
Don't polling to wait the exit of threads in CThreadPool destructor

OGAWAHirofumi:
All values are protected by m_mutex. So we don't need the polling to
wait m_num_threads==0 with wakeups, instead simply use CConditionVariable.

### DIFF
--- a/src/Threads.cpp
+++ b/src/Threads.cpp
@@ -84,15 +84,12 @@ void CThreadPool::finishJob(CJob *job) const {
 }
 
 CThreadPool::~CThreadPool() {
-	/* Anyone has an idea how this can be done less ugly? */
 	CMutexLocker guard(m_mutex);
 	m_done = true;
 
-	while (m_num_threads > 0) {
+	if (m_num_threads > 0) {
 		m_cond.broadcast();
-		guard.unlock();
-		usleep(100);
-		guard.lock();
+		m_exit_cond.wait(m_mutex);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


md-5:
:+1:

src/Threads.cpp
```diff
@@ -134,6 +131,9 @@ void CThreadPool::threadFunc() {
 	assert(m_num_threads > 0 && m_num_idle > 0);
 	m_num_threads--;
 	m_num_idle--;
+
+	if (m_num_threads == 0)
```
DarthGandalf: I'm not sure how this will interact with MAX_IDLE_THREADS... Let's check for m_done here.

OGAWAHirofumi: m_done is not what we want here. m_done is true if destructor is called even if there are several threads are still running. The destructor wants to wait all threads are gone.

About m_idle, it is always "m_idle <= m_num_threads".

DarthGandalf: I mean that thread may go out from while loop by 2 reasons: m_done, and if the thread is idle.
If at some point ZNC starts killing idle threads after some timeout for example, the last killed by this reason thread will trigger the condition var.
Of course, m_num_threads needs to be checked, I didn't mean m_done only :)
`if (m_num_threads == 0 && m_done)` would help.

OGAWAHirofumi: Ok. Make sense.

OGAWAHirofumi:
Updated commit was pushed. github seems to be picked up automatically.

Issue #368, opened 2013-06-06
Option to play back messages only once


miggyb:
My ZNC setup is one account with two clients, weechat on the same server that ZNC is on, as well as a phone client that connect/disconnects as needed.

Sometimes I'll connect on my phone just to see if there's anything interesting happening in any of the channels I'm on, see nothing, then disconnect. Since clearbufferonmsg doesn't see any activity, it doesn't clear the current buffer, and next time I connect, I get the same messages over again.

This might be a sane default for most people since it does seem hard to detect whether a message sent from ZNC was actually received by the user or not, but since my phone is my "secondary" device, I'd like to take that risk since I can always check with weechat to see if I've actually missed anything.

PR #825, opened 2015-01-25
Extend clearbufferonmsg module and some znc commands for flexible buffer clear

OGAWAHirofumi:
Current clearbufferonmsg clears buffer on some user actions (OnMsg,
OnCTCP, OnAction, OnNotice, OnPart, OnTopic). And user can't change
this fixed built-in rule.

This built-in rule to clear buffer is too inflexible. So, this
extends clearbufferonmsg to satisfy more users on different
situations.

1) Add OnUserQuit() callback. On smartphone, user doesn't want to see
same lines repeatedly. But, meanwhile, user doesn't want to miss
lines when connection was lost.

To do it, this uses OnUserQuit() callback. With this callback,
clearbufferonmsg can clear buffer if user quited client explicitly.
And when connection was lost, buffer is still not cleared.

2) Add args to clearbufferonmsg to allow that user choose callbacks to clear.

msg    - clear on OnUserMsg callback
ctcp   - clear on OnUserCTCP callback
action - clear on OnUserAction callback
notice - clear on OnUserNotice callback
part   - clear on OnUserPart callback
topic  - clear on OnUserTopic callback
quit   - clear on OnUserQuit callback

all    - set all options above

By default, clear on msg, ctcp, action, notice, part, topic. (except
quit for backward compatibility).

Example usage in znc.conf:

[set all options]
	LoadModule = clearbufferonmsg all

[on quit and part]
	LoadModule = clearbufferonmsg !all quit part

3) Add "-silent" to ClearBuffer/ClearAllChannelBuffers/ClearAllQueryBuffers

User may want more flexible option. Then, user's client can control
buffers with those commands.  For example, client can send "ZNC
ClearBuffer #channel" command when got focus to channel window.

But, those commands are assuming user interactive, and send back the
status messages to client. This spams client if those commands was
used in client plugin for example.

So, this adds "-silent" option to those commands to prevent spams client.


DarthGandalf:
I didn't read through the whole change yet, but let's not mix unrelated changes in one commit.

dgw:
:+1: to the overhaul, but of course I defer to @DarthGandalf when it comes to how the code is changed.

OGAWAHirofumi:
I separated this to 3 commits. Hope this time is reviewable.

OGAWAHirofumi:
Example plugin of hexchat (and maybe xchat) to clear buffer on focus. (save this as .config/hexchat/addons/clearonfocus.pl)

```perl
use strict;
use warnings;

Xchat::register('Clear On Focus', '1.00', 'Clear ZNC buffer on focus.');
Xchat::hook_print('Focus Tab', \&on_focus);

sub on_focus
{
	my $type = Xchat::context_info->{'type'};
	my $channel = Xchat::context_info->{'channel'};

	#Xchat::print("focus => $type, $channel");

	# is query or channel?
	if (($type == 2 or $type == 3) and length($channel) > 0) {
	    Xchat::command("ZNC ClearBuffer -silent $channel");
	}

	return Xchat::EAT_NONE;
}
```

OGAWAHirofumi:
This will close issue #368.

include/znc/Modules.h
```diff
@@ -720,6 +720,10 @@ class CModule {
 	 *  @return See CModule::EModRet.
 	 */
 	virtual EModRet OnUserTopicRequest(CString& sChannel);
+	/** This module hook is called when a user requests to quit from network.
+	 *  @param sMessage The quit message the client sent.
+	 */
+	virtual void OnUserQuit(CString& sMessage);
```
DarthGandalf: Not EModRet?

modules/clearbufferonmsg.cpp
```diff
@@ -15,72 +15,153 @@
...
+	}
 
 	void ClearAllBuffers() {
 		CIRCNetwork* pNetwork = GetNetwork();
+
 		if (pNetwork) {
 			const vector<CChan*>& vChans = pNetwork->GetChans();
 
-			for (vector<CChan*>::const_iterator it = vChans.begin(); it != vChans.end(); ++it) {
+			for (auto pChan : vChans) {
```
DarthGandalf: Any special need for `auto` here? `CChan*` looks fine

modules/clearbufferonmsg.cpp
```diff
@@ -15,72 +15,153 @@
...
+		}
+	}
+
+	virtual bool OnLoad(const CString& sArgs, CString& sMessage) override {
+		VCString vsOpts;
+
+		sArgs.Split(" ", vsOpts, false);
+
+		for (auto& sOpt : vsOpts) {
+			if (sOpt.Left(1).Equals("!"))
```
DarthGandalf: `.StartsWith()`

modules/clearbufferonmsg.cpp
```diff
@@ -15,72 +15,153 @@
...
+				SetRule(sOpt.substr(1), false);
```
DarthGandalf: `.TrimLeft()`

DarthGandalf:
As for `-silent`, please check #620... Also, @jpnurmi

include/znc/Modules.h
OGAWAHirofumi: Does return value make sure on QUIT command? User will disconnect soon.

modules/clearbufferonmsg.cpp
OGAWAHirofumi: No special, so auto. If CChan* was preferred, I'm fine with it.

modules/clearbufferonmsg.cpp
OGAWAHirofumi: Oh, sure. Will fix.

OGAWAHirofumi:
Changes.

- Remove "auto" usages.
- Use StartsWith() instead

OGAWAHirofumi:
ping

include/znc/Modules.h
DarthGandalf: Some module might want to abort the disconnect, and not make client to disconnect.

OGAWAHirofumi: It is client request. ZNC can't control whether disconnect or not. If you really want to return status, i will.

But I'd like to make sure it again.

OGAWAHirofumi:
Changed to return value

include/znc/Modules.h
DarthGandalf: QUIT doesn't cause disconnect itself. Usually after QUIT either server or
client closes connection.

2015-02-21 10:13 GMT-08:00 OGAWA Hirofumi <notifications@github.com>:

OGAWAHirofumi: Right. But it is QUIT from client, so it is clear client disconnect really soon. (Anyway, I added return value)

OGAWAHirofumi:
thanks

Issue #786, opened 2014-12-22
Why does ZNC ship with *NickServ and *Q?


Mikaela:
As far as I am aware, *q is only supported on Quakenet and nowadays there aren't many networks that have NickServ, but don't support SASL. I am only able to name two (OFTC has *cert though) networks where there is NickServ without SASL, but *NickServ module didn't even to do anything there. I think they used HybServ2.

It looks to me like they could be external modules.

Update: according to #znc at freenode, there are multiple networks with NickServ, but without SASL. *perform would probably work as well with them though.

> synIRC, dalnet, YourBNC, Rizon, Swiftirc, SlashNet - They all don't support sasl,  but have nickserv
##

DarthGandalf:
> Is the Q module still maintained? #554 is something I reported a long time ago and still seems to be in znc 1.4.

Apparently, it's not. @ahf ping?
Dropping Q module (aka making it external, and letting ahf to maintain it outside of ZNC repo) would solve this issue, and will help with #1494

ahf:
I have completely overlooked #554. I'm not using the Q module anymore and I'm hardly active on Quakenet either.

I could take a look at #554 during the upcoming week. I don't think I'm the right person to take the decision if the Q module should be maintained externally, internally, or dropped -- I don't know how many znc users that are using it. If #554 is the "only" issue right now I don't see why it would make sense to remove it though. It's not like the Quakenet development community releases early and often! :-)

dgw:
Or roll the Q module into #1496? As long as modules are being consolidated, why not?

DarthGandalf:
> why not?

Because Q is useful on a single network, while sasl and nickserv are useful much more widely. But every single piece of code is an additional burden for maintenance.

ahf:
I also don't think it's a good idea to merge it into a module that tries to standardise the authentication schemes used on the more up-to-date networks. To put it bluntly, I personally think the Quakenet authentication scheme should be avoided. I hope nobody else is using it.

DarthGandalf:
> I don't know how many znc users that are using it.

Without ZNC sending us its usage stats, it's really hard to tell.

> I don't think I'm the right person to take the decision if the Q module should be maintained externally, internally, or dropped

I didn't drop it before because you wanted to look after this module and had a strong opinion on having it in ZNC repo. I'll drop it after 1.7 then, which is exactly the same as making it external - it's only a matter of updating URL to the "fresh" version [here](https://wiki.znc.in/Q)

[...]

PR #833, opened 2015-01-31
nickserv: use NICKSERV instead of PRIVMSG

Mikaela:
Ref: https://github.com/znc/znc/issues/786#issuecomment-68104883

The NickServ command is usually server-side alias that ensures that
NickServ is U-lined server and not random user who has /nicked to
NickServ on misconfigured server that allows that.

NS would probably also work, but NICKSERV is more common.
Ref: https://github.com/ProgVal/Limnoria/pull/724#issuecomment-45381342

### DIFF
--- a/modules/nickserv.cpp
+++ b/modules/nickserv.cpp
@@ -84,7 +84,7 @@ public:
 		}
 
 		if (GetNV("IdentifyCmd").empty()) {
-			SetNV("IdentifyCmd", "PRIVMSG NickServ :IDENTIFY {password}");
+			SetNV("IdentifyCmd", "NICKSERV IDENTIFY {password}");
 		}
 
 		return true;


modules/nickserv.cpp
```diff
@@ -84,7 +84,7 @@ class CNickServ : public CModule {
 		}
 
 		if (GetNV("IdentifyCmd").empty()) {
-			SetNV("IdentifyCmd", "PRIVMSG NickServ :IDENTIFY {password}");
+			SetNV("IdentifyCmd", "NICKSERV :IDENTIFY {password}");
```
jpnurmi: should the colon also be removed?

Mikaela: I testedx on freenode and it works with both with and without colon. I am not sure if it should be there, but I added it just in case, but I can remove it.

md-5:
"usually"

Mikaela:
> "usually"

(I am not sure if my response to this is wanted, but) I mean that the networks usually have the NICKSERV command as it exists in most of example configurations and it's de-facto standard. The network can also be missing these commands, but in that case the network operators are often considered as incompetent.

jpnurmi:
I guess "usually" (so "modern" IRCd's?) is fine, because the command is configurable.

Mikaela:
> I guess "usually" (so "modern" IRCd's?) is fine, because the command is configurable.

At least it works on Charybdis, Unreal, InspIRCd, Hybrid and it should work on every IRCd that has services. And as you said, if some network is configured incorrectly, you can configure it by yourself.

Mikaela:
Is there something horribly wrong with this pull request and should this be moved to other branch?

IRC:

```
2015-02-19 06:20:26+0200 < Mikaela> jpnurmi: should I be moving my older pull requst to master to 1.6 branch?
2015-02-19 06:22:15+0200 < Mikaela> https://github.com/znc/znc/pulls/Mikaela
2015-02-19 06:23:20+0200 < ChauffeR> https://github.com/znc/znc/pull/833#issuecomment-72309934
2015-02-19 06:23:21+0200 < ZNC-Linker> https://github.com/znc/znc/pull/833 “nickserv: use NICKSERV instead of PRIVMSG” (open)
2015-02-19 06:23:23+0200 < Mikaela> I would also like to know if there is anything wrong with those pull requests that I could fix as they seem to be taking a long time to get merged
2015-02-19 06:24:00+0200 < ChauffeR> i don't see why that one would get merged
2015-02-19 06:24:17+0200 < Mikaela> ChauffeR: could you name me a network that has NickServ without the NICKSERV command? (use /quote if your IRC client doesn't send unnown commands to server)
2015-02-19 06:24:52+0200 < Mikaela> and also link me to IRCd that doesn't have it, all example configs that I have seen support that
2015-02-19 06:24:58+0200 < ChauffeR> i can think of a few
2015-02-19 06:25:09+0200 < Mikaela> it's also security issues
2015-02-19 06:25:09+0200 < Mikaela> like which?
2015-02-19 06:25:22+0200 < Mikaela> I must leave for school soon, but if you can get me some kind of list, I can take a look at those networks
2015-02-19 06:25:39+0200 < ChauffeR> it does not matter
2015-02-19 06:26:22+0200 < ChauffeR> and, "security issues" is only true if the user loads the module where he shouldn't
2015-02-19 06:26:28+0200 < Mikaela> no, it does matter. Are those network really missing the command?
2015-02-19 06:26:28+0200 < Mikaela> What is the first of those that comes to mind?
2015-02-19 06:26:38+0200 < ChauffeR> yes, they are
2015-02-19 06:27:04+0200 < Mikaela> which?
2015-02-19 06:31:45+0200 < Mikaela> ChauffeR: I am leaving now, but it would be nice if you could add comment with soome examples of networks where it doesn't work to that pull request, thanks in advance
2015-02-19 06:31:45+0200 < Mikaela> bye
2015-02-19 06:32:01+0200 [Mikaela away: Detached from tmux since 2015-02-19 06:32:01+0200]
2015-02-19 06:32:56+0200 < ChauffeR> er, for a number of reasons i can't tell you which ones
```

I guess that this means that this person got some networks from their head that really don't exist or they all support this command. I am still open for list of networks that don't support this.

Mikaela:
Oh and I am also open for list of IRCds that don't support it.

Issue #820, opened 2015-01-19
Revert "Don't expose ZNC version in CTCP VERSION."


KindOne-:
https://github.com/znc/znc/commit/94f2b4dc971311de1955f08f38f443042fba431b

Under what reasoning do we need to hide znc's version number when almost all IRC clients show the version number by default?
What exactly do we gain when we hide the version number?

```AdiIRC 1.9.6 Beta Build 2014/12/30 64 Bit```
```AndChat 1.4.3.2 http://www.andchat.net```
```Babbel v120107.14 Win32 Edition - http://www.babbelirc.com```
```bitchx-1.0c18(freebsd 9.2-stable): neonapple/v0.14i : dont you wish you had it?```
```Colloquy 2.4.3 (6011) - Mac OS X 10.10.1 (Intel) - http://colloquy.info```
```dircproxy 1.2.0 - http://dircproxy.googlecode.com/```
```eggdrop v1.6.20```
```HexChat 2.10.2 [x64] / Windows 7 SP1 [3.45GHz]```
```ircII 2.8.2-EPIC3.004 *IX - All the things phone wont include.....```
```irssi v0.8.15 - running on Linux x86_64```
```KVIrc 4.2.0 svn-6190 'Equilibrium' 20120701 - build 2012-07-04 14:48:08 UTC - Windows 8 (x64) (Build 9200)```
```Linkinus (Agent build 24570)```
```LimeChat for Mac 2.42```
```Miranda NG 0.95.3 build #11706 x64 (IRC v.0.11.0.2), (c) 2003-2014 J.Persson, G.Hazan```
```mIRC v7.32 Khaled Mardam-Bey```
```Quassel IRC v0.10.0 (git-575f27e) (built on Apr 8 2014 19:13:01) -- http://www.quassel-irc.org```
```Smuxi 0.8.9.2 (Debian 0.8.9.2-2) - running on GNU/Linux (elementary OS) i686```
```Textual IRC Client: www.textualapp.com — v5.0.3```
```Trillian IRC Module 5.0```
```WeeChat 1.1-rc1 (Dec 28 2014)```
```xchat 2.8.7f Windows 6.1 [i686/3.60GHz]```
```XChat-WDK 1508-3 [x64] / Windows 7 [2.00GHz]```
```X-Chat Aqua appstore-1.16.2-0-gb5ac411 (xchat 2.8.8-svn1523) Darwin 14.0.0 [x86_64/2.40GHz/SMP]```

[...]

PR #836, opened 2015-01-31
Add ZNC to CTCP VERSION reply even if client is connected.

DarthGandalf:
Previously, it replied to CTCP VERSION, but only if no client is connected.

See https://github.com/znc/znc/issues/820#issuecomment-70506203

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -234,6 +234,10 @@ void CClient::ReadLine(const CString& sData) {
 				sCTCP.LeftChomp();
 				sCTCP.RightChomp();
 
+				if (sCTCP.Token(0) == "VERSION") {
+					sCTCP += " via " + CZNC::GetTag(false);
+				}
+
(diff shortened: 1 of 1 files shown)


src/Client.cpp
```diff
@@ -234,6 +234,10 @@ void CClient::ReadLine(const CString& sData) {
 				sCTCP.LeftChomp();
 				sCTCP.RightChomp();
 
+				if (sCTCP.Token(0) == "VERSION") {
+					sCTCP += " via " + CZNC::GetTag(false);
```
md-5: This would probably need to use the option added in the other PR

DarthGandalf: `GetTag()` uses it...

md-5: Ignore me then

Mikaela:
How about CTCP FINGER? At least WeeChat defaults it to showing version and I have it overridden as `ZNC %VERSION%`.

PR #842, opened 2015-02-03
Fix CIRCNetwork::FindChans() and FindQueries() to be case-insensitive

jpnurmi:
The playback module failed to clear a buffer, because it tried to
clear "NickServ" whereas ZNC had internally stored it has "nickserv".


jpnurmi:
Maybe cleaner to extend CString::WildCmp() with an optional case sensitivity argument.

jpnurmi:
> Maybe cleaner to extend CString::WildCmp() with an optional case sensitivity argument.

Which means new API in CString and therefore targets master (#869). Meanwhile, this is fine as a temporary fix for the 1.6 series.

jpnurmi:
no objections for two weeks - cherry-picked to 1.6: https://github.com/znc/znc/commit/7345a6ee3ac8547da46e6906843937f8e43aed80

PR #843, opened 2015-02-03
Add ecdsa-nist256p-challenge into sasl #746

CustaiCo:
This fixes issue #746. The only issue i see with this is that it's important you back up the configuration files as there is no way to get the private key outside of just copying the file. The public key may be requested from the sasl module at any time.

### DIFF
--- a/modules/sasl.cpp
+++ b/modules/sasl.cpp
@@ -17,6 +17,14 @@
 #include <znc/IRCNetwork.h>
 #include <znc/IRCSock.h>
 
+#ifdef HAVE_LIBSSL
+#include <openssl/ec.h>
+#include <openssl/ecdsa.h>
+#include <openssl/evp.h>
+#include <openssl/rand.h>
+#include <openssl/sha.h>
+#endif //HAVESSL
+
 static const struct {
 	const char *szName;
 	const char *szDescription;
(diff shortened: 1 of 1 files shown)


DarthGandalf:
May you follow codestyle around?

CustaiCo:
Would you care to elaborate? I couldn't find a code style document on the wiki, and I've tried to make it as similar as possible to before the DH-BLOWFISH module was removed.

Mikaela:
I am not sure if I understand the pr message.

> The only issue i see with this is that it's important you back up the configuration files as there is no way to get the private key outside of just copying the file

If I understand correctly, ZNC generates the key and stores it to `~/.znc/configs/znc.conf` and these both seem weird to me. Shouldn't the key be generated outside of ZNC and be uploaded with webadmin like \*cert and then be stored to similar path which \*cert does.

*I am not a coder and this is just how I read the quoted section.*

jpnurmi:
Start with squashing the commits (eg. ```git rebase -i origin/master``` if origin is the upstream remote) and then:
```
pick de99fc4 Allows use of ECDSA-NIST256P-CHALLENGE for sasl
f b62350a Added ifdefs so can be built without ssl
f 55677ed Changed to use .pem files instead of fmemopen
f 9c968c9 fixed memory leak, minor whitespace cleanup
```
to make it a single commit. Once you have a single commit, if you want to make further amendments, you can simply do ```git commit --amend```. Then pass ```--force```when pushing to update the PR.

CustaiCo:
It actually writes it to ~/.znc/users/[user]/networks/[network]/moddata/sasl/sasl.pem
It used to store it in the .registery file in that same directory, but since fmemopen isn't supported on osx I had to change it to write it out to its own separate file. The point is that there is no command to print the private key (even though it'd not be a very long string either base64 or hex encoded). It's not signed or anything like that so there isn't really a reason to need to use a separate tool to generate the key. The format of the output file is because other ways of getting the key into and out of openssl are very painful.

jpnurmi:
Also, a descriptive commit title would be nice. "Fix for issue #746" doesn't tell much when browsing the git history. :)

Mikaela:
> It actually writes it to ~/.znc/users/[user]/networks/[network]/moddata/sasl/sasl.pem

What if there will be another method which uses some kind of certificate file? Maybe the filename should be something like `ecdsa.pem` or similar.

Can I upload my self-generated key from elsewhere (another client) to ZNC if I don't have access to the OS meaning webadmin?

DarthGandalf:
All the code around uses something like hungarian notation for variables, e.g. bFoo for bool, m_ for member fields.

modules/sasl.cpp
```diff
@@ -24,9 +32,132 @@ static const struct {
...
+	/* Get base64 encoded private key from key */
+	CString GetPubBase64(void) {
+		CString ret;
+		size_t keybuflen;
+		unsigned char *keybuf;
+
+		if (!loaded)
+		    return ret;
+
+		keybuflen = (size_t) i2o_ECPublicKey(key, NULL);
```
DarthGandalf: `size_t uKeyBufLen = ...`
No need to list all the variables in beginning of function.

modules/sasl.cpp
```diff
@@ -24,9 +32,132 @@ static const struct {
...
+		ret.Base64Encode();
+
+		free(keybuf);
+		return ret;
+	}
+
+	/* returns false if key fails to parse */
+	bool LoadKey(const CString& sPath) {
+		FILE *keyfile;
+		CString path = sPath + "/sasl.pem";
```
DarthGandalf: I agree with `ecdsa.pem`

modules/sasl.cpp
```diff
@@ -24,9 +32,132 @@ static const struct {
...
+		else
+			loaded = true;
+
+		return loaded;
+	}
+
+	bool SaveKey(const CString& sPath) {
+		FILE *keyfile;
+		CString path = sPath + "/sasl.pem";
+		if(!loaded)
```
DarthGandalf: space after `if`

Mikaela:
Has anything changed? There seems to be commit after the comments.

DarthGandalf:
Github does not notify about new comments

DarthGandalf:
@Mikaela And I don't see anything fixed in the new commit anyway.

Mikaela:
> Github does not notify about new commits

That is why I commented.

 

> @Mikaela And I don't see anything fixed in the new commit anyway.

ok.

DarthGandalf:
And this mechanism isn't needed anymore, see #746 for details.

Mikaela:
https://github.com/znc/znc/issues/746#issuecomment-119274435 to be accurate.

Issue #125, opened 2012-02-06
New NV system


DarthGandalf:
Current name-value pairs don't allow very much.

Also when user is cloned, need to clone his NV.
##

PR #850, opened 2015-02-06
Implemented StripControls for channel messages and notices

### LINKED ISSUE #387: Strip color/bold codes per network [$10]
I would like to see a feature where you can set networks indepently to strip color/bold codes etc from some nets and not others this is particually useful if you use eggdrop through znc and need to use plain text as a trigger rather than jumping through hoops trying to strip codes on the bot itself

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/651429-strip-color-bold-codes-per-network-10?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ManiacTwister:
Implemented StripControls for channel messages and notices. Either network wide or for selected channels.

This Fixes #387

### DIFF
--- a/modules/controlpanel.cpp
+++ b/modules/controlpanel.cpp
@@ -123,6 +123,7 @@ class CAdminMod : public CModule {
 				{"Encoding",            str},
 #endif
 				{"QuitMsg",             str},
+				{"StripControls",       boolean},
 			};
 			PrintVarsHelp(sVarFilter, nvars, ARRAY_SIZE(nvars), "The following variables are available when using the SetNetwork/GetNetwork commands:");
 		}
@@ -134,7 +135,8 @@ class CAdminMod : public CModule {
 				{"Buffer",              integer},
 				{"InConfig",            boolean},
 				{"AutoClearChanBuffer", boolean},
-				{"Detached",            boolean}
+				{"Detached",            boolean},
+				{"StripControls",       boolean},
 			};
 			PrintVarsHelp(sVarFilter, cvars, ARRAY_SIZE(cvars), "The following variables are available when using the SetChan/GetChan commands:");
 		}
@@ -486,6 +488,8 @@ class CAdminMod : public CModule {
 #endif
 		} else if (sVar.Equals("quitmsg")) {
 			PutModule("QuitMsg = " + pNetwork->GetQuitMsg());
+		} else if (sVar.Equals("stripcontrols")) {
+			PutModule("StripControls = " + CString(pNetwork->StripControls()));
 		} else {
 			PutModule("Error: Unknown variable");
 		}
@@ -580,6 +584,10 @@ class CAdminMod : public CModule {
 		} else if (sVar.Equals("quitmsg")) {
 			pNetwork->SetQuitMsg(sValue);
 			PutModule("QuitMsg = " + pNetwork->GetQuitMsg());
+		} else if (sVar == "stripcontrols") {
+			bool b = sValue.ToBool();
+			pNetwork->SetStripControls(b);
+			PutModule("StripControls = " + CString(b));
 		} else {
 			PutModule("Error: Unknown variable");
 		}
@@ -702,6 +710,8 @@ class CAdminMod : public CModule {
 				PutModule(pChan->GetName() + ": AutoClearChanBuffer = " + sValue);
 			} else if (sVar == "detached") {
 				PutModule(pChan->GetName() + ": Detached = " + CString(pChan->IsDetached()));
+			} else if (sVar == "stripcontrols") {
+				PutModule(pChan->GetName() + ": StripControls = " + CString(pChan->StripControls()));
 			} else if (sVar == "key") {
 				PutModule(pChan->GetName() + ": Key = " + pChan->GetKey());
 			} else {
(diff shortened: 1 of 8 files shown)


modules/controlpanel.cpp
```diff
@@ -123,6 +123,7 @@ class CAdminMod : public CModule {
 				{"Encoding",            str},
 #endif
 				{"QuitMsg",             str},
+				{"StripControls", boolean},
```
DarthGandalf: Please don't break formatting

ManiacTwister: Whoops. Fixed.

DarthGandalf:
What about queries?

ManiacTwister:
@DarthGandalf I'm unsure how the setting for queries should look like.. Reuse the network wide setting or a dedicated setting for (all) queries?

ManiacTwister:
@DarthGandalf Added extra setting to strip controls from private messages network wide. A blacklist for nicks would be overkill imho.

jpnurmi:
Do we necessarily need a separate setting for channels and queries? Or could the same network level StripControls setting be used for both queries and as a default for channels, and then individual channels can override if desired.

ManiacTwister:
Thats what i asked here: https://github.com/znc/znc/pull/850#issuecomment-73301956

I'm really unsure about it... I can imagine users which would like to allow colors in queries but disallow them in channels. In that case a separate setting comes in handy. On the other hand disabling stripcolors for all channels manually is still possible without a separate setting ...

I'm open for both ways, choose which one you prefer and i'll update the PR accordingly

jpnurmi:
There's always someone who thinks that this and that additional setting is useful for some corner case. Personally I'm in favor of keeping settings to the minimum. The less settings the better. :) Optimize for the convenience of the most common use. Controlpanel supports wildcards, so it's pretty easy to change the setting for all channels.

ManiacTwister:
@jpnurmi Agreed. :) Done.

DarthGandalf:
07.02.2015 21:20, J-P Nurmi пишет:
> Personally I'm in favor of keeping settings to the minimum
+1

DarthGandalf:
> Also: Is this failing travis build (https://travis-ci.org/znc/znc/jobs/49888849) my fault? 

I think it's OOM-killer. Modpython's source is already 4.3 MB of generated C++ code, and address sanitizer adds to it.
You're adding 6 more functions to be wrapped, and this is enough to go over the threshold.

md-5:
The memory limit is 3gb of ram :\

jpnurmi:
Hi, sorry for the delay. I've been a bit busy with stuff. If you want to make the submission perfect, you could fix the coding style in IRCSock.cpp. ZNC uses a space after if, before the opening parenthesis. It's not entirely consistent across the code base, just me being pedantic. :) Furthermore, the body of the for-loop in CIRCNetwork::SetStripControls() has a double indentation.

That was just nitpick, but I have also one real finding; I believe we should also strip controls for both channel and private CTCP ACTIONs. Otherwise looks all good. :)

ManiacTwister:
I'll fix the issues at weekend

ManiacTwister:
@jpnurmi Everything should've been fixed now.

modules/data/webadmin/tmpl/add_edit_network.tmpl
```diff
@@ -129,6 +129,25 @@
...
+					<div class="inputlabel">Strip Controls (Channels):</div>
+					<div class="checkbox">
+						<input type="checkbox" name="chanstripcontrols" id="chanstripcontrols_checkbox"
+								 title="Strips control codes (Colors, Bold, ..) from channel messages."
+							<? IF ChanStripControls ?>checked="checked"<? ENDIF ?> />
+						<label for="chanstripcontrols_checkbox">Enabled</label>
+					</div>
+				</div>
+				<div class="subsection">
+					<div class="inputlabel">Strip Controls (Queries):</div>
```
jpnurmi: i thought we already combined these two?

ManiacTwister:
@jpnurmi Well, you are totally right. Pushed the wrong branch, sorry. Fixed it :>

jpnurmi:
Merged to master. Thanks! :)

kylef:
Would this not be better suited as a module instead of bloating the core? I can't see that there would be a huge demand for this.

jpnurmi:
Good point, I didn't even think about it. :/ Oh well, we can still change it...

jpnurmi:
The module could be something as simple as https://github.com/jpnurmi/znc/commit/cdc4de9248db9013b3ded4b9821911c9990a5c1b. It could be loaded either on network or user level. The trade-off is that there wouldn't be control over individual channels. Would you guys see that as a big problem?

DarthGandalf:
Settings of it could have list of channels...
Or, if #125 is fixed, channel could have a setting for this module without
having a setting in core.

2015-02-24 12:57 GMT-08:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
The new settings system proposed in #125 sounds great. Did you just
volunteer? :)

DarthGandalf:
I didn't volunteer, I don't have enough time.
But I'd prefer a module instead of a new setting in core regardless.

jpnurmi:
https://github.com/znc/znc/pull/917

PR #851, opened 2015-02-06
Extend awaystore with -chans option, which records channel hilights to the awaystore as well as only PMs

reality:


DarthGandalf:
Watch module doesn't store highlights already?

reality:
Watch is for generic word tracking, this is particularly for storing hilights you got in channels when you're away and showing them when you return (in the same manner it already does for PMs).

applesjgtl:
+1

PR #854, opened 2015-02-10
Add experimental support for DANE / TLSA

### LINKED ISSUE #784: Add support for DANE / TLSA
ZNC should try to authenticate certificates using DANE when connecting to TLS servers.

[DANE](https://en.wikipedia.org/wiki/DNS-based_Authentication_of_Named_Entities) is a protocol which replaces or reinforces the X.509 validation of certificates used by SSL/TLS, which is usually done by checking the signature added by a certification authority. With DANE, the certificate fingerprint can also be checked with a DNS query, hopefully authenticated by DNSSEC. It is very useful with self-signed certificates, as it allows the server administrator to explicitly tell the client which certificate it should receive and to ignore the signature validation.

References:
- [RFC 6698](http://tools.ietf.org/html/rfc6698)
- [irssi's announcement](http://ahf.me/articles/2013/09/14/enhancing-ssl-security-for-irc-dane-support/) with a more detailed description and motivation
- [irssi's implementation](https://github.com/irssi/irssi/blob/8bd575df2ec0d4a17b6f59116d555b64f5bb1312/src/core/network-openssl.c#L209-L241) which uses the [dnsval library](http://dnssec-tools.org/) (libval-{bin,dev} on Debian Jessie)
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7182050-add-support-for-dane-tlsa?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Alef-Burzmali:
This commit adds experimental support for DANE validation of TLS server certificates with the libval library from the DNSSEC-Tools package, like the [irssi implementation](https://github.com/irssi/irssi/blob/8bd575df2ec0d4a17b6f59116d555b64f5bb1312/src/core/network-openssl.c#L209-L241).

You have to [patch Csocket](https://gist.github.com/Alef-Burzmali/d6ebbf6d57a6f4fab191) to use it, as the library requires the `ssl` object to test the certificate(s) directly. I'll open a PR there if this one is well received.

I've tested the `PKIX-EE(1)` and `DANE-EE(3)` usages with both valid and invalid records, I was unable to test `PKIX-TA(0)` or `DANE-TA(2)`. `DANE-EE` should be the most common usage anyway, it's the one that validates self-signed certs.

As this is the biggest C++ patch I've ever written, I'll appreciate any feedback. :-)

Fix #784.

### DIFF
--- a/src/Socket.cpp
+++ b/src/Socket.cpp
@@ -26,6 +26,11 @@
 #include <unicode/ucnv_cb.h>
 #endif
 
+#ifdef HAVE_DANE
+#include <validator/validator.h>
+#include <validator/val_dane.h>
+#endif
+
 #ifdef HAVE_LIBSSL
 // Copypasted from https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29 at 22 Dec 2014
 static CString ZNC_DefaultCipher() {
@@ -106,10 +111,47 @@ void CZNCSock::SSLHandShakeFinished() {
 		return;
 	}
 	CString sHostVerifyError;
-	if (!ZNC_SSLVerifyHost(m_HostToVerifySSL, pCert, sHostVerifyError)) {
+	bool bSSLHostOk = ZNC_SSLVerifyHost(m_HostToVerifySSL, pCert, sHostVerifyError);
+	if (!bSSLHostOk) {
 		m_ssCertVerificationErrors.insert(sHostVerifyError);
 	}
 	X509_free(pCert);
+
+#ifdef HAVE_DANE
+	struct val_danestatus* pDaneStatus = nullptr;
+	struct val_daneparams pDaneParams;
+	pDaneParams.port = GetPort();
+	pDaneParams.proto = DANE_PARAM_PROTO_TCP;
+
+	if (val_getdaneinfo(nullptr, m_HostToVerifySSL.c_str(), &pDaneParams, &pDaneStatus) == VAL_DANE_NOERROR) {
+		if (pDaneStatus != nullptr) {
+			int iPkixValidation = 1;
+			int iDaneResult = val_dane_check(nullptr, GetSSLObject(), pDaneStatus, &iPkixValidation);
+
+			if (iDaneResult == VAL_DANE_NOERROR) {
+				// DANE is succesful
+				if (pDaneStatus->usage == DANE_USE_TA_ASSERTION && !bSSLHostOk) {
+					DEBUG(GetSockName() + ": DANE: Valid TLSA record, but certificate is not valid for this host and usage is DANE-TA");
+				} else if (iPkixValidation == 0) {
+					DEBUG(GetSockName() + ": DANE: Valid TLSA record, no need for X509 validation");
+					val_free_dane(pDaneStatus);
+					return;
+				} else {
+					DEBUG(GetSockName() + ": DANE: Valid TLSA record, but X509 validation required");
+				}
+			} else if (iDaneResult == VAL_DANE_CHECK_FAILED) {
+				DEBUG(GetSockName() + ": DANE: Invalid TLSA record");
+				m_ssCertVerificationErrors.insert("DANE: TLSA record forbids that certificate");
+			} else {
+				DEBUG(GetSockName() + ": DANE: Internal error while checking TLSA record");
+			}
+			val_free_dane(pDaneStatus);
+		}
+	} else {
+		DEBUG(GetSockName() + ": DANE: No TLSA record");
+	}
+#endif /* HAVE_DANE */
+
 	if (m_ssCertVerificationErrors.empty()) {
 		DEBUG(GetSockName() + ": Good cert");
 		return;
--- a/src/Socket.cpp
+++ b/src/Socket.cpp
@@ -44,11 +44,19 @@ static CString ZNC_DefaultCipher() {
 }
 #endif
 
-CZNCSock::CZNCSock(int timeout) : Csock(timeout), m_HostToVerifySSL(""), m_ssTrustedFingerprints(), m_ssCertVerificationErrors() {
+CZNCSock::CZNCSock(int timeout)
+		: Csock(timeout),
+#ifdef HAVE_DANE
+		  m_iDaneLookupResult(VAL_DANE_INTERNAL_ERROR),
+		  m_pDaneStatus(nullptr),
+#endif
+		  m_HostToVerifySSL(""),
+		  m_ssTrustedFingerprints(),
+		  m_ssCertVerificationErrors()
+{
+	Init();
+
 #ifdef HAVE_LIBSSL
-	DisableSSLCompression();
-	FollowSSLCipherServerPreference();
-	DisableSSLProtocols(CZNC::Get().GetDisabledSSLProtocols());
 	CString sCipher = CZNC::Get().GetSSLCiphers();
 	if (sCipher.empty()) {
 		sCipher = ZNC_DefaultCipher();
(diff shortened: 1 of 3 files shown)


Mikaela:
> You have to patch Csocket to use it, as the library requires the ssl object to test the certificate(s) directly. I'll open a PR there if this one is well received.

I highly doubt this is merged until there is PR at Csocket.

https://github.com/jimloco/Csocket

src/Socket.cpp
```diff
@@ -103,6 +108,40 @@ void CZNCSock::SSLHandShakeFinished() {
...
 		return;
 	}
+
+#ifdef HAVE_DANE
+	struct val_danestatus* pDaneStatus = NULL;
+	struct val_daneparams pDaneParams;
+	pDaneParams.port = GetPort();
+	pDaneParams.proto = DANE_PARAM_PROTO_TCP;
+
+	if (val_getdaneinfo(NULL, m_HostToVerifySSL.c_str(), &pDaneParams, &pDaneStatus) == VAL_DANE_NOERROR) {
```
DarthGandalf: This will block ZNC until DNS query is done.

Alef-Burzmali:
Ok, I've put the DNS query in another thread if HAVE_THREADED_DNS is defined. However, as it is the first time I use threads in C++, I would really appreciate if someone could verify I did not do something stupid.

I've also tested `PKIX-TA(0)` or `DANE-TA(2)`, which work.

DarthGandalf:
Right now ZNC creates 2 jobs in separate threads for sockets: DNS resolving of target host, and DNS resolving of bind host. Maybe this should just be a third one?

Alef-Burzmali:
Yes, it should be easier :)

Thanks @psychon for your feedback. You were right, there was an issue with the destructor. With @DarthGandalf's suggestion, it's gone and there no need for CThreadPool::waitForJob() either.

src/Socket.cpp
```diff
@@ -51,6 +60,10 @@ CZNCSock::CZNCSock(int timeout) : Csock(timeout) {
 }
 
 CZNCSock::CZNCSock(const CString& sHost, u_short port, int timeout) : Csock(sHost, port, timeout) {
+#ifdef HAVE_DANE
+	m_iDaneLookupResult = VAL_DANE_INTERNAL_ERROR;
+	m_pDaneStatus = NULL;
+#endif
```
DarthGandalf: Probably it's time to move this from both constructors to some common function.

DarthGandalf:
@ahf you're more familiar with DANE than me. Does this look fine?

Alef-Burzmali:
Rebased on top of master and replaced NULL by nullptr.

Alef-Burzmali:
Rebased on top of master.
I've try to comply with the new coding style, but I may have forgotten some bits, please review it.

src/Socket.cpp
```diff
@@ -93,6 +128,67 @@ int CZNCSock::VerifyPeerCertificate(int iPreVerify, X509_STORE_CTX * pStoreCTX)
...
+	m_pDaneStatus = pDaneStatus;
+}
+
+bool CZNCSock::VerifyDaneAuthentication(SSL* pSSL, bool bSSLHostOk) {
+#ifndef HAVE_THREADED_DNS
+	u_short iPort = GetPort();
+
+	DEBUG(GetSockName() + ": DANE: initiating TLSA lookup for [" << m_HostToVerifySSL << "], port [" << iPort << "]");
+
+	struct val_daneparams pDaneParams;
```
DarthGandalf: "p" stands for pointer, this is not

src/Socket.cpp
```diff
@@ -93,6 +128,67 @@ int CZNCSock::VerifyPeerCertificate(int iPreVerify, X509_STORE_CTX * pStoreCTX)
...
+	m_iDaneLookupResult = val_getdaneinfo(nullptr, m_HostToVerifySSL.c_str(), &pDaneParams, &m_pDaneStatus);
+
+	DEBUG(GetSockName() + ": DANE: finished TLSA lookup for [" << m_HostToVerifySSL << "], port [" << iPort << "]");
+#endif /* !HAVE_THREADED_DNS */
+
+	if (m_iDaneLookupResult != VAL_DANE_NOERROR || m_pDaneStatus == nullptr) {
+		val_free_dane(m_pDaneStatus);
+		m_pDaneStatus = nullptr;
+
+		DEBUG(GetSockName() + ": DANE: No usable TLSA record");
```
DarthGandalf: Let's have different error messages for these cases. Easier to debug.

src/Socket.cpp
```diff
@@ -93,6 +128,67 @@ int CZNCSock::VerifyPeerCertificate(int iPreVerify, X509_STORE_CTX * pStoreCTX)
...
+	int iPkixValidation = 1;
+	int iDaneResult = val_dane_check(nullptr, pSSL, m_pDaneStatus, &iPkixValidation);
+
+	bool bCertVerified = false;
+	if (iDaneResult == VAL_DANE_NOERROR) {
+		// DANE is succesful
+		DEBUG(GetSockName() + ": DANE: Found TLSA record [" << m_pDaneStatus->usage << " " << m_pDaneStatus->selector << " " << m_pDaneStatus->type << "]");
+
+		if (m_pDaneStatus->usage == DANE_USE_TA_ASSERTION && !bSSLHostOk) {
+			DEBUG(GetSockName() + ": DANE: Valid TLSA record, but certificate is not valid for this host and usage is DANE-TA");
```
DarthGandalf: What's the reason for this branch?

Alef-Burzmali:
I've hopefully fixed the 2 issues, and added a comment to explain the reason behind the branch.

However, further testing has shown that the library I've used, libval 2.0, suffers from major flaws that prevent it from being used. The implementation of usage `DANE-TA(2)` is broken and allow to bypass the certificate validation entirely.

Some possible solutions:
* Disable `DANE-TA(2)`:
  * `DANE-EE(3)` is more likely to be used, as it works with any type of certificates, including self-signed ones;
  * `DANE-TA(2)` requires a self-signed CA used to sign several certificates, which is less common; and these certificates need to be valid (not expired, and match the server name);
  * `DANE-TA(2)` can be replaced by `DANE-EE(3)` by listing all the possible certificate in the DNS;
* Use libval 2.1, which is much more secure, but is not yet packaged and has a different API;
* Use another library. Maybe gnutls?
* Wait until DANE support is added to openssl.

I will try to use libval 2.1 instead, in the meantime, this PR should not be used.

DarthGandalf:
Hm.
Good that you've noticed the issue.

DarthGandalf:
Is there an upstream bug, or maybe CVE, about it?

Mikaela:
> * Use another library. Maybe gnutls?

GNUTLS won't work, because it's DANE depends on Unbound which again depends on OpenSSL and Debian and other distributions aren't packaging it with DANE support for this reason.

Via https://github.com/weechat/weechat/pull/121

Alef-Burzmali:
> Is there an upstream bug, or maybe CVE, about it?

I've not found any CVE nor bug report, even though the developers have been [aware](https://www.dnssec-tools.org/pipermail/users/2013-July/000177.html) of the issue since 2013, from [a review](https://www.ietf.org/mail-archive/web/dane/current/msg05758.html) on the DANE working group mailing-list.

However, version 2.0 is no longer supported and has been replaced by [version 2.1](https://dnssec-tools.org/download/#gotodnsval). The main reason I've used it is because it's the version packaged in Debian testing.

Mikaela:
GnuTLS issue on unbound (and OpenSSL or nss) depedency: https://gitlab.com/gnutls/gnutls/issues/21

Alef-Burzmali:
For those of you who were interested, here is the Debian [bug report for libval](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=797470)

Issue #40, opened 2011-04-25
Request: sortable tables in userlist [$5]


Efreak:
I'd like to be able to sort the tables in the webadmin. In particular the userlist; an alternative would be to place the connected users first, but I think it would be easiest to use something like <a href="http://freepages.genealogy.rootsweb.ancestry.com/~gearyfamily/expression-web/sort-table.html">sort-table</a>, which simply requires a line in the header, and applying the 'sortable' class to the table in order to be able to click the table headers.

Issue #772, opened 2014-12-14
Provide infrastructure to write integration tests


DarthGandalf:
It should setup a simple IRC server, connect ZNC to it, connect a simple client to ZNC.
Client and server could contain canned responses provided by the test.

As an option, ZNC itself (maybe a separate instance) can be both server and client, with special testing modules loaded.

Also see #771

PR #860, opened 2015-02-13
Implement table sorting in webadmin, addresses #40

ammaraskar:
Here's what it looks like

I still need to sort out the fixing odd and even rows after the tables are sorted.
Mainly I wanted your opinion on
* Whether this vanilla js solution is fine, or should I use the recent-ish added jquery for maximum compatibility. I believe the main limitation here is its >=IE9 with some of the functions I use. 
* Are the simple little arrows okay
* Is the formatting etc fine

Personally I agree with psychon on the issue page that using a really hefty library is kinda overkill. Which is why I opted to just make a simple sorting function.

### DIFF
--- a/modules/data/webadmin/files/webadmin.css
+++ b/modules/data/webadmin/files/webadmin.css
@@ -6,3 +6,10 @@
 .encoding-settings {
 	width: 500px;
 }
+
+table .sorted::after {
+	content:" ▼";
+}
+table .reverse-sorted::after {
+	content:" ▲";
+}
--- a/modules/data/webadmin/files/webadmin.css
+++ b/modules/data/webadmin/files/webadmin.css
@@ -8,8 +8,8 @@
 }
 
 table .sorted::after {
-	content:" ▼";
+	content:" ▾";
 }
 table .reverse-sorted::after {
-	content:" ▲";
+	content:" ▴";
 }
--- a/modules/data/webadmin/files/webadmin.js
+++ b/modules/data/webadmin/files/webadmin.js
@@ -11,3 +11,87 @@ function floodprotection_change() {
 	}
 }
 
+function make_sortable_table(table) {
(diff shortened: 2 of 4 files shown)


jpnurmi:
I also like the simple sort function approach, but I don't know what is the policy for supported browsers.

PS. Is that 'BLACK UP/DOWN-POINTING TRIANGLE'? To me, looks quite massive. :) How about 'BLACK UP/DOWN-POINTING SMALL TRIANGLE'?

ammaraskar:
Yup, here's what BLACK UP/DOWN-POINTING SMALL TRIANGLE looks like.



small looks better in my opinion, so that leaves the last concern as maximum browser compatibility with dependence on jquery vs a vanilla js solution that doesn't support less than IE9. Looks like that accounts for [1.4% of the total browser market share](http://www.w3schools.com/browsers/browsers_explorer.asp).

DarthGandalf:
I don't know web development at all, so disregard this my comment. But if it's possible to use jQuery to support more browsers, and still not bring some huge library, that would be good, I guess.

DarthGandalf:
It's hard to reason about compatibility with browsers, unless all supported browsers are actually tested (something like #772?)

ammaraskar:
>I don't know web development at all, so disregard this my comment. But if it's possible to use jQuery to support more browsers, and still not bring some huge library, that would be good, I guess.

I thought about it a bit and I have to say that znc is a technically orientated tool, I doubt there's much overlap between people with super outdated versions of IE and people who use znc. But anyway, I was initially off putted from jQuery by the discussion in the issue, namely the fact that it's so large and its use here would be so limited.
In conclusion: I feel that this is fine as it is.

>It's hard to reason about compatibility with browsers, unless all supported browsers are actually tested (something like #772?)

ehh, automated testing of browser compatibility is really overkill, I don't think there's nearly enough web code in this project to justify that

PR #868, opened 2015-02-17
fix savebuff timer initialization

Kriechi:
**Problem:**
Currently if a user loads the savebuff module it just sits there doing nothing. (read: it is not working)
This is because the timer which performs the periodical saves to disk is only added in `OnBoot` - which is not called in an already running ZNC instance.

**Solution:**
move the timer initialization to the `OnLoad` hook.
This way the timer is created when the module is loaded, no matter if ZNC is currently booting or already booted and up and running.

### DIFF
--- a/modules/savebuff.cpp
+++ b/modules/savebuff.cpp
@@ -93,13 +93,13 @@ public:
 		else
 			m_sPassword = CBlowfish::MD5(sArgs);
 
+		AddTimer(new CSaveBuffJob(this, 60, 0, "SaveBuff", "Saves the current buffer to disk every 1 minute"));
+
 		return( !m_bBootError );
 	}
 
 	virtual bool OnBoot() override
 	{
-		AddTimer(new CSaveBuffJob(this, 60, 0, "SaveBuff", "Saves the current buffer to disk every 1 minute"));
-
 		CDir saveDir(GetSavePath());
 		for (CFile* pFile : saveDir) {
 			CString sName;


jpnurmi:
The reason why OnBoot() was used instead of OnLoad(): https://github.com/znc/znc/pull/803#issuecomment-68644690

<blockquote>CIRCNetwork::ParseConfig() loads modules first, then adds configured channels. This is why savebuff loads buffers in OnIRCConnected() instead of OnLoad(). That's too late if a client connected before ZNC got connected to IRC.</blockquote>

=> OnLoad() alone is not a properly working choice either. Perhaps we have to have both.

Kriechi:
mhm. Well `OnBoot` is simply broken if you don't want to restart ZNC everytime a user wants to load savebuff...

jpnurmi:
Isn't that the main use case of savebuff? :)

Kriechi:
Sorry I wasn't clear.

If a new user wants to enable savebuff. The admin has to restart ZNC so the timer gets activated.
So eventually savebuff only works from the second restart onwards - and not immediately when the user loads the module - which is confusing and just wrong.

jpnurmi:
Oops, this was just about starting the timer, not about loading buffers. Sorry. :)

:+1: looks good then

jpnurmi:
Cherry-picked to 1.6 branch: https://github.com/znc/znc/commit/ac3570f18a870673dbfb9fbeccac8fb28f0150c1

PR #869, opened 2015-02-17
CString::WildCmp(): add an optional case-sensitivity argument

jpnurmi:
It's getting a common pattern to call AsLower() or MakeLower() on
the arguments passed to WildCmp(), we might as well add this for
convenience. It's tempting to make it case-insensitive by default,
since pretty much any IRC related comparison should be, but that
could potentially break some existing code.

### DIFF
--- a/include/znc/ZNCString.h
+++ b/include/znc/ZNCString.h
@@ -169,20 +169,26 @@ public:
 	 */
 	bool Equals(const CString& s, bool bCaseSensitive, CString::size_type uLen = CString::npos) const;
 	/**
-	 * Do a wildcard comparision between two strings.
+	 * Do a wildcard comparison between two strings.
 	 * For example, the following returns true:
 	 * <code>WildCmp("*!?bar@foo", "I_am!~bar@foo");</code>
 	 * @param sWild The wildcards used for the comparison.
 	 * @param sString The string that is used for comparing.
+	 * @param cs CaseSensitive (default) if you want the comparison
+	 *           to be case sensitive, CaseInsensitive otherwise.
+	 * @todo Make cs CaseInsensitive by default.
 	 * @return true if the wildcard matches.
 	 */
-	static bool WildCmp(const CString& sWild, const CString& sString);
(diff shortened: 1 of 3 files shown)


include/znc/ZNCString.h
```diff
@@ -174,15 +174,21 @@ class CString : public std::string {
 	 * <code>WildCmp("*!?bar@foo", "I_am!~bar@foo");</code>
 	 * @param sWild The wildcards used for the comparison.
 	 * @param sString The string that is used for comparing.
+	 * @param cs CaseSensitive (default) if you want the comparision
```
DarthGandalf: typo

include/znc/ZNCString.h
```diff
@@ -174,15 +174,21 @@ class CString : public std::string {
...
+	 *           to be case sensitive, CaseInsensitive otherwise.
+	 * @todo Make cs CaseInsensitive by default.
 	 * @return true if the wildcard matches.
 	 */
-	static bool WildCmp(const CString& sWild, const CString& sString);
+	static bool WildCmp(const CString& sWild, const CString& sString, CaseSensitivity cs = CaseSensitive);
 	/**
 	 * Do a wild compare on this string.
 	 * @param sWild The wildcards used to for the comparison.
+	 * @param cs CaseSensitive (default) if you want the comparision
```
DarthGandalf: ditto

src/ZNCString.cpp
```diff
@@ -95,9 +95,12 @@ bool CString::Equals(const CString& s, bool bCaseSensitive, CString::size_type u
 	}
 }
 
-bool CString::WildCmp(const CString& sWild, const CString& sString) {
+bool CString::WildCmp(const CString& sWild, const CString& sString, CaseSensitivity cs) {
+	const CString& sWld = (cs == CaseSensitive ? sWild : sWild.AsLower());
+	const CString& sStr = (cs == CaseSensitive ? sString : sString.AsLower());
```
DarthGandalf: I understand why references here should work. But some comment would be useful.

jpnurmi:
Fixed typos and added a clarifying comment

PR #878, opened 2015-02-19
modules/sasl.cpp: update help for "set"

Mikaela:
This is the same as #793, but for correct branch.

Closes #793


modules/sasl.cpp
```diff
@@ -70,7 +70,7 @@ class CSASLMod : public CModule {
 		AddCommand("Help",        static_cast<CModCommand::ModCmdFunc>(&CSASLMod::PrintHelp),
 			"search", "Generate this output");
 		AddCommand("Set",         static_cast<CModCommand::ModCmdFunc>(&CSASLMod::Set),
-			"username password", "Set username and password for the PLAIN mechanism");
+			"<username> [<password>]", "Set username and password for the mechanisms that need them. You can also omit the password to only set username");
```
jpnurmi: How about replacing "you can also omit ...." by "password is optional" to keep it short? :)

Mikaela: Done

PR #881, opened 2015-02-23
improve stickychan

Kriechi:
* save registry on every stick/unstick action
  * Currently the registry is only saved in the destructor - which might not get called if ZNC is crashes or gets terminated due to other reasons. 
  * This PR changes the behaviour so that after every stick or unstick action the registry is saved. Therefore state changes are immediately reflected in the `.registry` file.

* auto-save if channel key changes
  * Currently stickychan runs into troubles if a channel key changes and the user does not apply the same change again to stickychan (with the new key). 
  * This PR monitors channel mode changes and updates or deletes the key if needed.

Overall I think this makes stickychan much more reliable and useful.

### DIFF
--- a/modules/stickychan.cpp
+++ b/modules/stickychan.cpp
@@ -53,6 +53,20 @@ public:
 		return CONTINUE;
 	}
 
+	virtual void OnMode(const CNick& pOpNick, CChan& Channel, char uMode, const CString& sArg, bool bAdded, bool bNoChange) override {
+		if (uMode == CChan::M_Key) {
+			if (bAdded) {
+				// We ignore channel key "*" because of some broken nets.
+				if (sArg != "*")
+				{
+					SetNV(Channel.GetName(), sArg, true);
+				}
+			} else {
+				SetNV(Channel.GetName(), "", true);
+			}
+		}
+	}
+
 	void OnStickCommand(const CString& sCommand)
 	{
 		CString sChannel = sCommand.Token(1).AsLower();
@@ -60,7 +74,7 @@ public:
 			PutModule("Usage: Stick <#channel> [key]");
 			return;
 		}
-		SetNV(sChannel, sCommand.Token(2));
+		SetNV(sChannel, sCommand.Token(2), true);
 		PutModule("Stuck " + sChannel);
 	}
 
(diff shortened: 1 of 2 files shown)


jpnurmi:
:+1: imho can be cherry-picked to 1.6.x

DarthGandalf:
There was some network which sometimes sends `*` instead of key... Look at comments in `CChan::ModeChange` for details.

Kriechi:
@DarthGandalf you mean: https://github.com/znc/znc/blob/70c0ffb10b801456c51ca301e6d1da3f2af2dd54/src/Chan.cpp#L381-384

I've updated my code. Not sure what the best way to handle broken networks is, but I had to decide something. If the network only sends back `*` as new key, stickychan removes the key from the registry. This way the user gets a message from the network (hopefully) if the join fails at some point.

jpnurmi:
Why not just ignore the "*" like in Chan.cpp?

Kriechi:
@jpnurmi 
That was the other possibility - I did not choose.
If the key changed - stickychan would still have the wrong key in the registry.
It's either "the wrong key" or "no key"...
Either way - the manual intervention by the user is required to resolve this.

Would you like me to change it?

jpnurmi:
I don't know what are the exact circumstances when an IRCd might want to hide the channel key like that, but I'd expect it not to be a real key change, because clients would have no chance to keep up to date. It must be a less important mode reply or such. Therefore I'd just ignore the hidden key and keep the stored key intact.

Kriechi:
@jpnurmi as requested. done.

Would be great if somebody could remember one of those broken networks - I did not come across one yet...

jpnurmi:
Relevant commits:
- chansaver: Ignore channel key "*" https://github.com/znc/znc/commit/8a690170e3a9f4e7f22602596689f01050cce53a

> There are some nets out there which display the current channel key as "\*" if
> you are already in the channel. This means ZNC would remember the channel key as
> "\*" and thus be unable to rejoin the channel. Bad.

- Fix some channel key-related mess ups. https://github.com/znc/znc/commit/fe44108d90d21823db484a0998b1671c42b5a3b5

> The reason for the last change: Stupid quakenet. (RFC, anyone?)
> In /mode replies quakenet always shows '\*' as channel key, if one is set.
> This whole patch started with someone reporting that ZNC kept forgetting the
> current channel key and used '\*' instead...

PR #882, opened 2015-02-23
ClientCommand: C++11 range-based for loops

jpnurmi:


src/ClientCommand.cpp
```diff
@@ -89,20 +89,20 @@ void CClient::UserCommand(CString& sLine) {
 		Table.AddColumn("Ident");
 		Table.AddColumn("Host");
 
-		for (map<CString,CNick>::const_iterator a = msNicks.begin(); a != msNicks.end(); ++a) {
+		for (auto it : msNicks) {
```
DarthGandalf: const auto&

jpnurmi: Done

PR #889, opened 2015-02-25
Replace some C++98isms with C++11isms (#816)

jpnurmi:


DarthGandalf:
Thinking about override more... May you please remove "virtual" from
methods with "override" while you're doing this?
It makes it easy to "grep virtual" to find places where virtual hierarchy
starts.

2015-02-25 0:28 GMT-08:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
Good idea - done

md-5:
Interesting, I've always just used lowercase null rather than nullptr.

DarthGandalf:
@md-5 what is lowercase null? Uppercase `NULL` is older way to specify null pointer, `nullptr` is the new way.

modules/autocycle.cpp
```diff
@@ -31,7 +31,7 @@ class CAutoCycleMod : public CModule {
 
 	virtual ~CAutoCycleMod() {}
```
DarthGandalf: This could be override too :)

modules/keepnick.cpp
```diff
@@ -138,7 +138,7 @@ class CKeepNickMod : public CModule {
 		m_pTimer = NULL;
```
DarthGandalf: nullptr?

DarthGandalf:
Did you use clang-modernize for this?

jpnurmi:
Yes, clang-modernize was used. Interesting how it missed some NULL's.

PR #891, opened 2015-02-26
Fix dead assign/init found by clang static analyzer

jpnurmi:
Value stored to 'pUser' (during its initialization) is never read


DarthGandalf:
Hm, why not 1.6.x?

jpnurmi:
I didn't bother because this is not an important change at all :)

Issue #506, opened 2014-03-07
Suggestion: introduce dedicated message/command types


jpnurmi:
Instead of passing around raw string lines or a variable amount of tokens, dedicated message and command types could offer a convenient API to access & modify the relevant attributes of "lines" transferred between IRC servers, ZNC, and the clients. This could potentially help making the module hook API simpler and to avoid the need of ever parsing "things" twice (see for instance https://github.com/znc/znc/pull/501).

Message (received from an IRC server) attributes:
* sender/prefix/nick/ident/host
* type
* command/number
* parameters
* timestamp
* tags
* raw data (?)
* ...

The buffer storage could store these message types so the same attributes could be accessed later eg. during the buffer playback or whatever modules might want to do.

Command (sent to an IRC server) attributes:
* type
* parameters
* raw data (?)
* ...

jpnurmi:
#1000

PR #892, opened 2015-02-26
More C++11 range-based for loops

jpnurmi:


src/Template.cpp
```diff
@@ -81,15 +81,15 @@ CString CTemplateLoopContext::GetValue(const CString& sName, bool bFromIf) {
...
 
 CTemplate::~CTemplate() {
-	for (map<CString, vector<CTemplate*> >::iterator it = m_mvLoops.begin(); it != m_mvLoops.end(); ++it) {
-		vector<CTemplate*>& vLoop = it->second;
-		for (unsigned int a = 0; a < vLoop.size(); a++) {
-			delete vLoop[a];
+	for (const auto& it : m_mvLoops) {
+		const vector<CTemplate*>& vLoop = it.second;
+		for (CTemplate* pTemplate : vLoop) {
+			delete pTemplate;
```
DarthGandalf: Next step: switch this to std::unique_ptr and remove such cleanup lines :)

src/znc.cpp
```diff
@@ -1955,10 +1935,8 @@ void CZNC::ResumeConnectQueue() {
 
 void CZNC::AddNetworkToQueue(CIRCNetwork *pNetwork) {
 	// Make sure we are not already in the queue
-	for (list<CIRCNetwork*>::const_iterator it = m_lpConnectQueue.begin(); it != m_lpConnectQueue.end(); ++it) {
-		if (*it == pNetwork) {
-			return;
-		}
+	if (std::find(m_lpConnectQueue.begin(), m_lpConnectQueue.end(), pNetwork) != m_lpConnectQueue.end()) {
```
DarthGandalf: More modern way of writing this would be `std::begin(m_lpConnectQueue)` and `std::end(m_lpConnectQueue)`. I don't know what would be the benefit though.

src/Template.cpp
jpnurmi: It gets clumsy to keep backwards compatible ```std::vector<T*> GetXxx()``` -like methods if we store ```std::vector<std::shared_ptr<T>>``` internally. Are you willing to allow such API breakages? I'm also interested in #506 but that implies some bigger API changes as well.

PR #894, opened 2015-02-26
add homebrew locations for openssl, icu4c, python3

Kriechi:
improves building on Mac OS X:

* icu4c is auto-detected if installed via Homebrew
* openssl is auto-detected if installed via Homebrew
* python3 is auto-detected if installed via Homebrew

The default is now to use openssl from Homebrew if available, as fallback it uses the system openssl.
If the environment variable `USE_SYSTEM_OPENSSL` is `"true"` then the system openssl is always used, no matter if a Homebrew version is installed.

Travis now tests using the Homebrew OpenSSL and the system OpenSSL on OS X.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ matrix:
           env: BUILD_TYPE=tsan
         - os: osx
           compiler: gcc
+    include:
+        - os: osx
+          compiler: clang
+          env: BUILD_TYPE=normal USE_SYSTEM_OPENSSL=true
+
 before_install:
     - "echo os: [$TRAVIS_OS_NAME] build: [$BUILD_TYPE]"
     - if [[ "$BUILD_TYPE" == "normal" ]]; then export CFGFLAGS= MYCXXFLAGS= MYLDFLAGS=; fi
@@ -50,10 +55,8 @@ install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew config; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew list --versions; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install swig python3 icu4c jq; fi
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install swig python3 icu4c jq openssl; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew info --json=v1 --installed | jq .; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$(dirname $(find $(brew --cellar python3)/$(brew info --json=v1 python3 | jq -r '.[].installed | .[].version' | tail -n1) -name python3.pc)); fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$(dirname $(find $(brew --cellar icu4c  )/$(brew info --json=v1 icu4c   | jq -r '.[].installed | .[].version' | tail -n1) -name icu-uc.pc )); fi
     - "echo pkg-config path: [$PKG_CONFIG_PATH]"
 script:
     - ./bootstrap.sh
--- a/configure.ac
+++ b/configure.ac
@@ -93,6 +93,26 @@ case "${host_os}" in
 	;;
 	darwin*)
 		ISDARWIN=1
+
+		AC_ARG_VAR([USE_SYSTEM_OPENSSL],[set USE_SYSTEM_OPENSSL="true" to use the OpenSSL version provided by Mac OS X. Otherwise the Homebrew installed version is used if available. ])
+
+		AC_CHECK_PROG([BREW],brew, brew)
+		if test x$BREW = xbrew; then
+			dnl add default homebrew paths
+
+			icu4c_prefix=`$BREW --prefix icu4c`
+			export PKG_CONFIG_PATH="$icu4c_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
+
+			python3_prefix=`$BREW --prefix python3`
+			python3_prefix=`find "$python3_prefix/" -name python3.pc`
+			python3_prefix=`dirname "$python3_prefix"`
+			export PKG_CONFIG_PATH="$python3_prefix:$PKG_CONFIG_PATH"
+
+			if test -z "$USE_SYSTEM_OPENSSL" -o "$USE_SYSTEM_OPENSSL" = "false"; then
+				openssl_prefix=`$BREW --prefix openssl`
+				export PKG_CONFIG_PATH="$openssl_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
+			fi
+		fi
 	;;
 esac
 
@@ -321,7 +341,7 @@ if test "x$SSL" != "xno"; then
 			SSL=no
 			SSL_TEXT="no (openssl not usable)"
 		])
-		
+
 	fi
 
 	if test "x$SSL" = "xno" ; then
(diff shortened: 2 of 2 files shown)


Mikaela:
How about LibreSSL? :)

I think it's also in homebrew.

Kriechi:
@Mikaela since ZNC was not yet tested against LibreSSL so I didn't think about it.
Not sure if @DarthGandalf wants to support other TLS implementations (openssl-compatible)?

kerio92:
libressl works, but it needs an extra header to be included in csocket or it won't compile

Mikaela:
> it needs an extra header to be included in csocket or it won't compile

What kind of header? Does it break OopenSSL or anything else? If not, it should probably be PRed to Csocket.

Kriechi:
Additional headers could be added via configure as well.
If the ZNC maintainers officially want to support libressl I'm happy to work on a second PR.
Either way - out of the scope of this PR.

Mikaela:
I opened https://github.com/jimloco/Csocket/issues/47

Mikaela:
On .configure, I think you should use `$(command)` instead of \`command\` as that is what seems to be done on other parts of that file.

See also:
* http://mywiki.wooledge.org/BashFAQ/082

Kriechi:
@Mikaela thanks - I changed it accordingly.

DarthGandalf:
Proper openssl (like one from homebrew) is already tested in travis as part of linux test.
But osx has a different openssl itself, so I made travis to test that broken openssl instead of homebrew's one.

.travis.yml
```diff
@@ -50,10 +50,8 @@ install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew config; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew list --versions; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install swig python3 icu4c jq; fi
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install jq swig python3 icu4c openssl; fi
```
DarthGandalf: After this change `jq` is not needed anymore

Kriechi: I was not sure - because in the next line you use it print all installed versions from brew...
Not sure if it any benefit...

Kriechi:
ok - so **not** installing openssl via homebrew is ok for you?
or do you always want to configure with the OSX openssl?

Kriechi:
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/znc.rb#L26

so the Homebrew ZNC formula currently depends on openssl (from homebrew), not sure how that is working...

DarthGandalf:
You can add another travis build, with a different configurations, to test more (libressl/openssl from homebrew/etc). But I'm not sure it's very different from what happens on linux.
Stock openssl of OSX needs to be tested, somehow.
As for formula, I believe, it autoadds those paths itself while building ZNC from inside brew.

Mikaela:
>  As for formula, I believe, it autoadds those paths itself while building ZNC from inside brew.

Someone (@Kriechi?) was on \#znc earlier today asking how to build ZNC using brewed OpenSSL and I also think that it's better to install from there than possibly very old OS X's version of OpenSSL. I also got the picture that they wanted to compile ZNC by themselves possibly with custom modules which means compiling outside of Homebrew.

Kriechi:
I changed the Travis build matrix to do two jobs on OS X:
* default openssl
* openssl via Homebrew

I checked the Homebrew formula for ZNC, and yes it does automatically add homebrew-openssl paths.

As @Mikaela pointed out, this PR is of use if somebody wants to compile ZNC from the repo or their own fork. It is better to include the homebrew openssl because the OSX version is deprecated and might get removed with the next OSX release. And it creates lots of nasty warnings during compilation.

DarthGandalf:
@Mikaela `$()` doesn't work in `sh`. \`\` does.

configure.ac
```diff
@@ -93,6 +93,22 @@ case "${host_os}" in
...
+		if test x$BREW = xbrew; then
+			dnl add default homebrew paths
+
+			icu4c_prefix=$($BREW --prefix icu4c)
+			export PKG_CONFIG_PATH="$icu4c_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
+
+			python3_prefix=$(dirname "$(find "$($BREW --prefix python3)/" -name python3.pc)")
+			export PKG_CONFIG_PATH="$python3_prefix:$PKG_CONFIG_PATH"
+
+			if test -z "$USE_SYSTEM_OPENSSL" -o "$USE_SYSTEM_OPENSSL" = "false"; then
```
DarthGandalf: I don't think configure should depend on travis.

.travis.yml
```diff
@@ -50,10 +55,8 @@ install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew config; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew list --versions; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install swig python3 icu4c jq; fi
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install swig python3 icu4c jq openssl; fi
```
DarthGandalf: Just don't install openssl if it's not needed

Mikaela:
> @Mikaela $() doesn't work in sh. `` does.

```sh
mikaela@terasar ~
% sh
$ echo Hi $(whoami)
Hi mikaela
```

.travis.yml
Mikaela: As OpenSSL is required for SSL connections, I would consider it as quite needed.

Kriechi: openssl is required by python3 anyway.
Better make it explicit it should be installed.
(and it saves a conditional command for the second test run on osx)

configure.ac
Kriechi: IMHO this does not depend on travis.
If the user want to compile ZNC using the system openssl `USE_SYSTEM_OPENSSL` can be used in the same way.
Maybe I should add a description line to the configure script so this is documented somewhere?

Kriechi:
@DarthGandalf @Mikaela yes `sh` should support `$(...)`.

```sh
% ls -lah /bin/sh
-r-xr-xr-x  1 root  wheel   1.2M Jan 31 12:57 /bin/sh
% /bin/sh
sh-3.2$ echo Hi $(date)
Hi Fri Feb 27 10:21:28 GMT 2015
```
(tested on OSX with a real `sh`)

```sh
% ls -lah /bin/sh
lrwxrwxrwx 1 root root 4 Mar  1  2012 /bin/sh -> dash
% /bin/sh
$ echo Hi $(date)
Hi Fri Feb 27 11:06:38 CET 2015
```
(tested on Debian 7.8 with `dash`)

see http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03
see http://unix.stackexchange.com/a/126928 (third excerpt box)

Kriechi:
I added a description for the `USE_SYSTEM_OPENSSL` variable on OSX.

```
$ ./configure -h
[...]
Some influential environment variables:
[...]
  USE_SYSTEM_OPENSSL
              set USE_SYSTEM_OPENSSL="true" to use the OpenSSL version
              provided by Mac OS X. Otherwise the Homebrew installed version
              is used if available.
```

DarthGandalf:
Well, dash is not real sh.
If sh is just an alias for bash/dash, of course bash/dash understands $().
Plain sh does not, however.

2015-02-27 2:38 GMT-08:00 Thomas Kriechbaumer <notifications@github.com>:

DarthGandalf:
AFAIR, some BSD had a real sh

configure.ac
DarthGandalf: Ok. But please add some ifs for ssl/python/icu: if they are installed in homebrew, put some message that using version from homebrew.
Makes debugging easier.

Kriechi:
@DarthGandalf still not convinced, `configure.ac` already contains `$(...)`
https://github.com/znc/znc/blob/master/configure.ac#L166

would you like me to change it?

configure.ac
Kriechi: Not sure how to do that. I only add the pkg-config path here. If PKG_CHECK_MODULES finds it in there is a different question...
Maybe we should just print a summary of found libs at the end (with some nicer formatting)?
```
echo $openssl_LIBS
echo $python_LIBS
echo $icu_LIBS
```

Kriechi:
I changed the `$(...)` back to use backticks.
I added a short summary at the end of the configure output to indicate the found & used library paths.

DarthGandalf:
Oh, that was overlooked maybe. We just don't have enough tests for such broken systems.

Issue #748, opened 2014-11-12
Proposal: abandon the even-odd versioning system


jpnurmi:
ZNC has a special even-odd versioning system, where odd numbers are so called development versions, and even numbers are stable releases. This versioning system worked with version numbers like `0.096, 0.098, 0.202, 0.204, 0.206` because people didn't have expectations for such running _beta-like_ version numbers. In other words, the version numbers didn't have a strong significance - they looked more like version control revision numbers.

Things have changed when ZNC switched to the `MAJOR.MINOR` versioning system. Starting from version `1.0`, people have had certain expectations. Based on the version number, it looks like ZNC is following the industry standards. People get confused when they find out that this and that version never was/will be released. This confusion caused by ZNC's even-odd system keeps coming up on #znc, again and again.

The value added by the even-odd system is questionable. It's a source of confusion, and wastes good minor version numbers for no apparent reason. Any unreleased version should be considered a "development version". With ZNC it should be clear anyway, because if ZNC is built from Git, the version number has a git-tag together with a SHA1. Nightlies have a nightly-tag together with a date. Why is that not clear enough? Are you afraid that people will create their own releases with `make-tarball.sh` and pretend that they have an official version?

PS. I'd imagine that the even-odd system gets even more confusing with the upcoming `MAJOR.MINOR.PATCH` versioning system...

Stanzilla:
+1

jpnurmi:
How about the weechat -style versioning? The more I think about it, the more I like it.

* 1.7.0 // a released & tagged version
* 1.7-git-12-3456789 // a dev build from git
* 1.7-nightly-2015-02-19-3456789 // a nightly build
* (1.7 // a "self-made" tarball build if not using a released & tagged revision)

Mikaela:
:+1:

I am also WeeChat user :)

James-TR:
:goberserk: :goat:

Mikaela:
Is there some reason for this issue to be still open?

jpnurmi:
Was something decided? I haven't seen any release using odd version number. :)

Mikaela:
.Oh, I see. I am just assuming that `1.7.x` will become `1.7.0` in the distant future.

DarthGandalf:
1.6.1 was released.

PR #901, opened 2015-02-28
Display development version as "1.6.x" - ref #748

jpnurmi:


jpnurmi:
Poll: would you prefer ```1.6.x-git-19-bd044cf``` or ```1.6-git-19-bd044cf```?

Mikaela:
I think the first one as the .x would also be important information.

DarthGandalf:
`x` is fine. But won't `-1` in `VERSION_PATCH` break anything?

jpnurmi:
ZNC itself doesn't use ```VERSION_PATCH``` for anything, at least yet. :) I was considering making it a large number instead. I'm feeling a bit unsure, what would you prefer? Would it be a lie to claim that the current 1.6.x development version is greater than the upcoming 1.6.1? On the other hand, unsigned comparison for -1 would also give "wrong" result.

DarthGandalf:
And this is reason why (earlier?) we use(d?) odd/even versions :)

2015-02-28 8:14 GMT-08:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
ZNC is also the first project I've ever seen use .x :P Others just bump the version up by one as soon as the previous release has been tagged.

DarthGandalf:
It is you how changed output of `--version` to show `.x` :P
I did that only for branch name, which is not shown anywhere except git

Issue #270, opened 2013-01-03
Fix warnings of -Weffc++ [$5]


maconga:
There are lots of warnings when -Weffc++ is enabled
##

PR #903, opened 2015-02-28

jpnurmi:


include/znc/Client.h
```diff
@@ -33,8 +33,7 @@ class CClient;
 
 class CAuthBase {
 public:
-	CAuthBase(const CString& sUsername, const CString& sPassword, CZNCSock* pSock) {
-		SetLoginInfo(sUsername, sPassword, pSock);
```
DarthGandalf: This function is not used anywhere anymore

jpnurmi: yeah and overriding a virtual method that is called from the constructor probably didn't do what was the intention. :) so it's ok to remove? or is this something that modules could potentially use somehow?

jpnurmi:
:-1: crash in tests :o

jpnurmi:
I'll get back to this when I have time to debug what goes wrong.

PR #906, opened 2015-02-28

jpnurmi:


jpnurmi:
Better take it in steps. This is a simple low-risk subset of the original changes: https://github.com/znc/znc/pull/903

PR #907, opened 2015-03-01

jpnurmi:


DarthGandalf:
Is there a need to explicitly initialize all default-constructed fields?

jpnurmi:
It feels a bit silly that -Weffc++ complains about any missing member. We don't necessarily need to blindly "fix" all those, though. It's a bit challenging to cleanup the warnings anyway, because Csocket.h gets included to so many units: jimloco/Csocket#48.

Issue #433, opened 2013-11-22
reserved identifier violation [$5]


elfring:
I would like to point out that identifiers like "[`_WEBMODULES_H`](https://github.com/znc/znc/blob/3e56f093f2c21cd9ea72452773f9fc6a8fdba481/include/znc/WebModules.h#L17)" and "[`_SQL`](https://github.com/znc/znc/blob/b2dcad5fd49e6c0d604de38c311db0e08ed061a7/include/znc/ZNCString.h#L27)" [do not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL32-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL32-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29) to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?
##

denravonska:
What about [#pragma once](https://en.wikipedia.org/wiki/Pragma_once)? While not standard, just about every compiler supports it. And it's faster than preprocessor directives.

PR #908, opened 2015-03-01
Fix reserved identifier violation in header guards (#433)

jpnurmi:


jpnurmi:
It could be also more exlusive ```ZNCFOO_H```, but then config.h vs. zncconfig.h would conflict.

kylef:
Perhaps it would make more sense to move to `pragma once`. Since it's supported on all compilers that support C++11 from what I understand

DarthGandalf:
`ZNC_FOO_H_` works

jpnurmi:
Updated

Kriechi:
Actually some files already use `#pragma once` - would be great to stick to one syntax.
https://github.com/znc/znc/blob/9814f2a258356356f8d3d9a527ee28c370be7085/modules/modpython/ret.h#L17
https://github.com/znc/znc/blob/9814f2a258356356f8d3d9a527ee28c370be7085/modules/modperl/pstring.h#L17
https://github.com/znc/znc/blob/9814f2a258356356f8d3d9a527ee28c370be7085/modules/modperl/module.h#L17
https://github.com/znc/znc/blob/9814f2a258356356f8d3d9a527ee28c370be7085/modules/modpython/module.h#L17

I probably prefer the pragma - because it does not require a "terminating" line at the end.

PR #912, opened 2015-03-02
SASL: add webadmin interface (closes #910)

### LINKED ISSUE #910: SASL: add webadmin interface (like *perform/*cert/etc. have)
This has been asked often and it shouldn't be so complex as the module isn't so complex either. It would probably just need:

* Username
* Password
* Mechanism
* RequireAuth
    * I am not so sure on this as `no` is sane default (applies to mechanism (plain) too though.) and someone could enable this without having account on the network etc.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9105957-sasl-add-webadmin-interface-like-perform-cert-etc-have?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:


jpnurmi:
Btw, looks like this: http://i.imgur.com/jZ69eyU.png

Mikaela:
How do I specify which order I want to try mechanisms in? Like when we get https://github.com/znc/znc/issues/746, how do I select do I want to use ECDSA and then fallback to EXTERNAL or PLAIN or attempt EXTERNAL first?

jpnurmi:
:-1: I forgot the order. Should it be turned to a text field? :(

Mikaela:
I don't like idea of text field. Could it be something where the mechanisms can be dragged and dropped into preferred order?

jpnurmi:
I'll leave that to someone who's a bit more comfortable with web development.

Mikaela:
:(

Mikaela:
Couldn't this be temporary solution for some versions until someone has skills to make the option? ECDSA --> EXTERNAL --> PLAIN looks sane order to me from the newest to oldest and if you don't want to use PLAIN, you simply don't specify password and for not using EXTERNAL you don't use CertFP.

jpnurmi:
Makes sense to me. PR restored.

Mikaela:
I am still wondering fate of this PR, is there some reason why this hasn''t been merged?

jpnurmi:
See https://github.com/jpnurmi/znc/pull/1. Does re-ordering EXTERNAL and PLAIN make sense? Even when you throw ECDSA into the mix, does it still make sense? Shouldn't the order be from strong to weak (ECDSA->EXTERNAL->PLAIN), where any of them can be optionally turned off?

Mikaela:
As they all are disabled by default, I believe the oder should be

* ECDSA if user has key and wishes to use it. This is supported nowadays everywhere where Atheme is used.
* EXTERNAL if user has SSL certificate and wishes to use it. This is supported everywhere using Atheme or Anope and OFTC Services and probably many other services.
* PLAIN if user just wishes to use username or password. This is supported everywhere where SASL is supporte.

Was there some argument for changing the order?

PR #918, opened 2015-03-04
Fix warnings of -Weffc++ [-Weffc++] (#270)

jpnurmi:


include/znc/Socket.h
```diff
@@ -153,6 +158,11 @@ class CSockManager : public TSocketManager<CZNCSock> {
 	};
 	class CDNSJob : public CJob {
 	public:
+		CDNSJob() : sHostname(""), task(nullptr), pManager(nullptr), bBind(false), iRes(0), aiResult(nullptr) {}
```
DarthGandalf: I'm impressed why `: CJob(),` is not required

include/znc/Modules.h
```diff
@@ -356,7 +359,7 @@ class CModCommand {
 class CModule {
 public:
 	CModule(ModHandle pDLL, CUser* pUser, CIRCNetwork* pNetwork, const CString& sModName,
-			const CString& sDataDir);
+			const CString& sDataDir, CModInfo::EModuleType eType = CModInfo::NetworkModule);
```
DarthGandalf: Don't you need to update modperl/modpython?

jpnurmi: Gah, I don't want to touch them... I got some strange type errors for the enum argument. This is compatible since it has the default value.

include/znc/Socket.h
jpnurmi: I'll take it that you wanted it there :P

DarthGandalf: No I don't. I don't even want this boilerplate of listing every variable

include/znc/Modules.h
DarthGandalf: strange type errors?

jpnurmi: My bad. I probably put the argument in wrong place on first attempt. :P

DarthGandalf: So this default value is not needed anymore? :)

jpnurmi: Would be nice to keep it for source compatibility reasons for now

DarthGandalf: For now? Ok :)
Then add a TODO when it is going away

jpnurmi: I'm still dreaming about the day we start seriously breaking things and putting the API in shape :P

jpnurmi:
```
// TODO: remove default value in ZNC 2.x
```

PR #929, opened 2015-03-21
Refactor config parse

Kriechi:
This PR tries to reduce dangerous side effects of config parsing and `/znc rehash` invocation.

* remove very old ZNC settings and config parsing workarounds 
  * everything *before* `v0.206` - meaning configs older than 3.5 years are invalid with the next release
  * this affects: 
    * ispoof: renamed to identfile in 0.200 (51ac51a25)
    * config files without version string (introduced in 0.203)
* `/znc rehash` now only reloads the global config parts
  * includes all global modules, listeners, and other global settings
  * excludes reloading of users and all of their options
  * this should make it *less dangerous*
* `OnAddUser` hook is now only called if actually a new user is added
  * does not get called during ZNC startup any more (which used the same method)

During the refactor I tried splitting the parsing and config loading into smaller methods to improve readability a bit. I would have split it into independent PRs but was easier this way, since it is all connected to some extent. Let me know if I removed too much - or missed anything.

Bonus:
I changed a few for-loops to the new C++11 syntax for sub-config parsing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ __pycache__
 
 # Compiled Object files
 *.o
+*.o-*
 
 # Compiled Dynamic libraries
 *.so
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,15 @@
+# (unreleased)
+
+## Changes
+
+* workarounds during config parsing for old ZNC versions removed
+    * incompatible configs give an error during startup indicating the problem
+    * versions including 0.206 and newer are still supported
+* rehash only reloads global settings, including global modules and listeners
+    * users are not reloaded any more - which makes rehash less dangerous
+* OnAddUser hook is only called if actually a new user added
+    * it is not called during ZNC startup any more
+
 # ZNC 1.6.0 (2015-02-12)
 
 ## New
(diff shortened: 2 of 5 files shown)


Mikaela:
I don't think this is OK, because of some specific distributions shipping pre-1.0 versions.

Kriechi:
@Mikaela if you talk about Debian:
wheezy (stable) currently ships 0.206
jessie (currently testing, soon to be moved to stable) ships 1.4

So the next release of ZNC (with this PR) will probably not be shipped with Debian in the next couple of years.

So Debian users should have a valid upgrade path - without any problems.

DarthGandalf:
You'll be surprised to learn how many people don't upgrade to new debian
when new stable is released, and stick to "oldstable". Especially, if there
is LTS now...
Dropping everything before 0.206 is fine already though.

2015-03-21 18:59 GMT+00:00 Thomas Kriechbaumer <notifications@github.com>:

Kriechi:
Done.
* updated the PR + description.
* added a commit which clearly indicates old config compatibility parts
   * so in the future one can easily revert that

DarthGandalf:
Much of legacy compatibility code is inside CUser, e.g. for migration of `Server = ...` to `<Network default>`. You didn't touch that part?

Kriechi:
I didn't touch CUser right now - maybe that's better done in a second PR after this one (if accepted).

PR #938, opened 2015-04-04
sasl: enable all mechanisms by default

Mikaela:


Mikaela:
Conflict should be fixed now.

teward:
I have a question: what auth mechanisms are left enabled by default, then, at this point?  If BLOWFISH and others are also disabled, that effectively means SASL has no default auth mechanisms.

Mikaela:
@teward I think that you misunderstand this issue. ZNC currently attempts SASL PLAIN even without username/password being specified or at least it reports you that `PLAIN mechanism failed`.

ZNC 1.7.x where this pull request is to also has web interface for \*sasl where you can enable PLAIN at the same time as you write username/password. Setting the mechanism is also documented in the [example at wiki.](http://wiki.znc.in/Sasl#Example)

Mikaela:
The discussion from there seems quite confusing, but this seems the relevant part:

```
[10:45:17] <jpnurmi> who would try PLAIN first, and fallback to EXTERNAL? :)
[10:46:01] <Mikaela> someone insane
[10:46:03] <DarthGandalf> Reordering is available via *sasl?
[10:46:20] <jpnurmi> ECDSA->EXTERNAL->PLAIN vs. EXTERNAL->ECDSA->PLAIN ?
[10:46:44] <Mikaela> the first
[10:46:50] <DarthGandalf> why?
[10:47:06] <Mikaela> because if you don't want to use ECDSA, you don't enable it or specify username
[10:47:23] <DarthGandalf> Why it's not enabled by default?
[10:47:37] <jpnurmi> in the web ui you can just select the ones you want to use
[10:47:44] <Mikaela> because the only one enabled by default is PLAIN, because ECDSA/EXTERNAL require keys
[10:47:56] <DarthGandalf> PLAIN require key (password) too
[10:48:38] <DarthGandalf> psychon: you'd like to reimplement wheels yourself? :P
[10:49:24] <psychon> I'm all for inventing the wheel
[10:50:38] <Mikaela> DarthGandalf: and why is PLAIN enabled by default then?
[10:51:03] <Mikaela> Before 1.6 DH-* and PLAIN were all enabled by default and the one disabled by default was EXTERNAL
[10:51:34] <DarthGandalf> Good question
```

DarthGandalf:
Probably they all should be enabled, but if the required info is not set, the method is not attempted.

Mikaela:
Done. If it requires more than that, it's out of my skills.

PR #950, opened 2015-04-26
Implement IRCv3.2 echo-message capability on the "client side"

jpnurmi:
http://ircv3.net/specs/extensions/echo-message-3.2.html

TODO: Add echo-message support to CIRCSock when IRCds start supporting it.
Manual echoing won't be necessary for networks that support echo-message.

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -100,6 +100,7 @@ public:
 			  m_bAway(false),
 			  m_bServerTime(false),
 			  m_bBatch(false),
+			  m_bEchoMessage(false),
 			  m_bSelfMessage(false),
 			  m_bPlaybackActive(false),
 			  m_pUser(nullptr),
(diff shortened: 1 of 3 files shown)


Mikaela:
Does this mean between client and ZNC or ZNC and server? 

The TODO looks like the first, but I thought that between client and ZNC it was supported even before getting merged to IRCv3. Or has something changed with it?

jpnurmi:
> Does this mean between client and ZNC or ZNC and server?

With "client side" I mean between client and ZNC.

> The TODO looks like the first, but I thought that between client and ZNC it was supported even before getting merged to IRCv3. Or has something changed with it?

`self-message` was superceded by more generic `echo-message`. They are similar, but not same.

- https://github.com/ircv3/ircv3-specifications/commit/1bfba47843c2526707c902034b3395af934713c8
- https://github.com/ircv3/ircv3-specifications/pull/136

Mikaela:
I see.I am currently seeing `znc.in/echo-message` in my `cap ls`, but that is probably older version and if I understood this PR at all, it seems to now be just `echo-message`.

Mikaela:
My confusion was caused by

```
2015-04-26 22:32:56+0300 < jpnurmi> Mikaela: i abused the playback module for znc.in/echo-message :p
2015-04-26 22:55:13+0300 < Mikaela> jpnurmi: I was always sure that it was in core as I never installed any module which had anything to to with it :(
2015-04-26 22:55:58+0300 < jpnurmi> Mikaela: my bad. i was lazy to create yet another module just for that. at the time it wasn't looking like it would get accepted to ircv3.2
2015-04-26 22:56:45+0300 < Mikaela> at least my confusion is cleared now :)
```

PR #951, opened 2015-04-26
Update capability names as they are named in IRCv3.2.

DarthGandalf:
Old names will be available for a while, then removed.
self-message is superseded by echo-message and will be removed in future
too.

Also remove CAP CLEAR, because it's useless, and got deprecated in
IRCv3.1.

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -948,46 +948,6 @@ void CClient::HandleCap(const CString& sLine)
 	} else if (sSubCmd.Equals("LIST")) {
 		CString sList = CString(" ").Join(m_ssAcceptedCaps.begin(), m_ssAcceptedCaps.end());
 		RespondCap("LIST :" + sList);
-	} else if (sSubCmd.Equals("CLEAR")) {
-		SCString ssRemoved;
-		for (const CString& sCap : m_ssAcceptedCaps) {
-			bool bRemoving = false;
-			GLOBALMODULECALL(IsClientCapSupported(this, sCap, false), &bRemoving);
-			if (bRemoving) {
-				GLOBALMODULECALL(OnClientCapRequest(this, sCap, false), NOTHING);
-				ssRemoved.insert(sCap);
-			}
-		}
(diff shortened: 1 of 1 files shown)


src/Client.cpp
```diff
@@ -874,15 +874,24 @@ void CClient::HandleCap(const CString& sLine)
 	//TODO support ~ and = modifiers
 	CString sSubCmd = sLine.Token(1);
 
+	std::map<CString, std::function<void(bool bVal)>> mCoreCaps = {
```
jpnurmi: Is std::function not overkill for setting a boolean? :)

src/Client.cpp
```diff
@@ -948,46 +948,6 @@ void CClient::HandleCap(const CString& sLine)
 	} else if (sSubCmd.Equals("LIST")) {
 		CString sList = CString(" ").Join(m_ssAcceptedCaps.begin(), m_ssAcceptedCaps.end());
 		RespondCap("LIST :" + sList);
-	} else if (sSubCmd.Equals("CLEAR")) {
```
jpnurmi: "Also ..." in a commit message is often a good indication that it could/should be a separate commit.

src/Client.cpp
DarthGandalf: I just wanted to decrease code duplication... Now this list is listed 4 times (including CLEAR)

src/Client.cpp
DarthGandalf: Done

src/Client.cpp
jpnurmi: Yep, that's good. I was just thinking that the map value could've been a simple bool ptr/ref, but perhaps there will be caps with more complex handler functions in the future.

jpnurmi:
+1

jpnurmi:
Related/conflicting cap-notify mockup: https://github.com/jpnurmi/znc/commit/48d0c822dd013075b5ce834d166cd123a04d5d44

PR #952, opened 2015-04-27
Add a module "leavechans" that allows leaving all the channels matching ...

HeisSpiter:
...specifics names on last client disconnection on all the networks.

### DIFF
--- a/modules/leavechans.cpp
+++ b/modules/leavechans.cpp
@@ -1,0 +1,270 @@
+/*
+ * Copyright (C) 2004-2015 ZNC, see the NOTICE file for details.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Automaticaly leave chans and inform users on disconnect.
+ * Author: Pierre Schweitzer <pierre@reactos.org>
+ */
+
+#include <znc/IRCNetwork.h>
+#include <znc/Chan.h>
+#include <znc/User.h>
+
+class CLeaveChan {
+public:
+    CLeaveChan() {}
+
+    CLeaveChan(const CString& sLine)
+    {
+        FromString(sLine);
+    }
+
+    CLeaveChan(const CString& sLeftChan, const CString& sNotifyChan, const CString& sNotifyMsg) :
+               m_sLeftChan(sLeftChan),
+               m_sNotifyChan(sNotifyChan),
+               m_sNotifyMsg(sNotifyMsg) {
+        m_bLeftChan = false;
+    }
+
+    virtual ~CLeaveChan() {}
+
+    const CString& GetLeftChan() const { return m_sLeftChan; }
+    const CString& GetNotifyChan() const { return m_sNotifyChan; }
+    const CString& GetNotifyMsg() const { return m_sNotifyMsg; }
+    bool HasBeenLeft() const { return m_bLeftChan; }
+    void SetChanLeft() { m_bLeftChan = true; }
+    void ResetChanLeft() { m_bLeftChan = false; }
+
+    CString ToString() const
+    {
+        return m_sLeftChan + "\t" + m_sNotifyChan + "\t" + m_sNotifyMsg;
+    }
+
+    bool FromString(const CString& sLine)
+    {
+        m_sLeftChan = sLine.Token(0, false, "\t");
+        m_sNotifyChan = sLine.Token(1, false, "\t");
+        m_sNotifyMsg = sLine.Token(2, false, "\t", true);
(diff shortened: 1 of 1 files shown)


modules/leavechans.cpp
```diff
@@ -0,0 +1,270 @@
...
+                {
+                    CIRCNetwork* pNetwork = (*net);
+                    CChan* pChan = pNetwork->FindChan(it->second->GetLeftChan());
+                    if (pChan && pChan->IsOn())
+                    {
+                        PutModule("No one connected to ZNC, leaving " + it->second->GetLeftChan() + ".");
+                        pNetwork->PutIRC("PART " + it->second->GetLeftChan());
+                        if (it->second->GetNotifyMsg().empty())
+                        {
+                            pNetwork->PutIRC("PRIVMSG " + it->second->GetNotifyChan() + " :" + pNetwork->GetCurNick() + " automatically left " + it->second->GetLeftChan() + " due to client disconnection/timeout. Please check current requests.");
```
DarthGandalf: Yay for flood!
This is even worse than awaynick

modules/leavechans.cpp
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        m_msChans.clear();
+    }
+
+    virtual void OnClientDisconnect()
+    {
+        if (!m_pUser->IsUserAttached())
+        {
+            for (std::map<CString, CLeaveChan*>::const_iterator it = m_msChans.begin(); it != m_msChans.end(); it++)
```
DarthGandalf: Why not "auto", or even foreach loop?

modules/leavechans.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+        return true;
+    }
+
+    virtual ~CLeaveChans()
+    {
+        for (std::map<CString, CLeaveChan*>::iterator it = m_msChans.begin(); it != m_msChans.end(); it++)
+        {
+            delete it->second;
+        }
```
DarthGandalf: It's easier to store them in the map by value.

modules/leavechans.cpp
```diff
@@ -0,0 +1,270 @@
...
+    CString m_sLeftChan;
+    CString m_sNotifyChan;
+    CString m_sNotifyMsg;
+};
+
+class CLeaveChans : public CModule {
+public:
+    MODCONSTRUCTOR(CLeaveChans) {}
+
+    virtual bool OnLoad(const CString& sArgs, CString& sMessage)
```
DarthGandalf: bool OnLoad(...) override

DarthGandalf:
Why would anyone leave channels on disconnect? I thought bouncers are needed especially to not leave them?

teward:
@HeisSpiter Bouncers are typically designed such that you *don't* leave any channels.  You seem to want to be changing the actual design of what a bouncer does.  What's the actual usable case for this kind of module?

HeisSpiter:
@teward It's been designed for a really specific usage. And as more users have been requesting it, we're proposing it here.

So, basically, it's in use on an IRC network by the IRC staff to make them leave specific channels (such as: support channels) when they are not online. The purpose being to handle IRC staff who would have... poor connection. Which is the purpose of having a ZNC bouncer.
This also explains the notification which is done on line 122: it may happen that the staffer who got disconnected might have an undergoing ticket he was managing. It's here to ask the other staffers to have a look (hence the specific message).

This was the really main reason why it was designed. And finally, some of our users (not staff) started to appreciate it. I don't know their usage, though. But, I believe some other staffs may need it as well?

modules/leavechans.cpp
HeisSpiter: See usage comment below

modules/leavechans.cpp
HeisSpiter: This is really old code, before C++11 (it land before ZNC 1.X). A revamp would be doable, for sure.

modules/leavechans.cpp
HeisSpiter: Noted. If an inclusion is possible, I'll provide a fix there.

modules/leavechans.cpp
HeisSpiter: Noted. If an inclusion is possible, I'll provide a fix there.

teward:
@HeisSpiter Congrats, you've proven my point.

The module exists for a highly specific use case:  There is one IRC network where the IRC staff need users to leave specific channels when they are not online.  And only some of the users have appreciated the module on that one network.

The use case here is extraordinarily limited, for one network's requirements and needs.

My suggestion is to make this available as a third party module in a separate github repository, taking into account DarthGandalf's suggestions for code above, as well as implementation issues.  The narrow use case is great for one specific network and one specific ZNC instance or only a few, but I don't see any "global" benefit to all ZNC users - only for those on that one network.  With a literally infinite number of networks, that makes the use case for this hyper-narrow and specific, and that is a major negative point against the inclusion of the module, in my opinion.

DarthGandalf:
At least while it autospams channels on every disconnect, inclusion is not possible for sure.

teward:
If I may make a functionality suggestion, at the least, have the module have a delay after client disconnects - this would mean that, say, after 10 seconds of no clients being connected to the bouncer, then part.  This would help protect against join/part spam as a result of bad connections between the user's IRC client and ZNC itself.

(I still don't believe this should be included, but from a module code/actions structure i'll still give some suggestions)

Un1matr1x:
Since this is a realy special purpos, why not make a own github-repo for this module an link it @http://wiki.znc.in/Category:External_Modules - so if anybody is searching for such a module, they can find it quickly

PR #955, opened 2015-04-28
Fix #954: Startup failure when simple_away is loaded after awaynick

### LINKED ISSUE #954: Startup failure when simple_away is loaded after awaynick
ZNC 1.6.0 fails to start when the config file loads awaynick and then simple_away later on because the module is already loaded. Looks identical to the second report here: https://github.com/znc/znc/pull/663#issuecomment-57011375

awaynick automatically loads simple_away instead in 1.6.0. I'd expect it to either not automatically load a different module than requested (just have it do nothing) or ignore the second simple_away load because the first wasn't user-requested.

My two cents are that loading an altogether different module than requested is overzealous. Either ignore the LoadModule or fail outright, this feels like a half-baked compromise to make the awaynick removal easier to swallow, especially when it breaks otherwise-good configs. (This wasn't my config and I have no love for awaynick, kill it with fire.)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12320277-startup-failure-when-simple_away-is-loaded-after-awaynick?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:


src/User.cpp
```diff
@@ -1101,6 +1101,10 @@ bool CUser::LoadModule(const CString& sModName, const CString& sArgs, const CStr
 		CFile fNVFile = CFile(GetUserPath() + "/moddata/" + sModName + "/.registry");
 
 		for (vector<CIRCNetwork*>::iterator it = m_vIRCNetworks.begin(); it != m_vIRCNetworks.end(); ++it) {
+			if ((*it)->GetModules().FindModule(sModName)) {
```
DarthGandalf: Please add a comment.

Otherwise, LGTM

jpnurmi: Done

DarthGandalf:
Probably I accidentally rebased this instead of merging... But it's in.

PR #956, opened 2015-04-28
configure.ac: specify required compiler versions

### LINKED ISSUE #947: Don't suggest people to use ZNC 1.4
```
configure: No compiler with C++11 support was found
configure: error: Either upgrade your compiler, or use ZNC 1.4
```

Recommending 1.4 is not a good idea as it's old and unsupported version, doesn't check SSL certificates (https://github.com/znc/znc/issues/156) and is vulnerable (https://github.com/znc/znc/issues/621) to [POODLE] and [FREAK].

[POODLE]:https://en.wikipedia.org/wiki/POODLE
[FREAK]:https://en.wikipedia.org/wiki/FREAK

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11348213-don-t-suggest-people-to-use-znc-1-4?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mikaela:
and don't sugest using ZNC 1.4.

Closes #947

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,7 +32,7 @@ AC_PROG_CXX
 # "Optional" because we want custom error message
 AX_CXX_COMPILE_STDCXX_11([noext], [optional])
 if test x"$HAVE_CXX11" != x1; then
-	AC_MSG_ERROR([Either upgrade your compiler, or use ZNC 1.4])
+	AC_MSG_ERROR([Either upgrade your gcc to 4.7+ or clang to 3.2+])
 fi
 
 AC_PROG_INSTALL


Mikaela:
I am not sure if this should have went into 1.6.x so I thought that master is safe choice for PR.

Mikaela:
I forgot `to` in front of clang so I force-pushed the update. I am also not sure on other compilers, but specifying GCC and clang versions should be enough as ZNC is only tested against those two as far as I am aware. The version numbers come from issue #947 where KindOne knew the GCC requirement and clang requirement is mentioned at the project homepage.

DarthGandalf:
The requirements are listed at http://wiki.znc.in/ChangeLog/1.6.0

2015-04-28 15:05 GMT+01:00 Mikaela Suomalainen <notifications@github.com>:

Mikaela:
> Bump compiler requirements to support C++11. This means GCC 4.7+, Clang 3.2+, SWIG 3.0.0+.

Where does Clang 3.2 come from or doesn't ZNC need the whole C++11 support? http://clang.llvm.org/cxx_status.html

Mikaela:
PR updated to mentiong Clang 3.2+ instead. Thanks jpnurmi

PR #957, opened 2015-04-28
Add cap-notify & away-notify support

jpnurmi:


jpnurmi:
Nope, any unknown/unsupported network capabilities are not advertised to clients. The first commit adds `cap-notify` so clients have a way to get notified, and the second commit introduces the first network-dependent capability. I chose `away-notify` because it's the most interesting to me. If only Freenode added support for it, there was no longer need for clients to do that stupid who-polling...

Mikaela:
I am just wondering what is the status of this PR as the latest activity was 8 days ago.

DarthGandalf:
@jpnurmi If you think it's ready to merge, please do so

DarthGandalf:
I'll trust your choice...

jpnurmi:
Hi. I probably won't be active for a while.

jpnurmi:
Oops. That was an accident... :)

Mikaela:
@jpnurmi but will you merge this?

And can I ask why you won't be active for a while?

jvanegmond:
I've tested this today and this works beautifully. Using UnrealIRCd 3.2.10.4, a fresh build of the the cap-notify branch that is being merged here and HexChat 2.10.2 as client.

Mikaela:
I hope someone merges this in near future.

DarthGandalf:
Did you test cases like "first client connects, then znc", "first znc
connects, then client", "supporting server connects, then client connects,
then znc reconnects to non-supporting server", etc? With various
combinations of supporting/non-supporting the capability clients and
servers...

2015-05-28 15:13 GMT+01:00 Mikaela Suomalainen <notifications@github.com>:

jvanegmond:
I only tested with these particular versions which all support away-notify. We're running this in our environment now and don't plan to do any tests where either client or server is non-supporting. That still needs to be done.

src/Client.cpp
```diff
@@ -996,3 +1006,32 @@ void CClient::ParseIdentifier(const CString& sAuthLine) {
...
+	if (!m_ssNetworkCaps.count("away-notify") && ssCaps.count("away-notify")) {
+		ssNewCaps.insert("away-notify");
+	} else if (m_ssNetworkCaps.count("away-notify") && !ssCaps.count("away-notify")) {
+		ssDelCaps.insert("away-notify");
+		m_bAwayNotify = false;
+	}
+
+	if (!ssDelCaps.empty()) {
+		CString sCaps = CString(" ").Join(ssDelCaps.begin(), ssDelCaps.end());
+		PutClient(":irc.znc.in CAP " + GetNick() + " DEL :" + sCaps);
```
jpnurmi: missing HasCapNotify() check

src/Client.cpp
```diff
@@ -996,3 +1006,32 @@ void CClient::ParseIdentifier(const CString& sAuthLine) {
...
+	}
+
+	if (!ssNewCaps.empty()) {
+		CString sCaps = CString(" ").Join(ssNewCaps.begin(), ssNewCaps.end());
+		PutClient(":irc.znc.in CAP " + GetNick() + " NEW :" + sCaps);
```
jpnurmi: ditto

src/Client.cpp
```diff
@@ -996,3 +1006,32 @@ void CClient::ParseIdentifier(const CString& sAuthLine) {
...
+void CClient::SetNetworkCaps(const SCString& ssCaps)
+{
+	if (m_ssNetworkCaps == ssCaps) {
+		return;
+	}
+
+	SCString ssDelCaps;
+	SCString ssNewCaps;
+
+	if (!m_ssNetworkCaps.count("away-notify") && ssCaps.count("away-notify")) {
```
jpnurmi: 1019-1024 is ugly. ideas to make it prettier?

DarthGandalf: If you're talking about this exact logic, you can try something with having an array of 2 sets, and counting an index in that array; but it'll be more confusing than this.

src/Client.cpp
```diff
@@ -996,3 +1006,32 @@ void CClient::ParseIdentifier(const CString& sAuthLine) {
 		m_sUser = sAuthLine;
 	}
 }
+
+void CClient::SetNetworkCaps(const SCString& ssCaps)
```
DarthGandalf: There is some value in disabling caps on client side when server disconnects, and reenabling them again later when server connects. For client to clear cache of account-notify, so that upon reconnect of server, client can know that it knows nothing about who is on which account now.

Probably it doesn't matter for away-notify, because while network is disconnected, there are no channels.

jpnurmi:
I've refactored the PR. Reviews welcome.

include/znc/Client.h
```diff
@@ -208,6 +221,7 @@ class CClient : public CIRCSocket {
 	CString              m_sIdentifier;
 	std::shared_ptr<CAuthBase> m_spAuth;
 	SCString             m_ssAcceptedCaps;
+	SCString             m_ssNotifiedCaps;
```
DarthGandalf: Some comment what's this would help.

src/Client.cpp
```diff
@@ -19,11 +19,16 @@
 #include <znc/User.h>
 #include <znc/IRCNetwork.h>
 #include <znc/Query.h>
+#include <algorithm>
 
 using std::set;
 using std::map;
 using std::vector;
 
+static const std::map<CString, std::function<void(CClient* pClient, bool bVal)>> g_mNotifyableCaps = {
```
DarthGandalf: 1. constexpr
2. It's not member, no "m"
3. Notifiable?

DarthGandalf: Ah, "m" is map, sorry...

src/Client.cpp
```diff
@@ -19,11 +19,16 @@
...
+	{"away-notify", [](CClient* pClient, bool bVal) { pClient->SetAwayNotify(bVal); }},
```
DarthGandalf: This is very similar to mCoreCaps below. Can they look more similar too? E.g. single map with tuple of bool+function

src/Client.cpp
```diff
@@ -996,3 +1025,40 @@ void CClient::ParseIdentifier(const CString& sAuthLine) {
...
+void CClient::NotifyAvailableCaps()
+{
+	const CIRCSock *pSock = GetIRCSock();
+	if (!pSock)
+		return;
+
+	m_ssNotifiedCaps.clear();
+
+	const SCString& ssAvailCaps = pSock->GetAvailableCaps();
+	for (const CString& sCap : ssAvailCaps) {
```
DarthGandalf: you can use `for (const CString& sCap : pSock->GetAvailableCaps())` directly... I don't think attaching it to a local variable changes much.

src/IRCNetwork.cpp
```diff
@@ -1223,6 +1227,9 @@ void CIRCNetwork::SetIRCSocket(CIRCSock* pIRCSock) {
 }
 
 void CIRCNetwork::IRCConnected() {
+	for (CClient* pClient : m_vClients) {
+		pClient->IRCConnected();
```
DarthGandalf: Does this mean that `away-notify` will be announced for clients before capability negotiation between ZNC and server?
If server happens to not support it, client will already know that it doesn't need to send WHO every second, will rely on non-working away-notify offered by ZNC, and won't know away status of anyone at all.

DarthGandalf:
Did you test various corner cases as described above?

Mikaela:
Would away-notify go through ZNC properly if I added requesting it to \*perform on networks that are known to support it before this is merged?

include/znc/Client.h
jpnurmi: Done

src/Client.cpp
jpnurmi: 1. ```error: constexpr variable cannot have non-literal type 'const std::map<T1, T2>'```
2. now it is a member so it became ```m_mCoreCaps``` or should it also have the map,string,tuple,bool,func types as in ```m_mstbfCoreCaps```? :)
3. done

src/Client.cpp
jpnurmi: done

jpnurmi:
I've tried to test all possible combinations of servers and clients with and without cap-notify/away-notify support...

jpnurmi:
> I think capability shouldn't be offered to client until it got negotiated with server.

Good point, thanks for making me realize that GetAvailableCaps() was a relic from the times when I was still trying to let clients request server caps directly. Updated again.

include/znc/Client.h
```diff
@@ -111,12 +113,24 @@ class CClient : public CIRCSocket {
...
 			  m_spAuth(),
-			  m_ssAcceptedCaps()
+			  m_ssAcceptedCaps(),
+			  m_mCoreCaps({{"multi-prefix", std::make_tuple(false, [this](bool bVal) { m_bNamesx = bVal; })},
+			               {"userhost-in-names", std::make_tuple(false, [this](bool bVal) { m_bUHNames = bVal; })},
+			               {"echo-message", std::make_tuple(false, [this](bool bVal) { m_bEchoMessage = bVal; })},
+			               {"server-time", std::make_tuple(false, [this](bool bVal) { m_bServerTime = bVal; })},
+			               {"batch", std::make_tuple(false, [this](bool bVal) { m_bBatch = bVal; })},
+			               {"cap-notify", std::make_tuple(false, [this](bool bVal) { m_bCapNotify = bVal; })},
+			               {"away-notify", std::make_tuple(true, [this](bool bVal) { m_bAwayNotify = bVal; })}})
```
jpnurmi: so, should i reformat this somehow? any preferences?

DarthGandalf: Ah, it's fine. Probably std::pair instead of std::tuple would allow `{}`, but whatever...

jpnurmi:
Updated to use std::pair & {}'s, and tweaked the initializer list formatting.

src/IRCNetwork.cpp
```diff
@@ -659,6 +663,9 @@ void CIRCNetwork::ClientDisconnected(CClient *pClient) {
 	if (it != m_vClients.end()) {
 		m_vClients.erase(it);
 	}
+	if (m_pIRCSock) {
```
DarthGandalf: Why this `if`?

DarthGandalf:
Anyway, LGTM

src/IRCNetwork.cpp
jpnurmi: Thanks, cleaned up. An earlier version was querying the socket caps here.

PR #962, opened 2015-05-07
Add third class to the port input [Fixes #446]

wolfy1339:
Adding this class makes it so the input has a defined width.
Fixes znc/znc#446


bendem:
You didn't add a class, you changed the type to something invalid.

wolfy1339:
Thanks I fixed that

PR #978, opened 2015-06-21
log: Add -timestamp option (cont.)

anlutro:
Continuation of #695 - hope this is OK, this is a feature I'd really like to see added and the original PR seemed abandoned.

Changed the argument parsing to use `sArgs.Token(uIndex)` to allow any order of arguments.

Maybe the same thing is possible with `for (CString& sArg : vsArgs)` but I couldn't see how to grab the element in the next index. Let me know if this is the preferred way to do it.

Also see inline comments.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -239,7 +239,7 @@ void CLogMod::PutLog(const CString& sLine, const CString& sWindow /*= "Status"*/
 	if (!CFile::Exists(sLogDir)) CDir::MakeDir(sLogDir, ModDirInfo.st_mode);
 	if (LogFile.Open(O_WRONLY | O_APPEND | O_CREAT))
 	{
-		LogFile.Write(CUtils::FormatTime(curtime, "[%H:%M:%S] ", GetUser()->GetTimezone()) + (m_bSanitize ? sLine.StripControls_n() : sLine) + "\n");
+		LogFile.Write(CUtils::FormatTime(curtime, m_sTimestamp, GetUser()->GetTimezone()) + " " + (m_bSanitize ? sLine.StripControls_n() : sLine) + "\n");
 	} else
 		DEBUG("Could not open log file [" << sPath << "]: " << strerror(errno));
 }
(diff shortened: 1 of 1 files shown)


modules/log.cpp
```diff
@@ -268,15 +269,35 @@ CString CLogMod::GetServer()
 
 bool CLogMod::OnLoad(const CString& sArgs, CString& sMessage)
 {
+	CString sMyArgs = sArgs;
```
anlutro: Not sure if this copying is necessary, it was done this way in awaystore.cpp so I kept it like that.

modules/log.cpp
```diff
@@ -268,15 +269,35 @@ CString CLogMod::GetServer()
...
-	m_sLogPath = sArgs.Token(uIndex);
+	bool bHaveLogPath = false;
+
+	do {
+		sArg = s.sMyArgs.Token(uIndex);
+
+		if (sArg.Equals("-sanitize")) {
+			m_bSanitize = true;
+		} else if (sArg.Equals("-timestamp")) {
+			sArg = s.sMyArgs.Token(++uIndex);
```
anlutro: Can this error? If a user loads the module without a timestamp argument, i.e. `LoadModule = log -timestamp`, what will happen?

modules/log.cpp
```diff
@@ -268,15 +269,35 @@ CString CLogMod::GetServer()
...
-	m_sLogPath = sArgs.Token(uIndex);
+	bool bHaveLogPath = false;
+
+	do {
+		sArg = sMyArgs.Token(uIndex);
+
+		if (sArg.Equals("-sanitize")) {
+			m_bSanitize = true;
+		} else if (sArg.Equals("-timestamp")) {
+			sArg = sMyArgs.Token(++uIndex);
```
anlutro: Can this error? If a user loads the module without a timestamp argument, i.e. `LoadModule = log -timestamp`, what will happen?

modules/log.cpp
DarthGandalf: Well, do you need the copy?
Awaystore parses differently, it uses Token(0) every time, and chops the first token out of string every time. So it needs the string to be modifiable.

modules/log.cpp
DarthGandalf: `""`

modules/log.cpp
```diff
@@ -268,15 +269,35 @@ CString CLogMod::GetServer()
...
-		m_bSanitize = true;
-		++uIndex;
-	}
 
-	// Use load parameter as save path
-	m_sLogPath = sArgs.Token(uIndex);
+	bool bHaveLogPath = false;
+
+	do {
+		sArg = sMyArgs.Token(uIndex);
```
DarthGandalf: `.Split()` and iterating through `vector<string>` would be better

modules/log.cpp
```diff
@@ -268,15 +269,35 @@ CString CLogMod::GetServer()
...
+
+		if (sArg.Equals("-sanitize")) {
+			m_bSanitize = true;
+		} else if (sArg.Equals("-timestamp")) {
+			sArg = sMyArgs.Token(++uIndex);
+			m_sTimestamp += sArg + " ";
```
DarthGandalf: Does this mean that `-timestamp A -timestamp B` leads to `A B `?

modules/log.cpp
```diff
@@ -313,7 +334,7 @@ bool CLogMod::OnLoad(const CString& sArgs, CString& sMessage)
 		sMessage = "Invalid log path ["+m_sLogPath+"].";
 		return false;
 	} else {
-		sMessage = "Logging to ["+m_sLogPath+"].";
+		sMessage = "Logging to ["+m_sLogPath+"]. Using timestamp ["+m_sTimestamp+"]";
```
DarthGandalf: s/timestamp/timestamp format/

DarthGandalf:
Did you test this?

DarthGandalf:
> Maybe the same thing is possible with for (CString& sArg : vsArgs) but I couldn't see how to grab the element in the next index. Let me know if this is the preferred way to do it.

The preferred way would be to have a common way to parse arguments to modules, something like getopt. But ZNC doesn't currently provide such way.

jpnurmi:
I thought the plan was to kill module arguments altogether. Just make them
all configurable options?

DarthGandalf:
> I thought the plan was to kill module arguments altogether.

There are many different plans, and not all of them make sense together :)

anlutro:
Thanks for the comments, I've fixed up all the issues and tested locally. The method used for grabbing the actual timestamp format argument seems kinda primitive but I guess it works.

```
[ .. ] Loading global module [log]...
[ >> ] Logging to [/home/andreas/.znc/moddata/log/$USER/$NETWORK/$WINDOW/%Y-%m-%d.log]. Using timestamp format [%Y-%m-%d %H:%m%:%s] [/usr/local/lib/znc/log.so]
```

DarthGandalf:
@anlutro this pull request doesn't actually change any timestamp...
I'll repeat my question: Did you test this?

anlutro:
Been too long for me to remember exactly what I did. Could be something lost when I was working on the code. In any case, I fixed it up and pushed.

henbruas:
Why is this merged but not included in 1.6.2 nor in the 1.6.3 RCs?

DarthGandalf:
Because it's merged in master, not in 1.6.x.

dgw:
It was merged into master. The 1.6.x releases come from a different branch and receive only minor updates, mostly security fixes.

henbruas:
Oh, I'm assuming that means it won't be included until the next major release? Thanks for the very quick response, guess I'll build from the master branch instead of using 1.6.2.

dgw:
Correct, master will be used for the next major release (1.7). Meanwhile, [building from Git](http://wiki.znc.in/Git) is a good option if you need something that isn't in a release version yet.

Issue #960, opened 2015-05-05
Add support for IRCv3.2 message tags


3ventic:
The log module is not writing PRIVMSGs with IRCv3 tags into the log files. I'm not sure if it's specific tags causing the issue, all of the ones I'm trying to log have the same format and are from Twitch chat. Here's one such message captured with raw module:

```
[11:36:52] <*raw> IRC -> [@color=#FF0000;display-name=BrainsGames;emotes=;subscriber=0;turbo=0;user-type=;user_type= :brainsgames!brainsgames@brainsgames.tmi.twitch.tv PRIVMSG #alaskansavage :A S Hope all is well buddy!]
```

(More detailed [here](https://i.3v.fi/twitch-irc-v3.txt))

I've tried both ZNC 1.6.0 and 1.7.x-git-180-1cce7f6 with same results on Debian Jessie (8)

DarthGandalf:
Did you somehow force server to enable this capability? If not, why does server send tags at all?
ZNC doesn't yet support tags on IRC server connection.

3ventic:
Yes, by CAP REQ, because I need the information they contain.

Mikaela:
You shouldn't `cap req` anything that your client & bouncer don't support so I would say that this is invalid unless you are requesting ZNC to start supporting tags.

Mikaela:
You shouldn't `cap req` anything that your client & bouncer don't support so I would say that this is invalid unless you are requesting ZNC to start supporting tags.

3ventic:
It is more of a request to support tags. In my case, I'm running a script between ZNC and the server, so I simply [started stripping the tags before they reach ZNC](https://github.com/3ventic/tmi-irc-relay/commit/16daa616e82de02672152943a6a667d95b2f9a49). I would still like tags available on my client while having ZNC log the chat, though.

Mikaela:
> I'd also like to point out that IRCv3 working group has ZNC 0.207 or later listed as a compliant client

That is a little inaccurate as the list is IRCv3**.1** compatible clients and when making the website they removed the mentioning of .1. Tags are in IRCv3.2 and I am not sure if any client understands them yet.

* http://ircv3.net/specs/core/message-tags-3.2.html

DarthGandalf:
In order to support some capability, `cap req` is not enough. One also needs to actually add support for it (the enabled tags, other protocol changes, etc)
For proper support of tags on server side some refactoring will be needed (e.g. #506)
[...]

PR #981, opened 2015-06-24
Add support for parsing IRCv3 message tags

BtbN:
See http://ircv3.net/specs/core/message-tags-3.2.html for informations on what this is about.

While this does nothing with the tags by itself, it enables ZNC to propperly parse messages containing tags and makes them available for modules.

The tags are only available while inside of IRCSocks ReadLine function, and are automaticaly cleared again when the function is left.
It also adds convenience functions to CModule, so modules can access the tags in a more convenient way.

I'm primarily interested in this because Twitch recently started to offer additional information via tags, which can be usefull for a Twitch-Specific module.

### DIFF
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -27,6 +27,7 @@
 #include <sys/time.h>
 #include <unistd.h>
 #include <vector>
+#include <functional>
 
 static inline void SetFdCloseOnExec(int fd)
 {
@@ -345,6 +346,21 @@ protected:
 	typedef typename std::map<K, value>::iterator iterator;
 	std::map<K, value>   m_mItems;   //!< Map of cached items.  The value portion of the map is for the expire time
 	unsigned int         m_uTTL;     //!< Default time-to-live duration
+};
+
+/**
+ * @class COnDelete
+ * @author BtbN <btbn@btbn.de>
+ * @brief Helper class that calls the functor passed to its constructor when it goes out of scope
+ */
+class COnDelete {
+	COnDelete(const COnDelete&) = delete;
+public:
+	COnDelete():m_f([]() {}) {}
+	COnDelete(std::function<void()> f):m_f(std::move(f)) {}
+	~COnDelete() { m_f(); }
+private:
+	std::function<void()> m_f;
 };
 
 #endif // !ZNC_UTILS_H
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -20,6 +20,7 @@
 #include <znc/IRCNetwork.h>
 #include <znc/Server.h>
 #include <znc/Query.h>
+#include <znc/Utils.h>
 
 using std::set;
 using std::vector;
@@ -79,7 +80,8 @@ CIRCSock::CIRCSock(CIRCNetwork* pNetwork)
 		  m_iSendsAllowed(pNetwork->GetFloodBurst()),
 		  m_uFloodBurst(pNetwork->GetFloodBurst()),
 		  m_fFloodRate(pNetwork->GetFloodRate()),
-		  m_bFloodProtection(IsFloodProtected(pNetwork->GetFloodRate()))
+		  m_bFloodProtection(IsFloodProtected(pNetwork->GetFloodRate())),
+		  m_msCurrentTags()
 {
 	EnableReadLine();
 	m_Nick.SetIdent(m_pNetwork->GetIdent());
@@ -147,6 +149,15 @@ void CIRCSock::ReadLine(const CString& sData) {
 	sLine.TrimRight("\n\r");
 
 	DEBUG("(" << m_pNetwork->GetUser()->GetUserName() << "/" << m_pNetwork->GetName() << ") IRC -> ZNC [" << sLine << "]");
+
+	COnDelete clearTags([this]() {
+		m_msCurrentTags.clear();
+	});
+
+	if (sLine.StartsWith("@")) {
+		m_msCurrentTags = CUtils::GetMessageTags(sLine);
+		sLine = sLine.Token(1, true);
+	}
 
 	bool bReturn = false;
 	IRCSOCKMODULECALL(OnRaw(sLine), &bReturn);
--- a/src/Modules.cpp
+++ b/src/Modules.cpp
@@ -19,6 +19,7 @@
 #include <znc/Template.h>
 #include <znc/User.h>
 #include <znc/IRCNetwork.h>
+#include <znc/IRCSock.h>
 #include <znc/WebModules.h>
 #include <znc/znc.h>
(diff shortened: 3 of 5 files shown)


Mikaela:
Related issue https://github.com/znc/znc/issues/960 and you might want to add "Closes #960" to your pull request message or some commit message.

BtbN:
Didn't even notice there was a similar issue, added to primary commit.

DarthGandalf:
There is already some logic for parsing tags in `Utils.h`, it's not really used though.
Why do you need access to tag line?
The idea of adding the accessor functions to CModule is interesting. It's different from idea in #506 and I don't know which one is better. @jpnurmi ?
These functions work correctly only during receiving message from server, but not from client?

src/IRCSock.cpp
```diff
@@ -148,6 +150,70 @@ void CIRCSock::ReadLine(const CString& sData) {
...
+
+			if (pc == '\\') {
+				sValue += pc;
+			}
+
+			m_msCurrentTags[sKey] = sValue;
+		}
+
+		clearTagsHelper = std::unique_ptr<int, std::function<void(int*)>>(new int, [this](int *v) {
+			delete v;
```
DarthGandalf: What is this?

src/Modules.cpp
```diff
@@ -202,6 +203,52 @@ const CString& CModule::GetSavePath() const {
...
+
+CString CModule::GetCurrentTagValue(const CString &sKey) const {
+	if (!m_pNetwork || !m_pNetwork->GetIRCSock())
+		return CString();
+
+	auto map = m_pNetwork->GetIRCSock()->GetCurrentTags();
+	auto it = map.find(sKey);
+
+	if (it == map.end()) {
+		return CString();
```
DarthGandalf: `""`

src/IRCSock.cpp
BtbN: It's a helper to clear the current tags when the ReadLine function returns, as the tags are only valid in the context of the line currently beeing parsed.
So when the helper unique_ptr is deleted, the CurrentTags are cleared.

BtbN:
In my case, i need to access the tags on twitch.tv:
https://github.com/justintv/Twitch-API/blob/master/IRC.md#tags

I'm primarily interested in the display-name one, to propperly capitalize user nicknames, they send them full-lowercase except for in that tag.

BtbN:
Refactored to use the already existing parse function. Cleans up and shortens it quite a bit.

include/znc/Utils.h
```diff
@@ -347,4 +348,19 @@ class TCacheMap {
...
+/**
+ * @class COnDelete
+ * @author BtbN <btbn@btbn.de>
+ * @brief Helper class that calls the functor passed to its constructor when it goes out of scope
+ */
+class COnDelete {
+	COnDelete(const COnDelete&) = delete;
+public:
+	COnDelete():f([]() {}) {}
+	COnDelete(std::function<void()> fn):f(fn) {}
```
DarthGandalf: std::move

src/IRCSock.cpp
```diff
@@ -148,6 +151,18 @@ void CIRCSock::ReadLine(const CString& sData) {
...
 	DEBUG("(" << m_pNetwork->GetUser()->GetUserName() << "/" << m_pNetwork->GetName() << ") IRC -> ZNC [" << sLine << "]");
 
+	COnDelete cleanTags([this]() {
+		m_sCurrentTagLine = "";
+		m_msCurrentTags.clear();
+	});
+
+	if (sLine.StartsWith("@")) {
+		m_msCurrentTags = CUtils::GetMessageTags(sLine);
+		m_sCurrentTagLine = sLine.Token(0);
```
DarthGandalf: I don't really like calling Token(0) twice: here and inside GetMessageTags

src/IRCSock.cpp
DarthGandalf: Yes, that's what second type parameter for std::unique_ptr does. But you don't need int for this.

DarthGandalf:
> Why do you need access to tag line?

> These functions work correctly only during receiving message from server, but not from client?

jpnurmi:
> On 25 Jun 2015, at 02:52, Alexey Sokolov <notifications@github.com> wrote:
> 
> The idea of adding the accessor functions to CModule is interesting.

I’d like to see the CModule API being _reduced_ instead.

> It's different from idea in #506 and I don't know which one is better. @jpnurmi ?

Well, people usually like their own ideas, don’t they? :) I still believe passing the tags along in a message structure would be a good long term solution.

--
J-P Nurmi

BtbN:
Yes, a message class would be a more complete solution, but also a major api change.
I just wanted to get access to the message tags in the least disruptive way.

And i also think i already answered why i want to access the tags?
Twitch sends a bunch of information there which i want to use in a module, so yes, they only take care of the ZNC->Server part.

DarthGandalf:
Uhm. I didn't ask you why you want to access tags, that's pretty obvious.
"Tag line" is not tags. Map is not enough?

BtbN:
Oh, that's for the case some server decides to do something stupid and non-standard with the tags, so a module still has the chance to handle it.
Don't have any example for such a server though.

It also makes it easier to forward the tags to a client, without the need to re-construct and escape them.

DarthGandalf:
> It also makes it easier to forward the tags to a client, without the need to re-construct and escape them.

That's not correct. Let's say ZNC requested `cap-a` and `cap-b` from server, and they enable some tags. Client requested only `cap-a` from ZNC, and that enables only subset of tags which ZNC receives. If client doesn't support tags from `cap-b`, ZNC shouldn't send them to client. So reconstruction of tags is required.

jpnurmi:
Would it be enough for your module to override ```CModule::OnRaw()``` and do this with help of ```CUtils::GetMessageTags()``` and ```SetMessageTags()```?

BtbN:
Yes, i could get the information that way. But that would also mean that without my module, ZNC would be unable to interpret what the server is sending, as it has no idea about messages containing tags.

jpnurmi:
You would have to update the module when ZNC starts handling message tags properly, yes. To me it sounds like a better compromise than adding "temporary" API that ZNC will then have to support for years to come.

BtbN:
Wouldn't the message class change force every module to be updated anyway, so removing that API in the process wouldn't have much impact?
Not using strings for messages anymore sounds like a very api breaking change, or how do you intend to do that in a compatible way?

jpnurmi:
Sorry, but I don't see it as a valid reason to rush in sub-optimal API. #506 is a suggestion that nobody (to my knowledge) is actively working on.

BtbN:
Superseded by #1000

PR #982, opened 2015-06-24
nickserv.cpp: Add ability to specify entire NickServ mask.

ChasedSpade:
This adds the command 'SetNSMask' to the NickServ module. This command allows the user to specify the nickmask, example being '**NickServ!NickServ@services.**', that the module will check for if a message to identify is received.
 
If NSMask is not specified, it will be set to the mask of NickServ when a message to identify is received. This ensures the module will only send the identify command when requested by the NickServ service and not somebody impersonating the nick of said service.

### DIFF
--- a/modules/nickserv.cpp
+++ b/modules/nickserv.cpp
@@ -34,6 +34,11 @@ public:
 		DelNV("Password");
 	}
 
+	void SetNSMaskCommand(const CString& sLine) {
+		SetNV("NickServMask", sLine.Token(1, true));
+		PutModule("NickServ mask set");
+	}
+
 	void SetNSNameCommand(const CString& sLine) {
 		SetNV("NickServName", sLine.Token(1, true));
 		PutModule("NickServ name set");
@@ -58,7 +63,6 @@ public:
 		}
 		PutModule("Ok");
 	}
-
 	MODCONSTRUCTOR(CNickServ) {
 		AddHelpCommand();
 		AddCommand("Set", static_cast<CModCommand::ModCmdFunc>(&CNickServ::SetCommand),
@@ -67,6 +71,8 @@ public:
 			"", "Clear your nickserv password");
 		AddCommand("SetNSName", static_cast<CModCommand::ModCmdFunc>(&CNickServ::SetNSNameCommand),
 			"nickname", "Set NickServ name (Useful on networks like EpiKnet, where NickServ is named Themis)");
+		AddCommand("SetNSMask", static_cast<CModCommand::ModCmdFunc>(&CNickServ::SetNSMaskCommand), 
+			"NickServ!services@hostname.org", "Set NickServ mask (Useful for ensuring password is not sent to service impersonator(s))");
 		AddCommand("ClearNSName", static_cast<CModCommand::ModCmdFunc>(&CNickServ::ClearNSNameCommand),
 			"", "Reset NickServ name to default (NickServ)");
 		AddCommand("ViewCommands", static_cast<CModCommand::ModCmdFunc>(&CNickServ::ViewCommandsCommand),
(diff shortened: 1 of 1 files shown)


Mikaela:
I think people should use \*sasl instead and setting the mask manually just makes using the module more difficult. I am also not sure how supported this is and I believe better solution is using NICKSERV command instead of privmsg as is done with https://github.com/znc/znc/pull/833.

ChasedSpade:
I agree that  the sasl module should be used, however, some networks do not support authentication with SASL. 
Also, like the SetNSName command, it's preset to the value of NickServ's mask if the command isn't used. 
This command still uses the NICKSERV command, however, some network configurations simply set the command to be **/PRIVMSG NickServ**, and do not specify a host for NickServ. This prevents the module from sending the identify command to somebody impersonating NickServ.

teward:
Configuration brings a question to mind.  How are users supposed to know the full mask of services to configure it?  If there is no explicitly available documentation available for that, the functionality has no use since you have no configuration information...

teward:
There was additional discussion on this in IRC in #znc.

Timestamps below are Eastern US Daylight Time (UTC-04:00)

    [2015-06-24 16:06:52] <Mikaela> I think that commit just makes using the module more difficult and if someone cared about it that much they would set the command to PRIVMSG nickserv!nickserv@services.
    [2015-06-24 16:07:08] <Mikaela> the module has command for setting the identificarion command, doesn't it?
    [2015-06-24 16:10:44] <ChasedSpade> Mikaela: Yes, it does. I'm gonna make another change to that pull req. I've rethought how it should have been done.
    [2015-06-24 16:11:19] <Mikaela> does the change make it any more useful or easier?
    [2015-06-24 16:12:33] <ChasedSpade> It would remove my SetNSMask command, and change the identify command to "PRIVMSG NickServ@mask", so yes, it would make it easier as no commands would need to be run.
    [2015-06-24 16:13:01] <Mikaela> ChasedSpade: and how do you know the mask?
    [2015-06-24 16:14:14] <Mikaela> it can be configurd in network config file and is almost never the default, but services.example.org
    [2015-06-24 16:14:50] <Mikaela> and if you whoised the nickserv, what if it was impostor or missing ulinr and the is a service numeric?
    [2015-06-24 16:15:16] <Mikaela> the current NICKSERV is defacto standard and I don't think there is anything better which to change it to
    [2015-06-24 16:15:20] <Mikaela> ChasedSpade:
    [2015-06-24 16:20:32] <ChasedSpade> Sorry, internet outage due to weather.
    [2015-06-24 16:20:37] <teward> Mikaela: CSpade is commenting on the risk of IRC equiv of phishing
    [2015-06-24 16:20:50] <teward> he shared with me in PM that this would 'mitigate' that risk
    [2015-06-24 16:20:58] <teward> my question turned to configuration vs. documentation vs. practicality
    [2015-06-24 16:21:03] <teward> Makes it a little more complex? Yes.
    [2015-06-24 16:21:10] <Mikaela> teward: what risk?
    [2015-06-24 16:21:13] <teward> Do nets actively publicly document the mask that their services will use?  No.
    [2015-06-24 16:21:23] <teward> Can this be practically used in the case of NoDocumentation?  No.
    [2015-06-24 16:21:44] <Mikaela> teward: all IRCd exampme configs have u-line: yes in /nickserv alias that I have seen
    [2015-06-24 16:21:46] <teward> Mikaela: a little while back on freenode (and on other nets which are badly configured), 'nickserv' wasn't held/restricted so someone could be on that nick
    [2015-06-24 16:21:56] <teward> Mikaela: that's what CSpade said anyways, on my last comment
    [2015-06-24 16:22:03] <teward> Mikaela: and I brought that up in my discussion with CSpade in PM
    [2015-06-24 16:22:39] <teward> barring that, the question I had was on configurability and practicality.  Unless nets start publicly stating what the full mask for services will be, it's not practical to have the function for where documentation and information of what to configure it with doesn't exist
    [2015-06-24 16:22:41] <Mikaela> teward: yes and that has happened with all services at freenode and /NICKSERV that is currently used would have said that nickserv is impostor and refused to forward the command

DarthGandalf:
1. sasl is the future
2. this silently breaks the module, as it will just ignore auth request from every nickserv, unless user explicitly did something. I don't think any network has nickserv as `NickServ!services@hostname.org` literally

havokoc:
[201914] -!- NickServ [service@rizon.net]
[201914] -!-  ircname  : Nick Registration Service
[201914] -!-  server   : *.rizon.net [Where are you?]
[201914] -!- End of WHOIS

Mikaela:
> sasl is the future

when is this module going to go away then?

I believe `services.` (Atheme default) and `services.example.com` are more common.

PR #986, opened 2015-07-07
Don't print local ip address on 404 page, fixes #944

### LINKED ISSUE #944: Znc leaks the local IP of the machine it's running on.
On a 404, ZNC leaks the local IP of the server, even if you are not logged in.
I don't think, this is bad but it shouldn't happen, it's way too much information.

https://github.com/znc/znc/blob/master/src/HTTPSock.cpp#L650

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10842308-znc-leaks-the-local-ip-of-the-machine-it-s-running-on?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xPaw:
See #944 for discussion.

There's no reason to print it there, and it unnecessarily leaks information if the web server is sitting behind a proxy (CloudFlare for example) or a load balancer.

*Unrelated question: why is it sending xml stuff here, if the page is html?*

### DIFF
--- a/src/HTTPSock.cpp
+++ b/src/HTTPSock.cpp
@@ -653,10 +653,7 @@ bool CHTTPSock::PrintErrorPage(unsigned int uStatusId, const CString& sStatusMsg
 				"<h1>" + sStatusMsg.Escape_n(CString::EHTML) + "</h1>\r\n"
 				"<p>" + sMessage.Escape_n(CString::EHTML) + "</p>\r\n"
 				"<hr/>\r\n"
-				"<address>" +
-					CZNC::GetTag(false, /* bHTML = */ true) +
-					" at " + GetLocalIP().Escape_n(CString::EHTML) + " Port " + CString(GetLocalPort()) +
-				"</address>\r\n"
+				"<p>" + CZNC::GetTag(false, /* bHTML = */ true) + "</p>\r\n"
 			"</body>\r\n"
 		"</html>\r\n";
 


DarthGandalf:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address
It's HTML tag, though ZNC uses it incorrectly.

Also I think this should go to 1.6 branch, not to master.

2015-07-07 17:03 GMT+01:00 Pavel Djundik <notifications@github.com>:

xPaw:
I didn't mean address tag, I meant `<?xml` tag at the top, and `xmlns` on `html`. That stuff isn't needed.

DarthGandalf:
Ah, I wanted to switch to XHTML.
But I had to rollback the last part of it (content-type), because of MSIE.
So now it's XHTML... except that it's not.

2015-07-07 18:01 GMT+01:00 Pavel Djundik <notifications@github.com>:

xPaw:
Is there any particular reason why you wanted it to be XHTML?

DarthGandalf:
Just to make it more strict.
If XHTML parses correctly, there is a somewhat bigger chance that it
renders the same across browsers.
And in case of some error HTML is still successfully parsed, but
differently.

2015-07-07 18:07 GMT+01:00 Pavel Djundik <notifications@github.com>:

DarthGandalf:
LGTM, but please retarget it to 1.6 branch

xPaw:
I would need to open a new pull request for that.

DarthGandalf:
Ok, commited as 3bc7b820ace3bf00b8532e23e1a88b763f758955

PR #990, opened 2015-07-08
Allow reseting channel specific AutoClearChanBuffer & Buffer settings

jpnurmi:
The trick is to set the value to "-" in controlpanel. Notice that
controlpanel supports wildcards, so this can be used to easily
reset all channels back to defaults if desired.

### DIFF
--- a/src/Chan.cpp
+++ b/src/Chan.cpp
@@ -262,6 +262,11 @@ void CChan::InheritAutoClearChanBuffer(bool b) {
 	}
 }
 
+void CChan::ResetAutoClearChanBuffer() {
+	SetAutoClearChanBuffer(m_pNetwork->GetUser()->AutoClearChanBuffer());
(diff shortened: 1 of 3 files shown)


src/Chan.cpp
```diff
@@ -262,6 +262,11 @@ void CChan::InheritAutoClearChanBuffer(bool b) {
 	}
 }
 
+void CChan::ResetAutoClearChanBuffer() {
+	SetAutoClearChanBuffer(m_pNetwork->GetUser()->AutoClearChanBuffer());
```
jpnurmi: is a null check necessary? is it possible to have a floating CChan instance that is not associated to a network? the constructor already assumes that a valid network is passed

jpnurmi:
Does someone want to do the webadmin part? Some sort of reset-button.

PR #991, opened 2015-07-12
Add initializer_list ctors for CString & MCString

jpnurmi:


DarthGandalf:
I think MCString should be switched to be simple typedef instead, like done for VCString.
```
using MCString = std::map<CString, CString>;
``

jpnurmi:
The methods for reading from and writing to disk are obviously in a wrong place, but you wouldn't want to break compatibility, right?

DarthGandalf:
I don't like breaking compatibility, yes. But breaking all the modules with
changing parameters of *every* callback, and breaking only a few modules
which do what they shouldn't be doing anyway are different. (most modules
use GetNV/SetNV and don't care about MCString not being an actual std::map)
It's about tradeoffs.
If benefits of changing API overweight harm of breaking compatibility,
let's do it.
If it's possible (and easy) to preserve old API too, while still gaining
benefits of new API, and therefore not breaking compatibility, let's do it.

2015-07-13 11:32 GMT+01:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
Those message structure changes, if they had been backwards incompatible, would have justified a new major version. Does MCString? :P

DarthGandalf:
Your call. I don't like chasing high version numbers like Chrome does.

2015-07-13 16:11 GMT+01:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
I've added the same thing for CString, and a todo comment for replacing MCString  with ```using MCString = std::map<CString, CString>;```. Both are implicitly tested by the updated unit tests. The motivation for this change: managing message tags becomes easier.

PR #996, opened 2015-07-14
Add support for account-notify and extended-join

### LINKED ISSUE #316: Add support for 'account-notify' and 'extended-join' capabilities [$20]
Similar to #315. These two capabilities would allow clients to see users' services accounts without having to poll WHO or even WHOIS. This is supported by freenode server-side.

Both capabilities are usually used together:

 - http://ircv3.org/extensions/account-notify-3.1
 - http://ircv3.org/extensions/extended-join-3.1

---

(Bonus points if znc itself enables/disables the capabilities on the server depending on how many attached clients request them, to reduce traffic from the server (useless AWAY and ACCOUNT messages). On the other hand, this would probably make the znc implementation more complex than needed.)

<bountysource-plugin>
---
There is a **[$20 open bounty](https://www.bountysource.com/issues/141915-add-support-for-account-notify-and-extended-join-capabilities-20?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:
Close #316

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -98,6 +98,7 @@ public:
 			  m_bCapNotify(false),
 			  m_bAwayNotify(false),
 			  m_bAccountNotify(false),
+			  m_bExtendedJoin(false),
 			  m_bNamesx(false),
 			  m_bUHNames(false),
 			  m_bAway(false),
@@ -123,6 +124,7 @@ public:
 			               {"cap-notify", {false, [this](bool bVal) { m_bCapNotify = bVal; }}},
 			               {"away-notify", {true, [this](bool bVal) { m_bAwayNotify = bVal; }}},
 			               {"account-notify", {true, [this](bool bVal) { m_bAccountNotify = bVal; }}},
+			               {"extended-join", {true, [this](bool bVal) { m_bExtendedJoin = bVal; }}},
 			              })
 	{
 		EnableReadLine();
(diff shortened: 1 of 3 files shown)


DarthGandalf:
Can it be in a module?

2015-07-14 16:14 GMT+01:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
Controlling "server-dependent" caps from a module? Hmmph...

DarthGandalf:
I'm not fan of dumping everything to core.

DarthGandalf:
And it should be possible to implement various aspects of bouncing from a module. E.g. from a third-party module, if someone doesn't want to wait until ZNC gains some particular feature.

jpnurmi:
What made away-notify acceptable in core, and why did we waste time building that generic map of server-dependent capabilities if nothing else than away-notify will use it? away-notify, account-notify, and extended-join _together_ are the capabilities that make who-polling redundant...

TingPing:
I would say these CAPs are pretty 'core' to the irc experience there is no reason somebody wouldn't want them and they are extremely useful for any modern client.

jpnurmi:
Ping?

Mikaela:
Is the current status on this disagreement between should core IRC feature be in core (I agree with yes) or in module?

DarthGandalf:
@jpnurmi Ah, you were waiting for me?
If you think it should be merged, do it

oakkitten:
i totally :heart: you @jpnurmi

PR #1000, opened 2015-07-15
Say hello to a Message type

jpnurmi:


jpnurmi:
CMessage can be utilized to a greater degree all around the code base, but I deliberately tried to keep the changes minimal to minimize regressions. Also, the more I deviate from the mainline, the less likely it is to ever get merged back... :P

jpnurmi:
If the static_cast is unacceptable, one alternative option is:
```
class CMessage {
public:
    static CMessage* Parse(const CString& sLine);
    static CMessage* Create(const CNick& Nick, const CString& sCommand, ...);
};
```
I wanted to avoid that clumsy heap allocation, and have value semantics instead.

DarthGandalf:
16.07.2015 23:15, J-P Nurmi пишет:
>
> If the static_cast is unacceptable, one alternative option is:
>
> |class CMessage {
> public:
>     static CMessage* Parse(const CString& sLine);
>     static CMessage* Create(const CNick& Nick, const CString& sCommand, ...);
> };
> |
>
> I wanted to avoid that clumsy heap allocation, and have value
> semantics instead.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/znc/znc/pull/1000#issuecomment-122119655>.
>
I didn't say it's unacceptable...

jpnurmi:
I have rebased the PR to resolve conflicts. Good to go? For what it's worth, I've been running various versions of this in production for several weeks.

BtbN:
I'm very much looking forward for this to be merged.

DarthGandalf:
And new hooks are missing from modperl/modpython.
Otherwise, looks good, I guess.

jpnurmi:
> Please rename it to something which makes it obvious that it filters some tags.

I'd like to specify that ```CAP LS 303``` implies that a client can handle/parse arbitrary tags. Until then, some filtering is required to avoid breaking clients. I'd find it short-sighted to call the method PutMessageFilterTags(), because the goal is to _not_ filter anything whenever we can.

> And new hooks are missing from modperl/modpython.

Yeah, also
- docs for the new hooks and CMessage
- OnUserXxxMessage() hooks
- perhaps OnRaw(CMessage) & OnSendToClient(CMessage)?
- ZNC to request server-time
- ...

I've promised to follow up. Hopefully someone could help with some of this. Especially the SWIG stuff is not so my personal favorite. :)

DarthGandalf:
CAP LS 303 is not there yet, and it's not implies that.

Even **if** 303 will behave this way, when a client will send it, a new method PutMessage() can be written which does not filter tags, but old one still needs to exist for clients which support 302, but not 303.

jpnurmi:
> Even if 303 will behave this way, when a client will send it, a new method PutMessage() can be written which does not filter tags, but old one still needs to exist for clients which support 302, but not 303.

Only one version, a version that is guaranteed not to break clients, makes sense in the ZNC API. If a module wants to bypass the filter and potentially break clients, it's no more complex than ```PutClient(msg.ToString())```.

DarthGandalf:
Then please add doxygen to both overloads of `PutClient` which describe what they do, and why they behave differently.

jpnurmi:
Rebased again, to resolve conflicts. I would take care of the docs when things have settled. CMessage itself and all the new hooks need to be documented as well.

jpnurmi:
What if it was called ```PutMessage()``` instead? Since it's API of ```CClient```, one could argue that the "client" part in ```PutClient()``` is redundant.

DarthGandalf:
Sure :)
But that doesn't remove the need for docs. There will still be confusion why CClient::PutMessage silently drops some tags which I passed to it.

jpnurmi:
Anything besides the missing documentation?

jpnurmi:
Rebased & resolved conflicts again. I'm waiting for this to be merged until I develop things any further. Anyone willing to merge, or give green light to do so?

DarthGandalf:
I already asked you to add a comment to PutClient()s about why they behave differently.

DarthGandalf:
Well, either PutClient or PutMessage, but right now it *will* be confusing (and frustrating).

jpnurmi:
> I already asked you to add a comment to PutClient()s about why they behave differently.

I've promised to write docs when things have settled. It's not yet clear how unknown tags will be handled. There probably needs to be a module hook for that, as the TODO comment mentions. I'm not eager to write docs clarifying what the current unfinished implementation does, because they'd need to be rewritten when the missing module hook has been introduced.

> Well, either PutClient or PutMessage, but right now it will be confusing (and frustrating).

Which one do you prefer? ```PutMessage(CMessage)``` was an idea that could have distinguished it from ```PutClient(CString)```, but I chose to keep ```PutClient(CMessage)``` because we might want to have ```PutIRC(CMessage)``` overload later. Also, ```PutStatus(CTable)``` and ```PutModule(CTable)``` don't have _Table_ in their names.

BtbN:
I think PutClient is better, as with PutMessage it isn't imedaitely obvious where the message is sent, unless it's a member function of CClient, then i think PutMessage is better.

jpnurmi:
It's a member function of CClient. The PutXxx() convention seems to use a "target" name rather than what's being sent. PutClient(), PutStatus(), PutIRC(), PutModule()...

DarthGandalf:
> I'm not eager to write docs clarifying what the current unfinished implementation does

Well, when it will be finished? When IRCv3.3 is released with some extension which they don't even know about?
Module developers who want to use new shiny CMessage can start writing their modules right after this PR is merged.
I'm not asking you to write 200pages "document", but just a comment in the header.

> because we might want to have PutIRC(CMessage)

Why would you want `CClient::PutIRC` which sends something to server?

jpnurmi:
> Well, when it will be finished?

There's a subtle difference between "finished" and "settled". :) I meant to wait until the missing module hook has been implemented, nevertheless. That will certainly cause changes to the documentation that I have now written.

> When IRCv3.3 is released with some extension which they don't even know about?

That idea, if ever implemented, would allow us to not filter tags for certain compliant clients. It won't remove the need for the aforementioned module hook, though.

> Module developers who want to use new shiny CMessage can start writing their modules right after this PR is merged.

Ok :)

> I'm not asking you to write 200pages "document", but just a comment in the header.

https://github.com/znc/znc/commit/d77d21cfb9dd71ffcfce813aed7ccfe5e08a9ac6

>> because we might want to have PutIRC(CMessage)

> Why would you want CClient::PutIRC which sends something to server?

For whatever reason ```CClient::PutIRC(CString)``` exists, if it makes sense to have it at all, an overload that takes ```CMessage``` would be in many cases more convenient.

DarthGandalf:
> d77d21c

Thanks!

> For whatever reason CClient::PutIRC(CString) exists

Oh, wow. It exists...

jpnurmi:
Yay! Thanks! :)

PR #1001, opened 2015-07-16
Make adminlog path customizable

anlutro:


teward:
Have you tested this to verify that you have configuration capabilities added to support this?  I.E. this path is configurable by direct message or argument loadtime, as well as stored somewhere in the config?  (I haven't tested, but these're valid questions)

modules/adminlog.cpp
```diff
@@ -46,7 +46,11 @@ class CAdminLogMod : public CModule {
 		else
 			m_eLogMode = LOG_TO_FILE;
 
-		m_sLogFile = GetSavePath() + "/znc.log";
+		CString sPath = GetNV("path");
+		if (sPath.Equals(""))
+			m_sLogFile = GetSavePath() + "/znc.log";
+		else
+			m_sLogFile = sPath
```
teward: Missing a semicolon at the end here, so it'll Fail To Build.

anlutro:
That didn't occur to me as I manage my registry files manually. Thanks for the reminder! I'll implement a command for setting the path.

modules/adminlog.cpp
```diff
@@ -85,6 +87,30 @@ class CAdminLogMod : public CModule {
...
+
+		CFile LogFile(sPath);
+		CString sLogDir = LogFile.GetDir();
+		struct stat ModDirInfo;
+		CFile::GetInfo(GetSavePath(), ModDirInfo);
+		if (!CFile::Exists(sLogDir)) {
+			CDir::MakeDir(sLogDir, ModDirInfo.st_mode);
+		}
+
+		if (!sPath.empty()) {
```
jpnurmi: sPath is never empty at this point?

modules/adminlog.cpp
```diff
@@ -85,6 +87,30 @@ class CAdminLogMod : public CModule {
...
+			m_sLogFile = sPath;
+			SetNV("path", sPath);
+			sMessage = "adminlog file path set to [" + sPath + "]";
+			Log(sMessage);
+			PutModule(sMessage);
```
jpnurmi: it probably shouldn't do these when called from OnLoad()

modules/adminlog.cpp
```diff
@@ -27,6 +27,7 @@ class CAdminLogMod : public CModule {
 		AddHelpCommand();
 		AddCommand("Show", static_cast<CModCommand::ModCmdFunc>(&CAdminLogMod::OnShowCommand), "", "Show the logging target");
 		AddCommand("Target", static_cast<CModCommand::ModCmdFunc>(&CAdminLogMod::OnTargetCommand), "<file|syslog|both>", "Set the logging target");
+		AddCommand("Path", static_cast<CModCommand::ModCmdFunc>(&CAdminLogMod::OnPathCommand), "<path>", "If using the file target, set the log file path");
```
jpnurmi: just to throw an idea on the table... perhaps the path could be passed as an optional argument to the target command? not sure which is better.

anlutro:
Thanks for the feedback @jpnurmi - I agree with your suggestion of making the path a second arg to the target command. I suppose the main disadvantage is you can't switch between "file" and "both" without specifying the path over again.

modules/adminlog.cpp
```diff
@@ -85,6 +86,23 @@ class CAdminLogMod : public CModule {
 		Log("[" + sUsername + "] failed to login from " + sRemoteIP, LOG_WARNING);
 	}
 
+	void SetLogFilePath(CString& sPath) {
```
DarthGandalf: const CString&

DarthGandalf: Or, just CString. As it's modified below.
The point is, it shouldn't be modified on caller side.

modules/adminlog.cpp
```diff
@@ -46,7 +46,8 @@ class CAdminLogMod : public CModule {
 		else
 			m_eLogMode = LOG_TO_FILE;
 
-		m_sLogFile = GetSavePath() + "/znc.log";
+		CString sPath = GetNV("path");
+		SetLogFilePath(sPath);
```
DarthGandalf: After the fix below, `SetLogFilePath(GetNV("path"))` would be enough

anlutro:
@DarthGandalf done

PR #1005, opened 2015-07-23
src/IRCSock.cpp: use *status as kicker

### LINKED ISSUE #649: Change nick for "disconkick" module
`disconkick` kicks you using the `ZNC` nickname, which is a legal nickname. It should be changed to `*status`, `*disconkick`, or whatever.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4169484-change-nick-for-disconkick-module?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mikaela:
Closes #649

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -1115,7 +1115,7 @@ void CIRCSock::Disconnected() {
 
 	for (CChan* pChan : m_pNetwork->GetChans()) {
 		if(pChan->IsOn()) {
-			m_pNetwork->PutUser(":ZNC!znc@znc.in KICK " + pChan->GetName() + " " + GetNick()
+			m_pNetwork->PutUser(":*status!*status@znc.in KICK " + pChan->GetName() + " " + GetNick()
 				+ " :You have been disconnected from the IRC server");
 		}
 	}


Mikaela:
Actually this doesn't fix that as the user is able to change prefix and this is hardcoded with prefix \*. Or does it matter so much?

Mikaela:
https://github.com/znc/znc/issues/649#issuecomment-124179664 and if the hardcoded \* isn't too big issue.

DarthGandalf:
Of course it's an issue.
If user configured prefix to be ^, why would *status kick instead of ^status?
Also why * in ident? All the other stuff sent from *status doesn't have it.

Mikaela:
> Of course it's an issue.

Closing then.

>  Also why * in ident? All the other stuff sent from *status doesn't have it.

because \*modules_online gives hostmask `*status!*status@znc.in` when you whois \*status and I think having two idents for same "nick" would be confusing.

Do you have any opinions to #649 discussions?

DarthGandalf:
Response to `/msg *status help` is from `*status!znc@znc.in`, so *modules_online should be fixed then.

PR #1013, opened 2015-07-30
Support "CAP :END"

DarthGandalf:
Thanks to jaybe for complaining about this.


jpnurmi:
Provided that #1000 gets merged at some point, the issue would be eventually fixed in master (by a #1000 follow up). I'm fine with having this in 1.6.x regardless.

DarthGandalf:
Yep, that's why I made it for 1.6.x
#1000 can't go to 1.6 branch.

PR #1014, opened 2015-08-01
Make it possible to detach also queries

jpnurmi:


DarthGandalf:
Before I go too deep to code, may you explain what's the use of it?

jpnurmi:
I wanted to detach work-related queries while on vacation. It's like ignore with playback buffer. When I'm back to work, I can conveniently restore the discussion straight to my client by attaching the queries.

Issue #983, opened 2015-06-29
Why do we have <add/del/list>bindhost(s)?


Mikaela:
```
< *status> ------
< *status> Command: AddBindHost <host (IP preferred)>
< *status> Description: Adds a bind host for normal users to use
< *status> ------
< *status> Command: DelBindHost <host>
< *status> Description: Removes a bind host from the list
< *status> ------
< *status> Command: ListBindHosts 
< *status> Description: Shows the configured list of bind hosts
< *status> ------
```

> addbindhost is one of the most useless and confusing misfeatures of ZNC currently.—@DarthGandalf 

```
2015-06-29 07:49:18+0300 < Mikaela> DarthGandalf: so if you setuserbindhost, it's automatically added and used for all users too?
2015-06-29 09:15:18+0300 < DarthGandalf> Mikaela: no. addbindhost adds it to global list of hosts. If that global list is non-empty, non-admins are allowed to setbindhost only from that list. That's the only meaning of addbindhost
2015-06-29 09:16:15+0300 < DarthGandalf> poutine: znc -D
2015-06-29 09:16:17+0300 < Mikaela> DarthGandalf: but with the case of broken IPv6, isn't it wanted to addbindhost 0.0.0.0 so IPv6 isn't attempted by any user? I think the Happy Eyeballs issue isn't still fixed.
2015-06-29 09:16:44+0300 < DarthGandalf> Again. addbindhost 0.0.0.0 won't automatically fix it for any users.
2015-06-29 09:17:11+0300 < Mikaela> DarthGandalf: I thought that after adding it, ZNC sets all users to use that as they cannot use anything else?
2015-06-29 09:17:12+0300 < DarthGandalf> It will just limit non-admins' ability to setbindhost to select only 0.0.0.0 and nothing else (or whatever they have set already)
2015-06-29 09:17:49+0300 < Mikaela> in webadmin at least adding it there makes user default bindhost be "default" which is that and network bindhosts do the same
2015-06-29 09:17:56+0300 < DarthGandalf> addbindhost is one of the most useless and confusing misfeatures of ZNC currently.
2015-06-29 09:18:12+0300 < Mikaela> I see. I will sduggest setuserbindhost in the future then?
2015-06-29 09:19:13+0300 < DarthGandalf> One more reason: if admin thinks that addbindhost actually limits users' bindhosts to one of list, they are wrong again: if setbindhost is "", OS will choose any IP on computer by usual rules
```

PR #1018, opened 2015-08-04
Add CZNC::IsBindHostAllowed()

jpnurmi:


src/znc.cpp
```diff
@@ -338,6 +338,18 @@ bool CZNC::IsHostAllowed(const CString& sHostMask) const {
 	return false;
 }
 
+bool CZNC::IsBindHostAllowed(const CString& sBindHost) const {
```
DarthGandalf: This is as useless as it was before this PR.
If sBindHost is empty, OS is free to choose *any* IP and connect from it. And user is always free to make bindhost empty regardless of this list.

DarthGandalf:
Also see #983

jpnurmi:
So what do you suggest instead? My intention is not to fix any bindhost misbehavior, or change any behavior at all, but just share that piece of code that is already repeated in five different places within the ZNC codebase. Then I don't have to repeat it once again in https://github.com/jpnurmi/znc-settings...

DarthGandalf:
Just remove this feature? Right now it does more harm than good, and adding more API for it won't help.
If you have some idea how to make such list of hosts to be actually useful, the API can follow (maybe it would even stay the same)

PR #1019, opened 2015-08-04
CZNC: add missing SSL-related getters and setters

jpnurmi:
- SSLCiphers
- SSLProtocols
- SSLCertFile

### DIFF
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -1098,42 +1098,13 @@ bool CZNC::LoadGlobal(CConfig& config, CString& sError) {
 		m_bProtectWebSessions = sVal.ToBool();
 	if (config.FindStringEntry("hideversion", sVal))
 		m_bHideVersion = sVal.ToBool();
-
-	if (config.FindStringEntry("sslprotocols", m_sSSLProtocols)) {
-		VCString vsProtocols;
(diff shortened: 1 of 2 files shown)


src/znc.cpp
```diff
@@ -1098,42 +1098,12 @@ bool CZNC::LoadGlobal(CConfig& config, CString& sError) {
...
-			} else if (bDisable) {
-				m_uDisabledSSLProtocols |= uFlag;
-			} else {
-				m_uDisabledSSLProtocols = ~uFlag;
-			}
+	if (config.FindStringEntry("sslprotocols", sVal)) {
+		if (!SetSSLProtocols(sVal)) {
+			CUtils::PrintError("Invalid SSLProtocols value.");
+			CUtils::PrintError("The syntax is [SSLProtocols = [+|-]<protocol> ...]");
+			CUtils::PrintError("Available protocols are [SSLv2, SSLv3, TLSv1, TLSv1.1, TLSv1.2]");
```
DarthGandalf: LGTM, except that now this list is detached from implementation, which makes it more likely to forget to update, whenever TLS1.3 comes.

jpnurmi: Good point, updated

Issue #963, opened 2015-05-07
Per-network traffic info


sim642:
Currently ZNC can show traffic used by each user but it'd be nice if there would be separate counters for each network for an user.

Mikaela:
I am not sure if this is what you are after, but you might find \*listsockets interesting.

sim642:
That just lists the sockets in use for the connections but not how much traffic has gone through them.

Cajs:
This would be a nice edition, especially for those users running VPS' on "metered bandwidth" (For example, some smaller VPS' only restricted network throughput, e.g 1gb per month).

This would aid in knowing where the data is being used.

GLolol:
:+1: from me.

PR #1020, opened 2015-08-05
Calculate per-network traffic (#963)

jpnurmi:


DarthGandalf:
No indication of it in existing UI?

jpnurmi:
No, but [znc-admin](https://github.com/jpnurmi/znc-admin) (needs also #1019 and [341675c](https://github.com/jpnurmi/znc/commit/341675c230574e7d4c149a8ad14d8d4fc11fabc2)) can show per-network traffic

DarthGandalf:
znc-admin isn't going to replace webadmin, right? Please add this to Traffic tab in webadmin.

jpnurmi:
Ok, per-network traffic stats added to webadmin.

src/Client.cpp
```diff
@@ -73,7 +73,10 @@ CClient::~CClient() {
 		CClientAuth* pAuth = (CClientAuth*) &(*m_spAuth);
 		pAuth->Invalidate();
 	}
-	if (m_pUser != nullptr) {
+	if (m_pNetwork != nullptr) {
```
jpnurmi: btw, i'm not so sure if client side traffic should be associated to any network or not

include/znc/znc.h
```diff
@@ -89,6 +89,7 @@ class CZNC {
 	// generated through ZNC.
 	TrafficStatsMap GetTrafficStats(TrafficStatsPair &Users,
 			TrafficStatsPair &ZNC, TrafficStatsPair &Total);
+	TrafficStatsPair GetNetworkTrafficStats(const CString& sUsername, TrafficStatsMap& Networks);
```
DarthGandalf: Any comments?
This function looks like a reverse of previous one (order of args/return)

jpnurmi: Ok. Changed to use the same order of args/return.

PR #1021, opened 2015-08-06
Add <time.h> includes where appropiate

gustavoz:
Signed-off-by: Gustavo Zacarias <gustavo@zacarias.com.ar>


BtbN:
Shouldn't it be ctime instead of time.h?

PR #1024, opened 2015-08-11
Fix #930: Partyline is completely broken

### LINKED ISSUE #930: Partyline is completely broken [$10]
I've tested this on 2 different servers, with 1.6.0 and earlier versions. I've been poking around the source but I can't figure out what's wrong.

Partyline channels are only working for people connected to freenode. We've tested snoonet, dalnet, etc. Basically, it comes down to this:

Person A is on freenode. Person A joins the ~#partyline. Person A can send and receive.

Person B is on dalnet. Person B joins the ~#partyline. Person B can send, but not receive.

Person C is on snoonet. Person B joins the ~#partyline. Person B can send, but not receive.

Looking at the raw data, sending to the ~#partyline works fine. But nobody is receiving anything at all from the server.

If I use send_raw to send ":test!test@test PRIVMSG ~#partyline :whatever" that comes through OK.

Tested on Linux and Widows with mIRC, XChat, HexChat, HydraIRC, and Quassel. The issue is independent of the client.

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/9870101-partyline-is-completely-broken?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:


dgw:
:+1: This is so very welcome!

modules/partyline.cpp
```diff
@@ -188,33 +186,26 @@ class CPartylineMod : public CModule {
 		return CONTINUE;
 	}
 
-	virtual EModRet OnRaw(CString& sLine) override {
-		if (sLine.Token(1) == "005") {
+	EModRet OnSendToClient(CString& sLine, CClient& Client) override {
```
DarthGandalf: This would look nicer with CMessage

jpnurmi: Yeah, well, I wish it was already merged weeks ago. :P

jpnurmi:
Don't merge yet. If #1059 goes in, this can be made much simpler.

jpnurmi:
Will be fixed in master instead...

jpnurmi:
Replaced by https://github.com/znc/znc/pull/1064

PR #1026, opened 2015-08-11
Fix #124: OnChanMsg(): nick doesn't have perms

### LINKED ISSUE #124: OnChanMsg(): nick doesn't have perms
Nick parameter of the callback doesn't contain info on permissions.
To get the proper nick, currently need to FindNick() in the channel with Nick's nick and use it instead.
It's not very nice...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/690-onchanmsg-nick-doesn-t-have-perms?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:


src/IRCSock.cpp
```diff
@@ -970,9 +970,31 @@ bool CIRCSock::OnPrivMsg(CNick& Nick, CString& sMessage) {
 	return false;
 }
 
+// #124: OnChanMsg(): nick doesn't have perms
+static void FixupChanNick(CNick& Nick, CChan* pChan) {
```
DarthGandalf: `const CChan*`?

jpnurmi: needs write access to the chan nick (to sync ident & host), but const overload of FindNick() would return const CNick*

DarthGandalf: ah, right.

PR #1027, opened 2015-08-11
Add the pyeval module

kylef:
This PR adds the pyeval module I wrote many years ago.

/cc @DarthGandalf

http://wiki.znc.in/Pyeval

### DIFF
--- a/modules/modpython/pyeval.py
+++ b/modules/modpython/pyeval.py
@@ -1,0 +1,79 @@
+#
+# Copyright (C) 2004-2015 ZNC, see the NOTICE file for details.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


modules/modpython/pyeval.py
```diff
@@ -0,0 +1,79 @@
...
+        self.locals['znc'] = znc
+        self.locals['module'] = self
+        self.locals['user'] = self.GetUser()
+        self.locals['network'] = self.GetNetwork()
+        self.indent = re.compile(r'^>+')
+
+        return True
+
+    def OnModCommand(self, line):
+        self.locals['client'] = self.GetClient()
```
jpnurmi: should you move ```self.locals['network'] = self.GetNetwork()``` here so the network is set correctly when loaded as a user module?

kylef: My intention was that this would be the network you used when talking to the module. If you're not connected via a network, then GetNetwork() will simply return None, and the value for the `network` in local scope would be None.

jpnurmi: > My intention was that this would be the network you used when talking to the module.

It would behave the way you describe if you set 'network' here. Currently 'network' points to the network where the module was loaded if any. For example, if the module is loaded via webadmin, 'network' will always be None.

Mikaela:
Why in /modules/modpython instead of just /modules? Perleval is also there https://github.com/znc/znc/blob/master/modules/perleval.pm .

jpnurmi:
Mikaela is right, it should be modules/pyeval.py. Otherwise it won't be installed (and not found by --enable-run-from-source builds either).

jpnurmi:
Try evaluating ```help(user)``` :)

PR #1028, opened 2015-08-11
blockuser: Add ability for a block reason to be specified

### LINKED ISSUE #1015: blockuser: Add ability for a block reason to be specified with a bloc
It'd be fairly nice to be able to specify a block reason for blockuser.  Is it possible to add this functionality?

jpnurmi:
close #1015

### DIFF
--- a/modules/blockuser.cpp
+++ b/modules/blockuser.cpp
@@ -34,21 +34,20 @@ public:
 
 	bool OnLoad(const CString& sArgs, CString& sMessage) override {
 		VCString vArgs;
-		VCString::iterator it;
-		MCString::iterator it2;
+		MCString::iterator it;
 
 		// Load saved settings
-		for (it2 = BeginNV(); it2 != EndNV(); ++it2) {
+		for (it = BeginNV(); it != EndNV(); ++it) {
 			// Ignore errors
-			Block(it2->first, it2->second);
+			Block(it->first, it->second);
 		}
 
 		// Parse arguments, each argument is a user name to block
 		sArgs.Split(" ", vArgs, false);
 
-		for (it = vArgs.begin(); it != vArgs.end(); ++it) {
-			if (!Block(*it)) {
-				sMessage = "Could not block [" + *it + "]";
+		for (const CString& sArg : vArgs) {
+			if (!Block(sArg)) {
+				sMessage = "Could not block [" + sArg + "]";
 				return false;
 			}
 		}
@@ -182,16 +181,15 @@ private:
 
 		// Disconnect all clients
 		vector<CClient*> vpClients = pUser->GetAllClients();
-		vector<CClient*>::iterator it;
-		for (it = vpClients.begin(); it != vpClients.end(); ++it) {
-			(*it)->PutStatusNotice("Blocked: " + (sReason.empty() ? DEFAULT_REASON : sReason));
-			(*it)->Close(Csock::CLT_AFTERWRITE);
+		for (CClient* pClient : vpClients) {
+			pClient->PutStatusNotice("Blocked: " + (sReason.empty() ? DEFAULT_REASON : sReason));
+			pClient->Close(Csock::CLT_AFTERWRITE);
 		}
 
 		// Disconnect all networks from irc
 		vector<CIRCNetwork*> vNetworks = pUser->GetNetworks();
-		for (vector<CIRCNetwork*>::iterator it2 = vNetworks.begin(); it2 != vNetworks.end(); ++it2) {
-			(*it2)->SetIRCConnectEnabled(false);
+		for (CIRCNetwork* pNetwork : vNetworks) {
+			pNetwork->SetIRCConnectEnabled(false);
 		}
 
 		SetNV(pUser->GetUserName(), sReason);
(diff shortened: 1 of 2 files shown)


modules/blockuser.cpp
```diff
@@ -58,7 +58,8 @@ class CBlockUser : public CModule {
 
 	EModRet OnLoginAttempt(std::shared_ptr<CAuthBase> Auth) override {
 		if (IsBlocked(Auth->GetUsername())) {
-			Auth->RefuseLogin(MESSAGE);
+			CString sMessage = GetNV(Auth->GetUsername());
+			Auth->RefuseLogin(sMessage.empty() ? DEFAULT_MESSAGE : sMessage);
```
teward: ```Auth->RefuseLogin("Blocked: " + sMessage.empty() ? DEFAULT_MESSAGE : sMessage));``` would make this post the error message "Blocked: REASON", such that if no input is provided, then it'd say "Blocked: Your account has been disabled. Contact your administrator."

This works a little better for custom ban reasons, but this is just my opinion.

modules/blockuser.cpp
```diff
@@ -175,7 +180,7 @@ class CBlockUser : public CModule {
 		vector<CClient*> vpClients = pUser->GetAllClients();
 		vector<CClient*>::iterator it;
 		for (it = vpClients.begin(); it != vpClients.end(); ++it) {
-			(*it)->PutStatusNotice(MESSAGE);
+			(*it)->PutStatusNotice(sMessage.empty() ? DEFAULT_MESSAGE : sMessage);
```
teward: Like the other code suggestion I just made...

```(*it)->PutStatusNotice(Blocked: " + sMessage.empty() ? DEFAULT_MESSAGE : sMessage));``` would actually disconnect the user with a "Blocked: REASON" message, like "Blocked: Your account has been disabled. Contact your administrator." for the default.  Like the other suggestion, it works better this way with Custom ban reasons, but also fairly decently for the default message.

Again, just my two cents.

Comment (jpnurmi) [inline modules/blockuser.cpp]: Sounds good. Thanks for the suggestion. I've updated the PR.

File: modules/blockuser.cpp [hunk 3]
```diff
@@ -173,21 +181,22 @@ class CBlockUser : public CModule {
...
 		// Disconnect all networks from irc
 		vector<CIRCNetwork*> vNetworks = pUser->GetNetworks();
-		for (vector<CIRCNetwork*>::iterator it2 = vNetworks.begin(); it2 != vNetworks.end(); ++it2) {
-			(*it2)->SetIRCConnectEnabled(false);
+		for (CIRCNetwork* pNetwork : vNetworks) {
+			pNetwork->SetIRCConnectEnabled(false);
 		}
 
-		SetNV(pUser->GetUserName(), "");
+		SetNV(pUser->GetUserName(), sReason);
```
DarthGandalf: I wonder about future compatibility. If something else will need to be added, where will it go? e.g. admin-only notes why ban. Or e.g. date when the ban will be automatically lifted.

modules/data/blockuser/tmpl/blockuser_WebadminUser.tmpl
```diff
@@ -3,9 +3,17 @@
...
+				<div class="inputlabel">Block:</div>
+				<div class="checkbox">
+					<input type="checkbox" name="embed_blockuser_block" id="opt_embed_blockuser_block" value="1"<? IF Blocked ?> checked="checked"<? ENDIF ?><? IF Self ?> disabled="disabled"<? ENDIF ?> />
+					<label for="opt_embed_blockuser_block">Account is blocked</label>
+				</div>
+			</div>
+			<div class="subsection half">
+				<div class="inputlabel">Reason:</div>
+				<input type="text" name="embed_blockuser_reason" value="<? VAR Reason ?>" />
+				<br /><span class="info">Empty = use standard value</span>
```
DarthGandalf: This is better as `placeholder` attribute.

modules/blockuser.cpp
jpnurmi: How would you prefer to have it? Extending this later for instance to something like ```SetNV(pUser->GetUserName() + "/reason", sReason)``` should not be a problem.

modules/data/blockuser/tmpl/blockuser_WebadminUser.tmpl
jpnurmi: Done

modules/blockuser.cpp
DarthGandalf: Well, I'm spoiled by protobuf, but that certainly won't be a good dependency for ZNC.
`SetNV(pUser->GetUserName() + "/reason", sReason)` would make it harder to use, and to e.g. iterate over NV like it does now.
Probably MCString can be extended to provide [un]serialization of map? It already does that when writing NV to disk.
Or json?

jpnurmi: There are many things we could do with JSON in the future. Pick your favorite JSON parser and throw it in third_party and we'll start using it where we can... Distant plans don't have to block this PR, though.

DarthGandalf: I don't have any favorite JSON parser, you can pick one. And the format is compatible between different parsers anyway, so it can be changed to some other library in future if needed.
Format of user settings on the user's machine can't be changed so easily, that's why I'd prefer this PR to use JSON.

PR #998, opened 2015-07-14
Resolve #967: separate chan & query buffer size settings

jpnurmi:

Issue #967, opened 2015-05-31
Default Settings > Buffer Size affects query buffering


l-c-l:
Hi,

is it possible to split Buffer Size for channel AND queries ?

In my case I want no channel buffering but query buffering.

And thanks for your work. it is a real nice bouncer especially the webif. :-)

Greetings from germany,
martin

PR #1029, opened 2015-08-13
Chan settings: renamed 'buffer' to 'buffer size'

ChasedSpade:
Changed the 'buffer' variable in SETCHAN and GETCHAN to match the 'buffercount' variable used in SET.

### DIFF
--- a/modules/controlpanel.cpp
+++ b/modules/controlpanel.cpp
@@ -130,7 +130,7 @@ class CAdminMod : public CModule {
 			static const char* cvars[][2] = {
 				{"DefModes",            str},
 				{"Key",                 str},
-				{"Buffer",              integer},
+				{"BufferSize",          integer},
 				{"InConfig",            boolean},
 				{"AutoClearChanBuffer", boolean},
 				{"Detached",            boolean}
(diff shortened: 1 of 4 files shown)


jpnurmi:
In master, there's a separate user-level setting for channel and query buffer sizes (#967). They are called ```ChanBufferSize```and ```QueryBufferSize```, respectively. If the chan-level setting is renamed, a) it should follow the same naming convention, and b) backwards compatibility should be kept like #998 did.

modules/controlpanel.cpp
```diff
@@ -705,12 +705,12 @@ class CAdminMod : public CModule {
 		for (CChan* pChan : vChans) {
 			if (sVar == "defmodes") {
 				PutModule(pChan->GetName() + ": DefModes = " + pChan->GetDefaultModes());
-			} else if (sVar == "buffer") {
+			} else if (sVar == "buffercount") {
```
jpnurmi: ```} else if (sVar == "buffersize" || sVar == "buffer") {```
to keep backwards compatibility

modules/controlpanel.cpp
```diff
@@ -763,11 +763,11 @@ class CAdminMod : public CModule {
 			if (sVar == "defmodes") {
 				pChan->SetDefaultModes(sValue);
 				PutModule(pChan->GetName() + ": DefModes = " + sValue);
-			} else if (sVar == "buffer") {
```
jpnurmi: ```} else if (sVar == "buffersize" || sVar == "buffer") {```
to keep backwards compatibility

jpnurmi:
:+1: 'buffer count' was very confusing, considering that there's another setting called MaxQueryBuffers that limits the count of query buffers. Anyway, the combined diff looks good. The commits could be squashed to one. Would you like to implement the same for webadmin?

ChasedSpade:
Sure. Would you like me to submit a new pull request or just keep using this one?

jpnurmi:
Either is fine. Whichever you prefer :)

jpnurmi:
Thanks for the contribution! :)

PR #1031, opened 2015-08-17
Implement OnUserXxxMessage(CXxxMessage&) module hooks

jpnurmi:
Supersedes #685


jpnurmi:
I've been running this in production since the PR was opened. No problems noticed so far.

jpnurmi:
Rebased to resolve conflicts in MessageTest.cpp

jpnurmi:
https://github.com/znc/znc/commit/b8088a5b025a50aac3ebe2ddc6405c4e93f48798 had an undesired side-effect that buffered messages (":{text}") would get an extra colon, so I've added ExcludeColon to CMessage::FormatFlags and used it in CClient::AddBuffer().

jpnurmi:
Notice that GitHub shows the commits in wrong order. Unlike it shows here, "Make CBufLine hold a CMessage internally" is not the last commit, because "CClient::ReadLine(): use CMessage" relies on it. Compare to the commits in the PR source repo/branch: https://github.com/jpnurmi/znc/commits/user-hooks

jpnurmi:
Sorry for a suboptimal interim state, by the way. There are some unnecessary conversions back and forth between CMessages and raw string lines. During buffer playback, at least. I will refactor CBuffer and CBufLine to eat and spit out CMessages later. I did some preliminary prototyping and realized that the patch gets surprisingly big. I didn't want to include it here, because the PR is already large enough. I will follow up, but as usual, I'd like to have this merged first...

PS. There are several tasks waiting for this PR. Should I convert my todo-list to GitHub tasks?

jpnurmi:
GitHub can no longer keep up with the changes. It's still showing the commits in wrong order. :P

jpnurmi:
Press the button today and get a discount on the next review!

PR #1035, opened 2015-08-17
Make ZNC request server-time when available (close #839)

### LINKED ISSUE #839: ZNC doesn't request znc.in/server-time-iso
```
2015-02-01 19:22:59+0200 < Mikaela> jpnurmi: a little offtopic here, but do you know if ZNC requests znc.in/server-time-iso when the server has it? Someone at #znc had issues some days ago with messages still having timestamps and it was that ZNC-chaining thing from FAQ.
2015-02-01 19:22:59+0200 < Mikaela> I just remembered, because you commented on some issue about server-time
2015-02-01 19:26:57+0200 < khronosschoty> I wonder if I compiled a binary of communi, if it would work on OS X 10.6
2015-02-01 19:27:17+0200 < khronosschoty> if I compiled the binary on 10.10
2015-02-01 20:57:51+0200 < jpnurmi> Mikaela: i don't think znc itself requests it
2015-02-01 20:58:12+0200 < Mikaela> and that probably causes issues with chained ZNCs as suggested in FAQ = bug?
2015-02-01 20:58:25+0200 < Mikaela> I can open issue soon
2015-02-01 20:58:55+0200 < jpnurmi> i hope people will no longer have to chain znc instances
2015-02-01 21:00:07+0200 < Mikaela> There are distributions like Debian Stable and that FAQ entry etc.
```

* [That FAQ entry](http://wiki.znc.in/FAQ#How_to_store_private_messages_even_when_user_is_attached.2C_so_other_clients_can_see_them.3F)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8202645-znc-doesn-t-request-znc-in-server-time-iso?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpnurmi:


include/znc/IRCSock.h
```diff
@@ -135,6 +136,7 @@ class CIRCSock : public CIRCSocket {
 	bool                                m_bAwayNotify;
 	bool                                m_bAccountNotify;
 	bool                                m_bExtendedJoin;
+	bool                                m_bServerTime;
```
DarthGandalf: Is it ever used?

DarthGandalf:
Let's also request `znc.in/server-time-iso` if available

include/znc/IRCSock.h
jpnurmi: Storing the boolean should be practically free due to memory alignment and padding, and potentially useful if we introduce a ShowNetwork -like command that shows all the details of a network.

jpnurmi:
> Let's also request ```znc.in/server-time-iso``` if available

Ok, updated.

PR #1049, opened 2015-08-29
Add an %empty% substitution in ExpandString that expands to the empty string.

kerio92:
This can be useful in various situations, for instance for quit messages.


jpnurmi:
:+1: i like the trick :)

Issue #384, opened 2013-07-16
Clone user results in network IRC nick that cannot be edited


pcolmer:
When I clone an existing user, I reset the Authentication values and the IRC Information values and then click on Clone. When I then edit the user, the cloned network information shows "IRC Nick" as having a value that seems to be derived from the original user, e.g. my original IRC Nick is "philip_colmer" but the cloned IRC Nick is "philip_c-". That value does not seem to be editable, even if I edit the network of the cloned user and change all of the Network Info values.

The only solution seems to be to delete the network and add it back, but that rather defeats the purpose of cloning the user.

jimpriest:
I've run into this as well.

* O User = Original user
* C User = Cloned user.

I clone the OUser.

When I look under CUser's networks I see "OUser" as the network 'name'.

If I edit the network and update all references of OUser to CUser and save, looking at the Networks list - the Network is still named "OUser" as is the IRC Nick which is confusing.

This can be seen in the URL as well:
https://my.url.com/mods/global/webadmin/editnetwork?user=cuser&network=ouser

Like previous posters - the only work around seems to be to delete the network and recreate it.

aj701:
I've been getting this error as well, restarting the znc server works, but a bit of a pain.

DarthGandalf:
Where are the details? What ZNC version?

aj701:
So I was having the same error as the above user. I was cloning user accounts via the web interface, and no matter what individual changes I made to account info the irc nickname for the server under the cloned account would default to the original IRC alt nick information. Even when he irc nick was changed and saved over, all the writeable fields would reflect the irc nick as changed, yet the server info would still default to the  alt of the original account.

One solution is to delete the server and create a new one, another solution is to just run 'pkill znc' , then relaunch it from a terminal. When this is done all the changed information that was saved via the web interface takes effect and there is no sign of the old 'alt nick' on the cloned account.

This is running the most recent version of znc 1.6.1

Mikaela:
Do you happen to use patched ZNC? There was some webadmin theme which required patching and has caused issues like this in the past, but I don't remember what it was called as.

aj701:
Hi Mikaela,

I have encountered this error on multiple servers with the default znc theme as well as using the patched firrre theme. The problem and solution for both remained the same for me.

Mikaela:
This should possibly also be mentioned here (separate issues removed for readability).

```
[20:32:47] <aj701> mikaela - did I just repsond to you on github? haha
[...]

PR #1050, opened 2015-08-30
CONTRIBUTING.md: ask for unmodified ZNC

Mikaela:
Specify unmodified ZNC in case the issue is caused by modified ZNC (such as
the SSL certificate validation check removal or firre webadmin theme
(see znc/znc#384 where after fixing patches have reintroduced it)).

<s>Thanks to @kerio92 & @Zarthus for wording suggestions.</s> DarthGandalf reworded it.

[CI SKIP]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,6 +4,8 @@ Reporting bugs
 * When reporting a bug:
     * Please ensure that you are on the latest version in case the bug you
       are reporting is already fixed.
+    * If you did some custom modification to ZNC, please make sure that
+      the bug isn't caused by that modification.
     * Please include the following information:
         * OS/distribution version
         * `/znc version`


Mikaela:
Potential issue: lsb_release is in redhat-lsb in Fedora-side. Can that be ignored with the word "usually" and assuming that it probably comes by default?

It ships by default with at least Debian and Ubuntu and the only Debian where it wasn't by default that I am aware of is Raspbian.

Suggestions are welcome.

Mikaela:
lsb-release also should be widely installed by default or available based on https://en.wikipedia.org/wiki/Linux_Standard_Base

Mikaela:
I updated the PR anyway, I don't think there are more package names on where it can be and as I said before there is the word *usually*.

Mikaela:
I removed the `lsb_release` as if I understood @DarthGandalf correctly at IRC, he was against it.

PR #1051, opened 2015-08-30
ControlPanel: Move syntax of GetNetwork and SetNetwork

ChasedSpade:
Fix command syntax of GetNetwork and SetNetwork.

### DIFF
--- a/modules/controlpanel.cpp
+++ b/modules/controlpanel.cpp
@@ -452,6 +452,11 @@ class CAdminMod : public CModule {
 
 		CIRCNetwork *pNetwork = nullptr;
 
+		if (sVar.empty()) {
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Uh, I should test stuff better before merging :(
This crashes when doing `/msg *controlpanel getnetwork nick $me` because `pNetwork == nullptr` (thx @KindOne- and Uptime)

ChasedSpade:
Here's the patch for controlpanel. Sorry about that. https://gist.github.com/ChasedSpade/456a82c28fa2dbc7ad82

DarthGandalf:
@ChasedSpade this would fail when you're editing some other user, but *you* are connected to a network

ChasedSpade:
Updated the Gist with a better patch (hopefully)

PR #1054, opened 2015-09-01
Add CMessage::GetType()

jpnurmi:


include/znc/Message.h
```diff
@@ -31,6 +31,21 @@ class CMessage {
 	explicit CMessage(const CString& sMessage = "");
 	CMessage(const CNick& Nick, const CString& sCommand, const VCString& vsParams = VCString(), const MCString& mssTags = MCString::EmptyMap);
 
+	enum Type {
```
DarthGandalf: enum class

jpnurmi: Done.

PS. I'd have preferred to keep ```CMessage::Notice``` instead of ```CMessage::Type::Notice```...

jpnurmi:
I didn't touch CClient::ReadLine() PRIVMSG/NOTICE handling for now, because it's a bit more complex change and conflicts with #1059.

jpnurmi:
The vision:
```
void CIRCSock::ReadLine(const CString& sData) {
	...
	CMessage Message(sLine);
	...
	bool bReturn = false;
	IRCSOCKMODULECALL(OnRawMessage(Message), &bReturn);
	if (bReturn) return;
	...

	switch (Message.GetType()) {
		...
		case CMessage::Type::Notice:
			bReturn = OnNoticeMessage(static_cast<CNoticeMessage&>(Message));
			break;
		case CMessage::Type::Numeric:
			bReturn = OnNumericMessage(static_cast<CNumericMessage&>(Message));
			break;
		...
	}
	if (bReturn) return;

	m_pNetwork->PutUser(Message);
}
```

DarthGandalf:
Let it be

PR #1060, opened 2015-09-02
Add a test for --makeconf

DarthGandalf:


jpnurmi:
Would it be better to answer no to "Launch ZNC now"? appveyor would pass?

DarthGandalf:
Hm, you're right. Launch is good, but it should be in separate test.

jpnurmi:
Aha, that's why you wanted to launch it! :P

PR #1064, opened 2015-09-04
Fix #930: Partyline is completely broken

### LINKED ISSUE #930: Partyline is completely broken [$10]

jpnurmi:


jpnurmi:
Btw, your mileage may vary.

The biggest remaining issue is that a networkless user is sent a very limited amount of messages on connection. Before it was only RPL_WELCOME. This PR ensures that a minimal RPL_ISUPPORT (CHANTYPES) is always sent, regardless of network. Any client that relies on a more complete RPL_ISUPPORT, or any other typical connection time messages such as RPL/ERR_MOTD*, might still not be 100% functional.

jpnurmi:
https://github.com/znc/znc/commit/078eef5ba9c5250c000afa3e28bb32a877bfb273

PR #1065, opened 2015-09-05
znc.service.in: start after network-online.target

Mikaela:
network.target only means that the network management is up, it doesn't
mean that there is connectivity.

network-online.target means that the network management thinks that
there is network connectivity which means that the system has IP
address.

https://wiki.freedesktop.org/www/Software/systemd/NetworkTarget/

### DIFF
--- a/znc.service.in
+++ b/znc.service.in
@@ -1,6 +1,6 @@
 [Unit]
 Description=ZNC, an advanced IRC bouncer
-After=network.target
+After=network-online.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


Mikaela:
Any comments?

DarthGandalf:
My comment is still "If you have systemd, remove it and switch to something else"

Mikaela:
I still don't have any issues with systemd, where are the other service files? :smile_cat:

PR #1068, opened 2015-09-07
Fix stickychan joining inaccessable channels.

Zarthus:
Channels that raise this error are unlikely to ever be unblocked, either due to a jupe, or because the name contains illegal characters.

This change unsticks channels that raise numeric 479 - available on the hybrid family of IRCds.

The following output is produced upon this error:
```
(13:27:22) <*stickychan> Stuck #jupedChannel
(13:27:34) <*stickychan> Joining [#jupedChannel]
(13:27:34) <*stickychan> Channel [#jupedChannel] cannot be joined, it is an illegal channel name. Unsticking.
(13:27:34) <*stickychan> Unstuck #jupedChannel
```

The numeric does not seem to have any collisions, according to grawity's irc-docs numerics (extracted from RFC 1459), or https://www.alien.net.au/irc/irc2numerics.html

### DIFF
--- a/modules/stickychan.cpp
+++ b/modules/stickychan.cpp
@@ -191,6 +191,28 @@ public:
 		return false;
 	}
 
+	EModRet OnRaw(CString& sLine) override {
+		CString sNumeric = sLine.Token(1);
+
+		if (sNumeric.Equals("479")) {
+			// ERR_BADCHANNAME (juped channels or illegal channel name - ircd hybrid)
+			// prevent the module from getting into an infinite loop of trying to join it.
+			// :irc.network.net 479 mynick #channel :Illegal channel name
+	
+	                CString sChannel = sLine.Token(3);
+	                for (MCString::iterator it = BeginNV(); it != EndNV(); ++it)
+        	        {
+                	        if (sChannel.Equals(it->first))
+                        	{
+					PutModule("Channel [" + sChannel + "] cannot be joined, it is an illegal channel name. Unsticking.");
+                                	OnUnstickCommand("unstick " + sChannel);
+                        		return CONTINUE;
+				}
+                	}
+		}
+		
+		return CONTINUE;
+	}
 };
 
 


modules/stickychan.cpp
```diff
@@ -191,6 +191,28 @@ class CStickyChan : public CModule
 		return false;
 	}
 
+	EModRet OnRaw(CString& sLine) override {
+		CString sNumeric = sLine.Token(1);
+
+		if (sNumeric.Equals("479")) {
+			// ERR_BADCHANNAME (juped channels or illegal channel name - ircd hybrid)
```
jpnurmi: juped?

Zarthus: Made unavailable by network staff, similarily to blocking a name.

jpnurmi:
This could be fixed in the 1.6.x branch.

Zarthus:
I can re-submit the pull request for that branch if you'd like.

modules/stickychan.cpp
KindOne-: Oh I hate quoting UrbanDictionary, http://www.urbandictionary.com/define.php?term=Jupe

modules/stickychan.cpp
```diff
@@ -191,6 +191,28 @@ class CStickyChan : public CModule
 		return false;
 	}
 
+	EModRet OnRaw(CString& sLine) override {
+		CString sNumeric = sLine.Token(1);
```
jpnurmi: The latest master requests server-time when available so we can no longer assume tokens. I suppose this would be fine in the 1.6.x branch, though.

jpnurmi: Nevermind, I think we rip off tags from OnRaw() for this exact reason. :P

jpnurmi:
Cherry-picked to 1.6.x: https://github.com/znc/znc/commit/b67f9182d60fe9828104cc67402bcdee64f3a492

PR #1071, opened 2015-09-07
It's all about CMessage

jpnurmi:


DarthGandalf:
I'm quite sure such changes will break some weird server (or client)

DarthGandalf:
I've just noticed "get all params params" in a commit message

jpnurmi:
https://github.com/znc/znc/compare/cb84eca...c83a41c

PR #1072, opened 2015-09-07
Add ATTACH command

jpnurmi:
The main benefit is that (unlike JOIN) ATTACH allows wildcards. It's also nicely symmetric with DETACH.

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -258,6 +258,7 @@ private:
 	void EchoMessage(const CMessage& Message);
 
 	std::set<CChan*> MatchChans(const CString& sPatterns) const;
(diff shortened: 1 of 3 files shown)


jpnurmi:
The idea of detaching queries didn't seem too popular, so I've separated the attach-related commits from #1014.

DarthGandalf:
How do you measure popularity of a feature which isn't even merged yet? Especially if we don't have any stats on what features are used by users

KindOne-:
A few people have joined #znc asking how to re-attach to all the channels at once or whatever.

jpnurmi:
https://github.com/znc/znc/compare/1a3e9ecb8660...7f0348489937

PR #1074, opened 2015-09-09
src/znc.cpp: change realname to User of ZNC

### LINKED ISSUE #818: "Got ZNC?" is a little weird real name
At one IRC channel we were discussing the real name and "Got ZNC?" is a little weird. I think that something like "ZNC %VERSION%" (Supybot does this too) would be good default (if expandstrings are allowed there). There was opposition as this would make znc more exposed for targeted vulnerabilities, but it might also make users upgrade faster if that happened.

There was also another advatance of changing the default real name to anything of tracking old ZNC versions.

And of course, users can still set it as whatever they wish.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7838953-got-znc-is-a-little-weird-real-name?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mikaela:
Closes #818

### DIFF
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -702,7 +702,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
 	}
 	CUtils::GetInput("Ident", sAnswer, sUser);
 	vsLines.push_back("\tIdent      = " + sAnswer);
-	CUtils::GetInput("Real name", sAnswer, "Got ZNC?");
+	CUtils::GetInput("Real name", sAnswer, "User of ZNC");
 	vsLines.push_back("\tRealName   = " + sAnswer);
 	CUtils::GetInput("Bind host", sAnswer, "", "optional");
 	if (!sAnswer.empty()) {


jpnurmi:
+1

Mikaela:
Updated based on https://github.com/znc/znc/issues/818#issuecomment-141694766 except I changed the u to U as it seems more proper in realname (how while ZNC is capitalized?).

jpnurmi:
How about just an empty default and make ```CUser::GetRealName()``` do the same thing than ```CUser::GetQuitMsg()``` does?

Mikaela:
I don't understand that, but if you propose to make real name empty, that won't work. https://github.com/znc/znc/issues/818#issuecomment-138973335

Mikaela:
Another update, now it's *User of ZNC*.

Mikaela:
Closed in favour of https://github.com/znc/znc/pull/1109

PR #1075, opened 2015-09-09
travis: only send success emails when state changes

Mikaela:
I think everyone is on the channel anyway so it's not necressary to
receive emails every time the build succeeds. With this change emails
are sent:

* always on build failure.
* when the build is fixed again.

I would personally set `email: false`, but I think it was discussed
before somewhere related to ZNC and wasn't liked.

I am PRing this to 1.6.x, because pull requests also come there and it's
merged into master sooner or later.

As `travis lint` passes,

[CI SKIP]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,3 +82,6 @@ notifications:
             - "irc.freenode.net#znc"
         on_success: always
         on_failure: always
+    email:
+        on_success: change
+        on_failure: always


Mikaela:
Changing IRC notifications, at least the `success` to `change` would also reduce spam at freenode by forks that enable travis.

Zarthus:
:+1:

DarthGandalf:
These values are the default... http://docs.travis-ci.com/user/notifications/

Mikaela:
It's sometimes possible that defaults make more sense than what the user specifies by themselves.

DarthGandalf:
Sorry, I don't understand you.
This PR doesn't change anything, as far as I can see.
And you even agree that defaults make sense.

Mikaela:
> This PR doesn't change anything, as far as I can see.

https://github.com/znc/znc/pull/1075/files

Mikaela:
Oh, I see what you mean, I originally intended to also change IRC notifications.

PR #1087, opened 2015-09-13
route_replies: make the timeout error more clear

Mikaela:


Mikaela:
I am sure I did something horribly wrong, please tell me what and how do I fix it.

1.6.x, because this isn't such a big change and doesn't affect functionality if I did this properly, but I probably didn't.

modules/route_replies.cpp
```diff
@@ -369,7 +369,10 @@ class CRouteRepliesMod : public CModule
 		// The timer will be deleted after this by the event loop
 
 		if (!GetNV("silent_timeouts").ToBool()) {
-			PutModule("This module hit a timeout which is possibly a bug.");
+			PutModule("This module hit a timeout which is possibly a " +
```
jpnurmi: C++ does not allow concatenating raw string literals ("foo" + "bar"). Either write the whole message on the same line, or leave out the '+' mark. I'd prefer the former, because if you choose the latter, then the indentation of the second line also needs to be fixed.

    PutModule("This module hit a timeout which is possibly a connectivity issue.");

modules/route_replies.cpp
```diff
@@ -369,7 +369,10 @@ class CRouteRepliesMod : public CModule
...
+						"connectivity issue.");
+			PutModule("However if you can provide steps to reproduce this " +
+						"issue all the time, plese do report a bug."
```
jpnurmi: Same than above, and missing ```);```at the end of the line

Mikaela:
This pull request is waiting for #1090 or someone to tell me how do I tell Vim to turn those spaces to tabs.

If I wanted to turn everything as spaces I would just `/retab`.

modules/route_replies.cpp
```diff
@@ -369,7 +369,8 @@ class CRouteRepliesMod : public CModule
 		// The timer will be deleted after this by the event loop
 
 		if (!GetNV("silent_timeouts").ToBool()) {
-			PutModule("This module hit a timeout which is possibly a bug.");
+			PutModule("This module hit a timeout which is probably a connectivity issue.");
+			PutModule("However, if you can provide steps to reproduce this issue all the time, please do report a bug.");
```
DarthGandalf: please remove 'all the time'

Mikaela: Sure, but it's trying to communicate "don't report this issue if you just get this message sometimes and have no idea how to get it".

Mikaela: Done.

DarthGandalf: That's what "steps to reproduce" means. But "all the time" isn't even very English.

PR #1089, opened 2015-09-13
Fix #811: Undernet channel key issue

### LINKED ISSUE #811: UnderNet IRC Channel Key Issue
On UnderNet IRC, the channel key is only shown when someone is op'd in the channel.

ZNC gets the channel modes from raw message 324 which sends the channel modes when op'd in the channel.

If the channel is set with mode +tnsk key, and raw message 324 sends the modes in this order, +knst, it just shows +knst *.

If it sends raw message 324 and the modes are in this order, +stnk, it shows +stnk key.

Version 1.4
User Modules Loaded:
chansaver
controlpanel
perform
webadmin

Network Modules Loaded:
autoop
chansaver
keepnick
perform
simple_away

jpnurmi:
Force CChan to be created so that a key passed in the JOIN command is saved.
Otherwise, if CChan didn't exist at this point and we weren't able to save
the key, it would be lost in case the server later presented the key as '*'
when ZNC requested the channel modes.

### DIFF
--- a/modules/chansaver.cpp
+++ b/modules/chansaver.cpp
@@ -77,6 +77,18 @@ public:
 			Channel.SetInConfig(false);
 		}
 	}
+
+	EModRet OnUserJoin(CString& sChannel, CString& sKey) override {
+		CIRCNetwork* pNetwork = GetNetwork();
+		if (pNetwork) {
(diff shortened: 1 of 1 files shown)


jpnurmi:
-1

Hmm, actually, we don't want that CChan object there if the join fails. I'll revise later.

PR #1096, opened 2015-09-18
Travis CI: reduce the amount of builds to 5

jpnurmi:
So all (max 5) builds can run in parallel. This should speedup the CI
rounds a lot.

Candidates:
- Linux/Clang/normal
  We already have address and thread sanitizer -enabled Clang builds on
  Linux, and a normal Clang build on OSX.
- OSX/USE_SYSTEM_OPENSSL
  The system OpenSSL headers have been deprecated since OSX 10.7, and
  are being removed in OSX 10.11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ matrix:
           env: BUILD_TYPE=coverage
         - os: linux
           compiler: clang
-          env: BUILD_TYPE=normal
-        - os: linux
-          compiler: clang
           env: BUILD_TYPE=asan
         - os: linux
           compiler: clang
@@ -22,9 +19,6 @@ matrix:
         - os: osx
           compiler: clang
           env: BUILD_TYPE=normal
-        - os: osx
-          compiler: clang
-          env: BUILD_TYPE=normal USE_SYSTEM_OPENSSL=true
 before_install:
     - "echo os: [$TRAVIS_OS_NAME] build: [$BUILD_TYPE]"
     - if [[ "$BUILD_TYPE" == "normal" ]]; then export CFGFLAGS= MYCXXFLAGS= MYLDFLAGS=; fi
--- a/configure.ac
+++ b/configure.ac
@@ -182,8 +182,6 @@ case "${host_os}" in
 	darwin*)
 		ISDARWIN=1
 
-		AC_ARG_VAR([USE_SYSTEM_OPENSSL],[set USE_SYSTEM_OPENSSL="true" to use the OpenSSL version provided by Mac OS X. Otherwise the Homebrew installed version is used if available. ])
-
 		AC_PATH_PROG([BREW], [brew])
 		if test -n "$BREW"; then
 			# add default homebrew paths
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -22,9 +19,6 @@ matrix:
         - os: osx
           compiler: clang
           env: BUILD_TYPE=normal
-        - os: osx
-          compiler: clang
-          env: BUILD_TYPE=normal USE_SYSTEM_OPENSSL=true
```
DarthGandalf: This option should be removed from configure.ac as well

DarthGandalf:
At some point a new build will be added for sure, and that build would move the number of builds above the threshold 5, so this speedup looks temporary for me.

jpnurmi:
I think leaving USE_SYSTEM_OPENSSL support for there for now is fine. 10.11 hasn't been even released yet, and some people are still happily running 10.5.

The OpenSSL headers have been removed from the OS X v10.11 SDK. (21232069)
https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html

Mikaela:
10.5 hasn't been supported in ages and OS X has given free updates to newer OS X version since a long time ago and I don't think people running EOL operating system should be cared about.

10.11 is currently in Golden Master and it's unlikely for any major changes to happen to it before the release which is also soon, on 2015-09-30.

jpnurmi:
Updated.

I misunderstood the purpose of ```USE_SYSTEM_OPENSSL```. I thought it was useful to some, but it was there just for the CI?

.travis.yml
jpnurmi: Done

jpnurmi:
> At some point a new build will be added for sure, and that build would move the number of builds above the threshold 5, so this speedup looks temporary for me.

I'm happy if it gives faster turnaround times even for a month or two. :)

DarthGandalf:
Who would be so insane to choose system outdated openssl over one from homebrew if both are installed? :)
Yes, it's mostly for CI, to test build on systems where homebrew is not used (and openssl from it is not installed)

Mikaela:
When homebrew isn't installed there is usually macports instead.

.travis.yml
DarthGandalf: No, not done.

jpnurmi:
The same people who run Debian? If it's old, it must be stable!

Mikaela:
```
% lsb_release -sd && openssl version 
Debian GNU/Linux testing (stretch)
OpenSSL 1.0.2d 9 Jul 2015
```

Seems to be the latest :P

configure.ac
```diff
@@ -182,8 +182,6 @@ case "${host_os}" in
 	darwin*)
 		ISDARWIN=1
 
-		AC_ARG_VAR([USE_SYSTEM_OPENSSL],[set USE_SYSTEM_OPENSSL="true" to use the OpenSSL version provided by Mac OS X. Otherwise the Homebrew installed version is used if available. ])
```
jpnurmi: No, not done?

jpnurmi: Ooh :P

jpnurmi:
Fixed

jpnurmi:
https://github.com/znc/znc/commit/f29363752b0b1343bbfd7fb8b717898921bf4dad

PR #1099, opened 2015-09-19
perform: say "number" instead of "nr"

Mikaela:
I think it's inconsistent that perform uses abbreviation "nr" of "number"
while it talks about "command" instead of "cmd".


jpnurmi:
https://github.com/znc/znc/commit/f54ad1f22075940dff3b6ec6d4c39e92425793d7

PR #1101, opened 2015-09-19
perform: add descriptions for commands

Mikaela:
Thanks @ChasedSpade

Closes #1100
Closes #1102

PR #1102, opened 2015-09-19
[perform] Add help text for commands

### LINKED ISSUE #1100: /znc *perform help is missing descriptions
```
<*perform> ------
<*perform> Command: Add <command>
<*perform> ------
<*perform> Command: Del <nr>
<*perform> ------
<*perform> Command: Execute 
<*perform> ------
<*perform> Command: Help search
<*perform> Description: Generate this output
<*perform> ------
<*perform> Command: List 
<*perform> ------
<*perform> Command: Swap <nr> <nr>
<*perform> ------
```

Only the help has Description on what it does, while the messages might seem obvious, having small descriptions on what they do would be nice.

* Adds perform command to be sent to the server on connect
* Delete a perform command
* Send the perform commands to the server now
* *already descripted*
* List the perform rules
* Swap two perform commands

I don't know how to add descriptions for commands and #1090 prevents me from even trying to add new lines.

* * * * *

* Debian GNU/Linux 7.9 (wheezy)
    * ZNC 1.7.x-git-397-2c0dfbb - http://znc.in

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26797869-znc-perform-help-is-missing-descriptions?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dgw:
Fixes #1100, supersedes #1101 (fixes incorrect function calls that omit the arguments)

### DIFF
--- a/modules/perform.cpp
+++ b/modules/perform.cpp
@@ -90,13 +90,15 @@ public:
 	MODCONSTRUCTOR(CPerform) {
 		AddHelpCommand();
 		AddCommand("Add",     static_cast<CModCommand::ModCmdFunc>(&CPerform::Add),
-			"<command>");
+			"<command>", "Add a command to be sent to the server on connect.");
 		AddCommand("Del",     static_cast<CModCommand::ModCmdFunc>(&CPerform::Del),
-			"<nr>");
-		AddCommand("List",    static_cast<CModCommand::ModCmdFunc>(&CPerform::List));
-		AddCommand("Execute", static_cast<CModCommand::ModCmdFunc>(&CPerform::Execute));
+			"<number>", "Delete command <number> in the output of 'List'.");
+		AddCommand("List",    static_cast<CModCommand::ModCmdFunc>(&CPerform::List),
+			"", "List the currently configured commands.");
+		AddCommand("Execute", static_cast<CModCommand::ModCmdFunc>(&CPerform::Execute),
+			"", "Send the commands to the server now.");
 		AddCommand("Swap",    static_cast<CModCommand::ModCmdFunc>(&CPerform::Swap),
-			"<nr> <nr>");
+			"<cmd1> <cmd2>", "Modify the order in which the commands are sent.");
 	}
 
 	virtual ~CPerform() {}


Mikaela:
It seems that you bring even more inconsistence by changing nr to cmd so now the module uses different words for the same thing three times.

See also https://github.com/znc/znc/pull/1099

Mikaela:
I also think you opened this towards wrong branch, as this is bug fix, this should go to 1.6.x in my opinion.

dgw:
I can certainly update this PR to use `<number>` instead of `<cmd1>` and `<cmd2>`, but frankly I think the way I wrote it is clearer.

@jpnurmi can tell me if it's on the wrong branch.

Mikaela:
> I can certainly update this PR to use <number> instead of <cmd1> and <cmd2>, but frankly I think the way I wrote it is clearer.


Which doesn't help as the module would still change between "cmd" and "command" and be inconsistent.

dgw:
Is it really so bad to use abbreviations in some places and not in others? Sometimes it's not worth the extra length to spell everything out, and that's why abbreviations exist.

Mikaela:
> Is it really so bad to use abbreviations in some places and not in others? Sometimes it's not worth the extra length to spell everything out, and that's why abbreviations exist.
￼
Yes, you either abbreviate or you don't, otherwise it looks weird. Not abbreviating is also more clear in my opinion.

dgw:
Force-pushed an updated commit that uses `<number>` instead of `<nr>`. Let's see what others think of `Swap`'s help text.

Mikaela:
You are still talking about command and cmd.

modules/perform.cpp
```diff
@@ -90,13 +90,15 @@ class CPerform : public CModule {
 	MODCONSTRUCTOR(CPerform) {
 		AddHelpCommand();
 		AddCommand("Add",     static_cast<CModCommand::ModCmdFunc>(&CPerform::Add),
-			"<command>");
+			"<command>", "Add a command to be sent to the server on connect.");
```
Mikaela: To be accurate, you talk about command here.

modules/perform.cpp
```diff
@@ -90,13 +90,15 @@ class CPerform : public CModule {
...
-		AddCommand("List",    static_cast<CModCommand::ModCmdFunc>(&CPerform::List));
-		AddCommand("Execute", static_cast<CModCommand::ModCmdFunc>(&CPerform::Execute));
+			"<number>", "Delete command <number> in the output of 'List'.");
+		AddCommand("List",    static_cast<CModCommand::ModCmdFunc>(&CPerform::List),
+			"", "List the currently configured commands.");
+		AddCommand("Execute", static_cast<CModCommand::ModCmdFunc>(&CPerform::Execute),
+			"", "Send the commands to the server now.");
 		AddCommand("Swap",    static_cast<CModCommand::ModCmdFunc>(&CPerform::Swap),
-			"<nr> <nr>");
+			"<cmd1> <cmd2>", "Modify the order in which the commands are sent.");
```
Mikaela: And change it as cmd herẹ. Why you don't either have both as command (which would be more clear) or as `cmd` (which I fear would possibly bring questions on what it means)?

modules/perform.cpp
```diff
@@ -90,13 +90,15 @@ class CPerform : public CModule {
 	MODCONSTRUCTOR(CPerform) {
 		AddHelpCommand();
 		AddCommand("Add",     static_cast<CModCommand::ModCmdFunc>(&CPerform::Add),
-			"<command>");
+			"<command>", "Add a command to be sent to the server on connect.");
 		AddCommand("Del",     static_cast<CModCommand::ModCmdFunc>(&CPerform::Del),
-			"<nr>");
-		AddCommand("List",    static_cast<CModCommand::ModCmdFunc>(&CPerform::List));
-		AddCommand("Execute", static_cast<CModCommand::ModCmdFunc>(&CPerform::Execute));
+			"<number>", "Delete command <number> in the output of 'List'.");
```
Mikaela: This line looks a little strange to me, the command is removed from the commands to be sent to server, not only from the output of 'List'.

dgw: Would it be better as "Delete command <number> from the list."? I'm open to suggestions, but only saying why it looks weird isn't particularly helpful. It leads to spending hours going back and forth over what I think you meant.

modules/perform.cpp
dgw: Why don't I make them both "command"? Because I think that would look far too long. The arguments are going to be mostly single-digit integers, so using two placeholders that are each 8x that length would look weird to me.

modules/perform.cpp
Mikaela: Have you ever used perform? You add or remove commands that are sent to the server when you are connected, you don't only add/remove commands from list.

My suggestion was "Delete a perform command".

modules/perform.cpp
Mikaela: It wouldn't look so long without the tables and having different words is just inconsistent like I have said before.

modules/perform.cpp
dgw: I use perform on about half a dozen networks.

What I asked was whether my suggestion was better than what is currently in this commit and/or what you suggested.

modules/perform.cpp
dgw: I'm sorry, but I still don't see the problem with using abbreviations for the purpose for which they were created.

Mikaela: The problem is not the abbreviations, but changing between command and cmd everywhere.

Windows command interpreter has been cmd for years, but I don't believe it has ever been command. In this pull request you use word "command" and abbreviation "cmd" and I am saying that to be consistent you must either decide that you use "command" or "cmd".

And I am suggesting "command" because it's more clear to not-so-advanced users who e.g. use ZNC provider X.

Mikaela:
There are also merge conflicts here now.

modules/perform.cpp
DarthGandalf: > Windows command interpreter has been cmd for years, but I don't believe it has ever been command

It was `command.com` in MS-DOS.

But I have another comment about this line:
If I have "1. foo" and "2. bar" in the list of commands, after this change, I should send `swap foo bar` instead of `swap 1 2`?

Zarthus: If digits are being swapped, I think it would make much more sense to use `num` (previously `nr`) over `cmd`.  This would prevent users from misleadingly inserting the full command, and make it accurate that we are referring to an identifier.

That said, the idea of first demonstrating `<command>`, and then using `<cmd>`, has both goods and bads. The bad one is that it's not consistent, the good one is that users are learning what it means by looking at previous examples, and the description.

My favour would be to assume users read the description and s/command/cmd on r93, but all in all it is rather irrelevant as a long shot. Consistency can't hurt.

Mikaela: It has also been `number` since my earlier PR got merged https://github.com/znc/znc/commit/f54ad1f22075940dff3b6ec6d4c39e92425793d7.

modules/perform.cpp
Zarthus: I agree. It currently seems to imply `Del` only silences the output from `List`.

I would, at minimum, remove the reference to `output of list`.

dgw:
#1101 was merged, so this is redundant. Closing.

PR #1103, opened 2015-09-19
cert: add description for the "info" command

Mikaela:

PR #1104, opened 2015-09-19
Add descriptions for missing commands in lastseen & certauth & route_replies

Mikaela:
I checked all modules that I had loaded and only these three in addition to perform (https://github.com/znc/znc/pull/1102) and cert (https://github.com/znc/znc/pull/1103) are missing descriptions in commands.

### DIFF
--- a/modules/route_replies.cpp
+++ b/modules/route_replies.cpp
@@ -218,7 +218,7 @@ public:
 
 		AddHelpCommand();
 		AddCommand("Silent", static_cast<CModCommand::ModCmdFunc>(&CRouteRepliesMod::SilentCommand),
-			"[yes|no]");
+			"[yes|no]", "Decides whether to show the timeout messages or not");
 	}
 
 	virtual ~CRouteRepliesMod() {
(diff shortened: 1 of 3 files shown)


Mikaela:
I am not sure if I should squash these commits, but I assume no, because they touch different modules.

In case I should squash, I am waiting for opinion from someone with repo-access.

modules/route_replies.cpp
```diff
@@ -218,7 +218,7 @@ class CRouteRepliesMod : public CModule
 
 		AddHelpCommand();
 		AddCommand("Silent", static_cast<CModCommand::ModCmdFunc>(&CRouteRepliesMod::SilentCommand),
-			"[yes|no]");
+			"[yes|no]"), "Decides whether to show the timeout messages or not";
```
Zarthus: Does this compile? It seems doubtful with the missing brace.

Mikaela: Apparently it did, but it's still wrong.

Mikaela: Fixed.

DarthGandalf:
5 commits for this is certainly too much. During `git rebase -i` you can select what exactly happens with which exact commit. Reordering and `fixup` looks right here.
Though 1 commit would be fine too.

Mikaela:
Done

Issue #1076, opened 2015-09-10
[META] mode +r & topic?


Mikaela:
Currently the channel doesn't require identifying to services, but that could reduce the amount of spam. ZNC has the \*sasl module so more +r shouldn't cause too big issues.

The topic currently says:

> http://znc.in | ZNC 1.6.1: http://wiki.znc.in/ChangeLog/1.6.1 | http://znc.in/nightly | https://github.com/znc/znc | Debian stable: 1.4 is 1.5 year old | 1.2 is 2 years old | 1.0 is 3 years old | Debian oldstable: 0.206 is 3.5 years old | don't randomly PM people | http://blog.max-mehl.com/2014/i-love-znc-because-ilovefs/

I think that at least how old all releases are could be removed or replaced with link to git tags and if mode +r is acceptable mention SASL:

> www.znc.in | ZNC 1.6.1: http://wiki.znc.in/ChangeLog/1.6.1 | http://znc.in/nightly | https://github.com/znc/znc | How old is my ZNC? https://github.com/znc/znc/tags | don't randomly PM people | http://blog.max-mehl.com/2014/i-love-znc-because-ilovefs/ | Trouble joining? http://wiki.znc.in/Sasl

Mikaela:
With bad math and `/who #znc %na` it looks like <100 of 946 aren't identified to services and if they wish to continue being on the channel after netsplit or something removes them, they will surely take a look at identifying.

Mikaela:
And while I am having IRC-monologue there I would also like to point out that I haven't received any more highlight spam than before being visibly op, so I would encourage other ops to get +O flag too.

* https://mikaela.info/english/2015/04/01/keep-the-ops-opped.html

KindOne-:
> ... but that could reduce the amount of spam ....

Clearly you have never seen the totally idle efnet channel.

> .... so more +r shouldn't cause too big issues.

And the random unregistered people that just need to ask one or two questions and never return? Do we just force them to register some nick to ask questions in the channel? Could totally create a ```#znc-unregistered``` and just forward them into that channel and help them in that channel. (I don't like that idea).

> I think that at least how old all releases are could be removed or replaced with link to git tags

I like people on Debian/Ubuntu LTS/etc... with old versions seeing the /topic, knowing they might be on an old version of znc, but everybody knows nobody reads the /topic. /sarcasm 

[...]

PR #1106, opened 2015-09-19
travis: add znc-dev & inform success on change

Mikaela:
I interpret @DarthGandalf's "no strong opinion" and https://github.com/znc/znc/issues/1076#issuecomment-141648088 as why not.

Again PRing to 1.6.x because development also happens there and it will be merged to master at some point giving the change to both.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,5 +80,6 @@ notifications:
     irc:
         channels:
             - "irc.freenode.net#znc"
-        on_success: always
+            - "irc.freenode.net#znc-dev"
+        on_success: change
         on_failure: always


Mikaela:
jpnurmi has been requested to `/cs flags #znc-dev $chanacs:#znc +vVhHoOirtAe`, because freenode doesn't have GroupServ and that is the easiest way to do it.

*permission given to strip Hh which don't exist at freenode, but would get ignored and OV if auto-ops/voices aren't wanted.*

Zarthus:
please don't split up in another channel, just have irc notifications that are a success only tell #znc on `change`.

Mikaela:
> please don't split up in another channel,

it seems it already more or less happened.

>  just have irc notifications that are a success only tell #znc on change.

I would have said that I need opinion from developer for that, but I read the doc and it seems that isn't possible (corrections welcome):

* http://docs.travis-ci.com/user/notifications/#IRC-notification

DarthGandalf:
> it seems it already more or less happened.

It can be reverted if needed...

Zarthus:
it's possible, I use it for all my projects.

```yaml

notifications:
  irc:
    channels:
      - 'irc.esper.net#zarthus'
    template:
      - '%{repository}#%{build_number} (%{branch} - %{commit}) %{author}: %{message} - %{build_url}'
    on_failure: always
    on_success: change
```

Mikaela:
Oh, I misunderstood what you meant.

I thought all notifications to #znc-dev and only changes in status to #znc-dev.

Mikaela:
Changed. In case success should always be sent I can change it back, but as I said in some other PR I prefer success too.

jpnurmi:
> please don't split up in another channel, just have irc notifications that are a success only tell #znc on change.

irc notifications may have triggered this proposal, but in the end it's not really about the notifications. :) the main reasons i suggested a separate dev channel:

- to give developers a chance to discuss development matters without a shout competition with the support discussion that is sometimes quite lively.
- even more importantly, to let developers see what has been discussed while they were away. it's often hard to harvest this information from the chat history of a channel with nearly 1000 users.
- also, to give developers a choice to participate in the community support when they feel so.

Zarthus:
alright, if you think it's for the best. I can definitely see the reasons you bring up.

The only appeal I'd like to make is to ensure all messages of development also go to `#znc`, and not just to `#znc-dev`. (like in this issue)

It may be personal preference, I just liked the setup we had going there and rather not see it change.

Mikaela:
I think this might be better to leave for some developer to commit directly however they wish.

PR #1109, opened 2015-09-20
Resolve #818: "Got ZNC?" is a little weird real name

### LINKED ISSUE #818: "Got ZNC?" is a little weird real name

jpnurmi:


jpnurmi:
https://github.com/znc/znc/commit/6ad7cdb4a2d740804916a41d82d0c6c4c0c1660f

PR #1112, opened 2015-09-21
Move CI notifications over to #znc-dev

jpnurmi:
Double notifications get annoying in the long run, and build results
of individual commits fit the development channel better than a support
channel. Users that want to closely follow the development are welcome
to join the development channel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,6 @@ after_success:
 notifications:
     irc:
         channels:
-            - "irc.freenode.net#znc"
+            - "irc.freenode.net#znc-dev"
         on_success: always
         on_failure: always


Mikaela:
Did anyone ever say if support for writing modules also moves to #znc-dev or not? I personally consider it as ZNC development, but the only thing anyone else has said was @DarthGandalf's "good question".

Mikaela:
Shouldn't this go to 1.6.x or will 1.6.x notifications come to #znc until 1.7.0 is released?

jpnurmi:
> Did anyone ever say if support for writing modules also moves to #znc-dev or not? I personally consider it as ZNC development, but the only thing anyone else has said was @DarthGandalf's "good question".

IMHO asking basic questions related to writing modules is fine on the support channel. If it gets complicated, for instance, they need to know something about the internals, or they request new API, they could be forwarded to the dev channel.

> Shouldn't this go to 1.6.x or will 1.6.x notifications come to #znc until 1.7.0 is released?

This should go to 1.6.x, indeed.

Mikaela:
There is just the issue that (I think) most of people on the basic support channel aren't so familiar with module writing.

To be more accurate, what I am asking is that no one gets thrown out of #znc-dev for asking module-writing questions there?

jpnurmi:
You're getting a bit ahead of yourself. Please give us some time to digest the whole idea and setup things.

> There is just the issue that (I think) most of people on the basic support channel aren't so familiar with module writing.

IMHO there's no issue there. I've seen people tackling such questions well.

> To be more accurate, what I am asking is that no one gets thrown out of #znc-dev for asking module-writing questions there?

Nobody needs to be thrown out from anywhere. Let people ask their questions, and if a development related question remains unanswered, you may hint that the dev channel might be able to help.

Mikaela:
> You're getting a bit ahead of yourself.

Ahead of myself or someone else?

> Please give us some time to digest the whole idea and setup things.

Every time you say this I start feeling #znc-dev is a worse idea than I felt before.

jpnurmi:
@DarthGandalf  and @psychon, are you both ok with this?

DarthGandalf:
As I said earlier somewhere, I don't have a strong opinion about #znc-dev.

But if you want this channel, please also bring other people to it, including those from efnet #znc-dev

PR #1113, opened 2015-09-21
.travis.yml: enable fast_finish

Mikaela:
When fast_finish is True, Travis will inform that the build has failed
when one build fails which is usually desirable instead of waiting for
everything to finish.

However builds that haven't finished yet will continue, this doesn't
stop them or do anything else to them.

`travis lint` passes so,

[CI SKIP]


DarthGandalf:
> [CI SKIP]

Changes to CI are better not to be skipped...

Mikaela:
> Changes to CI are better not to be skipped...

I will try to not in the future then. It just seems stupid to run all builds while only CI is changed and CI test app says that everything is OK.

PR #1119, opened 2015-09-24
Remove awaynick module from /modules

Zarthus:
This module has been deprecated for a year now, and no longer serves an useful function in ZNC

Previously, the code would instruct ZNC to load simple away instead if awaynick was already loaded, and instruct users of its deprecation when tried to load.
But seeing as time has gone by, I think now is the time it would be fair to state
'this module cannot be found' over
'this module is deprecated - see this wiki page' or 'we're loading something else for you.'

It simply removes a bit of boilerplate code that in itself does not really do anything.

With the next release from znc still being far in sight, it might not be unreasonable to suggest this module can be safely removed at this point in time.

---
To summarize the PR in short, all this does is change the following:

```
/znc loadmodule awaynick
(10:35:04) Status Window <-*status> Module [awaynick] aborted: retired module - see http://wiki.znc.in/awaynick
```
to
```
/znc loadmodule awaynick
(10:34:13) Status Window <-*status> Unable to find modinfo [awaynick] [Unable to find module [awaynick]]
```

--

I understand this topic might or might not be an ideal way of handling it, but the change was small enough for it to be a pull request without prior issue discussing the matter, and instead to have conversation initialize here if necessary.

With this all said, I hope I'm not coming across as too eager to remove it. The change is rather abysmal, I just think that most people that care about running znc are already aware on how to load external modules, and that by the time znc stable releases enough time has gone past to make everybody aware of the fact.

In addition to this change, I suggest making the wiki page reference that awaynick is now an external module. The page comes across as a bit hostile, which is something I personally would not want users to look at when they are trying to load a deprecated module.

The only way to now know about it is by google 'awaynick', and the first result would be the wiki page. Which should be honest, but also not try to misdirect users - the module is no longer available by default, but we need not make another Mission Impossible movie just to give the user what they're looking for.


Mikaela:
When I upgrade from 1.4 to 1.7.0 or newer, will my ZNC start without issues or will it die to missing awaynick module?

I also think the log below should be included here:

```
2015-09-24 10:54:59+0300 < Zarthus> is it considered rude to pull request removal of a module
2015-09-24 11:21:24+0300 < Mikaela> Zarthus: I don't think so, but it won't get merged unless it handles removal of the module gracefully (e.g. awaynick loads simple_away and doesn't abort starting up because of missing module after upgrading)
2015-09-24 11:21:41+0300 < Zarthus> i'm doing it as graceful as it deserves to be xP
2015-09-24 11:21:49+0300 < Zarthus> i'm talking about removing awaynick
2015-09-24 11:22:00+0300 < Zarthus> it's been a year, and currently does nothing of interest except load simpleaway
2015-09-24 11:22:15+0300 < Mikaela> I think awaynick will be removed in a few versions
2015-09-24 11:22:37+0300 < Zarthus> i've looked at the code, it is already external
2015-09-24 11:23:44+0300 < Zarthus> the only thing it really does is add complicated boilerplate, and the people who are not yet aware of it being gone will probably never be 
2015-09-24 11:26:07+0300 < Mikaela> but doesn't it unload itself  (and remove itself from autoload) after loading simple_away? If that is the case, I would wait for Debian to go so much forwards to have at least 1.6.0 in stable (and CentOS would also be nice) to avoid any issues with future upgrades
2015-09-24 11:27:40+0300 < Zarthus> as far as I know it does not entirely do that
2015-09-24 11:27:52+0300 < Mikaela> :(
2015-09-24 11:27:57+0300 < Zarthus> the code I'm looking at implies if the module failed to load (i.e. currently loadmodule returns false on loading)
2015-09-24 11:28:03+0300 < Zarthus> when the module cannot be found, the same happens
```

psychon:
Awaynick was removed in 14e09e1a10a62fd91 which is part of ZNC 1.6 (rightt?).
Since debian is widely considered as being too stable:

Debian stable has ZNC 1.4, oldstable has 0.206

So in some years, when a new stable release of debian comes out, it's quite possible that users of awaynick will get the upgrade path that #1120 complains about instead of this nicer one.

Zarthus:
I believe the change was introduced somewhere along 1.5, so the official stable release was 1.6, yes.

Zarthus:
Closing, this is probably best left for a more distant feature seeing as some issues are blocking a smooth transition.

PR #1121, opened 2015-09-24
Better booting messages and errors

Zarthus:
CUtils::PrintStatus with no message parameter just pastes 'ok' in your terminal.

This message is rather useless, and sometimes gets used in repeat.

This change fixes that by being more verbose, telling the user exactly WHAT happened.

New:
```
[ >> ] List of available modules has been found.
[ .. ] Opening config [/home/zarthus/.znc/configs/znc.conf]...
[ >> ] Reading config...
[ .. ] Loading global module [webadmin]...
[ >> ] Global module [webadmin] loaded successfully.
[ .. ] Binding to port [+1111]...
[ >> ] Listening on secure port [+1111].
[ ** ] Loading user [zarthus]
[ ** ] Loading network [freenode]
[ .. ] Adding server [chat.freenode.net +6697 ]...
[ >> ] Added server [chat.freenode.net +6697 ]
[ .. ] Loading user module [chansaver]...
[ >> ] User module [chansaver] loaded successfully.
[ .. ] Loading user module [controlpanel]...
[ >> ] User module [controlpanel] loaded successfully.
[ .. ] Forking into the background...
[ >> ] [pid: 28281]
[ ** ] ZNC 1.7.x-git-442-2a57f69 - http://znc.in
```

Old:
```
[ >> ] ok
[ .. ] Opening config [/home/zarthus/.znc/configs/znc.conf]...
[ >> ] ok
[ .. ] Loading global module [webadmin]...
[ >> ] [/home/zarthus/znc-build/lib/znc/webadmin.so]
[ .. ] Binding to port [+1111]...
[ >> ] ok
[ ** ] Loading user [zarthus]
[ ** ] Loading network [freenode]
[ .. ] Adding server [chat.freenode.net +6697 ]...
[ >> ] ok
[ .. ] Loading user module [chansaver]...
[ >> ] ok
[ .. ] Loading user module [controlpanel]...
[ >> ] ok
[ .. ] Forking into the background...
[ >> ] [pid: 28778]
[ ** ] ZNC 1.7.x-git-441-d66cb36 - http://znc.in
```

### DIFF
--- a/src/IRCNetwork.cpp
+++ b/src/IRCNetwork.cpp
@@ -448,7 +448,9 @@ bool CIRCNetwork::ParseConfig(CConfig *pConfig, CString& sError, bool bUpgrade) 
 	pConfig->FindStringVector("server", vsList);
 	for (const CString& sServer : vsList) {
 		CUtils::PrintAction("Adding server [" + sServer + "]");
-		CUtils::PrintStatus(AddServer(sServer));
+		
+		bool bResult = AddServer(sServer);
+		CUtils::PrintStatus(bResult, bResult ? "Added server [" + sServer + "]" : "Failed to add server [" + sServer + "].");
 	}
 
 	pConfig->FindStringVector("trustedserverfingerprint", vsList);
--- a/src/User.cpp
+++ b/src/User.cpp
@@ -437,7 +437,7 @@ bool CUser::ParseConfig(CConfig* pConfig, CString& sError) {
 
 		bool bModRet = LoadModule(sModName, sArgs, sNotice, sModRet);
 
-		CUtils::PrintStatus(bModRet, sModRet);
+		CUtils::PrintStatus(bModRet, bModRet ? "User module [" + sModName + "] loaded successfully." : sModRet);
 		if (!bModRet) {
 			// XXX The awaynick module was retired in 1.6 (still available as external module)
 			if (sModName == "awaynick") {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -294,7 +294,7 @@ int main(int argc, char** argv) {
 		pZNC->GetModules().GetAvailableMods(ssUserMods, CModInfo::UserModule);
 		pZNC->GetModules().GetAvailableMods(ssNetworkMods, CModInfo::NetworkModule);
 		if (ssGlobalMods.empty() && ssUserMods.empty() && ssNetworkMods.empty()) {
-			CUtils::PrintStatus(false, "");
+			CUtils::PrintStatus(false, "Failed to find any modules.");
 			CUtils::PrintError("No modules found. Perhaps you didn't install ZNC properly?");
 			CUtils::PrintError("Read http://wiki.znc.in/Installation for instructions.");
 			if (!CUtils::GetBoolInput("Do you really want to run ZNC without any modules?", false)) {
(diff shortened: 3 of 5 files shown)


Mikaela:
```
[ >> ] List of available modules has been found.
```

* I think this string could have some improvement.

```
[ .. ] Opening config [/home/zarthus/.znc/configs/znc.conf]...
[ >> ] Reading config...
[ .. ] Loading global module [webadmin]...
[ >> ] [/home/zarthus/znc-build/lib/znc/webadmin.so]
```

* How about `Loaded module webadmin from /home/zarthus/znc-build/lib/znc/webadmin.so`?

```
[ .. ] Binding to port [+1111]...
[ >> ] Listener on port [+1111] has been added.
```

* I know that ZNC talks about listeners, but for other users `Listening on SSL port 1111` would be more clear or at least `Listening on port +1111`.

```
<snip>
[ .. ] Loading user module [chansaver]...
[ >> ] Module [chansaver] loaded successfully.
[ .. ] Loading user module [controlpanel]...
[ >> ] Module [controlpanel] loaded successfully.
<snip>
```

* Where did the paths disappear?

Zarthus:
>I think this string could have some improvement.
I'm turning some blanks. None of what I can think of can be phrased in a way that would say "your installation does not appear to be horribly fucked." 

>How about Loaded module webadmin from /home/zarthus/znc-build/lib/znc/webadmin.so?
Will be improved in my next commit to follow in line.

>`[ >> ] Listener on port [+1111] has been added.`
Changed to `Listening on secure port [+1111]`

>Where did the paths disappear?
Not sure if this problem should be tackled in this pull request, a followup where the three modules get standard outputs might be nice.

psychon:
Some history on where these "[ ok ]" come from.
Once upon a time we used some magic to overwrite the "[ .. ]" at the beginning of the line with "[ ok ]". That magic (mostly just a `\r`) caused problems and was removed.

Having said that, I wonder: Can we just remove these "ok" messages? If no error is reported, then we can assume that things worked successfully.

Alternative idea:

    Checking for list of available modules...ok

(Instead of printing the "ok" on its own line, append it to the previous line)

Zarthus:
I like the idea of removing CRLF for the 'ok' messages, but since it was used in CUtils I wasn't sure how often and where it was used, and what it would affect. So I took the safer and more verbose approach.

A snippet from the dev channel:
```
(13:15:43) <Zarthus> znc is really all over the place and it's difficult to find all 'unintentional' empty strings being passed
(13:16:04) <Zarthus> for a while I considered passing a third optional parameter that would determine a default message
(13:16:23) <Zarthus> the other thing I considered is going "the linux way" and not giving output if all is okay
(13:16:47) <Zarthus> but all of it boils down to making a workaround for the existing problem
```

In addition to removing them, znc's terminal uses several states (error, success, and pending). And it would look weird to see everything pending and no real green.

I don't know for sure what the most favourable approach really is, but I think the one that will satisfy the most people is to just give more of a description of what happened. I also vaguely recall seeing multiple "ok"s in a row, so by removing \r\n you might end up with `Doing something awesome... okokok`

Mikaela:
> `Checking for list of available modules...ok`

I like this idea.

DarthGandalf:
Um, this seems to conflict with #1124

Zarthus:
@DarthGandalf: Correct, I kept both PR's open in case users preferred one way over another.  I stated the conflict in issue 1124 -- sorry if that was not clear enough

With that said, this PR should probably be closed.

PR #1139, opened 2015-10-02
Expandstring network, alphabeticize expandstring list

### LINKED ISSUE #1115: Expandstring %network%
%network% would expand to name of the current network in znc, example use case would be `/msg *watch add * *highlight-%network% *%nick%*` when https://github.com/znc/znc/issues/514 is done and watch has became user-level module.

* * * * *

* Debian GNU/Linux 7.9 (wheezy)
* ZNC 1.7.x-git-441-d66cb36 - http://znc.in

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26861924-expandstring-network?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Zarthus:
Allows use of %network% in ExpandString for CIRCNetwork.

Additionally, the alphabeticizing was just a nice touch. I found it difficult to navigate the existing list.

Fixes #1115

### DIFF
--- a/src/IRCNetwork.cpp
+++ b/src/IRCNetwork.cpp
@@ -1434,12 +1434,14 @@ CString CIRCNetwork::ExpandString(const CString& sStr) const {
 
 CString& CIRCNetwork::ExpandString(const CString& sStr, CString& sRet) const {
 	sRet = sStr;
-	sRet.Replace("%defnick%", GetNick());
-	sRet.Replace("%nick%", GetCurNick());
+
 	sRet.Replace("%altnick%", GetAltNick());
-	sRet.Replace("%ident%", GetIdent());
-	sRet.Replace("%realname%", GetRealName());
 	sRet.Replace("%bindhost%", GetBindHost());
+	sRet.Replace("%defnick%", GetNick());
+	sRet.Replace("%ident%", GetIdent());
+	sRet.Replace("%network%", GetName());
+	sRet.Replace("%nick%", GetCurNick());
+	sRet.Replace("%realname%", GetRealName());
 
 	return m_pUser->ExpandString(sRet, sRet);
 }
--- a/src/User.cpp
+++ b/src/User.cpp
@@ -543,17 +543,20 @@ CString& CUser::ExpandString(const CString& sStr, CString& sRet) const {
 	CString sTime = CUtils::CTime(time(nullptr), m_sTimezone);
 
 	sRet = sStr;
-	sRet.Replace("%user%", GetUserName());
-	sRet.Replace("%defnick%", GetNick());
-	sRet.Replace("%nick%", GetNick());
 	sRet.Replace("%altnick%", GetAltNick());
-	sRet.Replace("%ident%", GetIdent());
-	sRet.Replace("%realname%", GetRealName());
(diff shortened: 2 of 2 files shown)


src/User.cpp
```diff
@@ -543,18 +543,19 @@ CString& CUser::ExpandString(const CString& sStr, CString& sRet) const {
 	CString sTime = CUtils::CTime(time(nullptr), m_sTimezone);
 
 	sRet = sStr;
-	sRet.Replace("%user%", GetUserName());
-	sRet.Replace("%defnick%", GetNick());
-	sRet.Replace("%nick%", GetNick());
 	sRet.Replace("%altnick%", GetAltNick());
+	sRet.Replace("%bindhost%", GetBindHost());
+	sRet.Replace("%defnick%", GetNick());
+	sRet.Replace("%empty%", "");
```
DarthGandalf: I think %empty% should be in the very end, to make it possible to write `%ni%empty%ck%` as a workaround to put a non-expanded `%nick%`. If such usage is to be supported at all...

src/Modules.cpp
```diff
@@ -184,6 +184,8 @@ CString CModule::ExpandString(const CString& sStr) const {
 CString& CModule::ExpandString(const CString& sStr, CString& sRet) const {
 	sRet = sStr;
 
+        sRet.Replace("%network%", GetNetwork()->GetName());
```
DarthGandalf: Yay for crashing on user modules.

DarthGandalf: This belongs to CIRCNetwork::ExpandString, not here.

src/User.cpp
Zarthus: I guess it's better to have than not to have, I will update it in a followup commit.

src/IRCNetwork.cpp
```diff
@@ -1434,12 +1434,14 @@ CString CIRCNetwork::ExpandString(const CString& sStr) const {
...
 CString& CIRCNetwork::ExpandString(const CString& sStr, CString& sRet) const {
 	sRet = sStr;
-	sRet.Replace("%defnick%", GetNick());
-	sRet.Replace("%nick%", GetCurNick());
+
 	sRet.Replace("%altnick%", GetAltNick());
+	sRet.Replace("%bindhost%", GetBindHost());
+	sRet.Replace("%defnick%", GetNick());
 	sRet.Replace("%ident%", GetIdent());
+	sRet.Replace("%network%", this->GetName());
```
DarthGandalf: You don't need `this->`, as `GetAltNick()` doesn't need it.

DarthGandalf:
Please notify after you make the commit, github notifies only about comments

Zarthus:
Yes, I was going to notify you on irc but hadn't found the time yet, next
time I'll leave a comment.

On 15 Oct 2015 9:31 am, "Alexey Sokolov" <notifications@github.com> wrote:
>
> Please notify after you make the commit, github notifies only about
comments

PR #1148, opened 2015-10-12
Fix #1147: show new server name when jumping

### LINKED ISSUE #1147: Feature request: add server details in *status messages

         me jump

jreese:
This updates the connect command in *status to retrieve the next server
object before triggering the jump, thereby allowing it to display the
next server's name rather than a generic message.


jreese:
Going to submit a new PR with different commit details.

DarthGandalf:
You don't need a new PR for that...

PR #1150, opened 2015-10-13
Fix #1133: Add MinClients option to simple_away

### LINKED ISSUE #1133: Add client_count_less_than option to simple_away
Assume you use multiple IRC clients, of which one is always connected no matter what. If you could tell simple_away to always set you away whenever the number of connected clients drops below the configurable number N instead of N being hardcoded to 1, this would be great. The push module <a href="https://github.com/jreese/znc-push#conditions">has this</a>.

<b>Why would you need this?</b>
I have one IRC client that's always connected because it's a webchat client running on the same server as my znc (<a href="http://shout-irc.com">shout</a> for those who are interested), so I can use my bouncer from a browser when I need to. Combined with the wonderful <a href="http://wiki.znc.in/Clientbuffer">clientbuffer</a> module, this is a great workaround for the (yet) non-existent ZNC webchat. The only downside is that you're not automatically set away anymore when you close your (other) IRC clients, because shout stays connected even when you close your browser.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26990972-add-client_count_less_than-option-to-simple_away?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jreese:
This adds a new option and command to the module, "MinClients", that
specifies the minimum number of clients that must be connected for the
module to considered the user as available.  If fewer clients are
connected than the minimum, the user will be marked away, even if there
are still clients attached.  The default value is "1", which replicates
the existing behavior of the module.

To modify the MinClients value, `/msg *simple_away MinClients [<value>]`
is available.  Running the command without parameters will display the
current value instead of modifying it.

### DIFF
--- a/modules/simple_away.cpp
+++ b/modules/simple_away.cpp
@@ -39,6 +39,7 @@ class CSimpleAway : public CModule {
 private:
 	CString      m_sReason;
 	unsigned int m_iAwayWait;
+	unsigned int m_iMinClients;
 	bool         m_bClientSetAway;
 	bool         m_bWeSetAway;
 
@@ -46,6 +47,7 @@ public:
 	MODCONSTRUCTOR(CSimpleAway) {
 		m_sReason        = SIMPLE_AWAY_DEFAULT_REASON;
 		m_iAwayWait      = SIMPLE_AWAY_DEFAULT_TIME;
+		m_iMinClients    = 1;
 		m_bClientSetAway = false;
 		m_bWeSetAway     = false;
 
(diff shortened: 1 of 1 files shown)


modules/simple_away.cpp
```diff
@@ -141,6 +150,15 @@ class CSimpleAway : public CModule {
 		PutModule("Timer disabled");
 	}
 
+	void OnMinClientsCommand(const CString& sLine) {
+		if (sLine.Token(1).empty()) {
+			PutModule(CString(m_iMinClients));
```
DarthGandalf: Please make the response more verbose

modules/simple_away.cpp
```diff
@@ -204,6 +227,11 @@ class CSimpleAway : public CModule {
 		m_iAwayWait = iAwayWait;
 	}
 
+	void SetMinClients(unsigned int iMinClients, bool bSave = true) {
```
Zarthus: It would be nice if it were rejected when user input was malformed. i.e. `if (iMinClients == 0 || iMinClients > 100) { ... }`, it shouldn't actually change anything but I think it'd add to the polish.

modules/simple_away.cpp
```diff
@@ -141,6 +150,15 @@ class CSimpleAway : public CModule {
 		PutModule("Timer disabled");
 	}
 
+	void OnMinClientsCommand(const CString& sLine) {
+		if (sLine.Token(1).empty()) {
+			PutModule(CString(m_iMinClients));
+		} else {
+			SetMinClients(sLine.Token(1).ToUInt());
+			PutModule("MinClients set");
```
Zarthus: The existing module returns feedback with the number of seconds it would set to, for consistency, I think it would be nice to have this too return the number of clients it was set to. Like my previous comment, it would just add to the polish, but overall not change a whole lot.

PR #1152, opened 2015-10-15
Revert "Fix #1147: show new server name when jumping"

### LINKED ISSUE #1147: Feature request: add server details in *status messages

DarthGandalf:
Reverts znc/znc#1149

This made it to jump over the next server, skipping one (and actually lying to user about what server it connects to)


DarthGandalf:
@jreese

PR #1154, opened 2015-10-15
Fix #1147: show new server name when jumping

### LINKED ISSUE #1147: Feature request: add server details in *status messages

jreese:
This updates the connect command in *status to retrieve the next server
object before triggering the jump, thereby allowing it to display the
next server's name rather than a generic message.

### DIFF
--- a/include/znc/IRCNetwork.h
+++ b/include/znc/IRCNetwork.h
@@ -120,7 +120,7 @@ public:
 	bool DelServer(const CString& sName, unsigned short uPort, const CString& sPass);
 	bool AddServer(const CString& sName);
 	bool AddServer(const CString& sName, unsigned short uPort, const CString& sPass = "", bool bSSL = false);
-	CServer* GetNextServer();
+	CServer* GetNextServer(bool bAdvance = true);
 	CServer* GetCurrentServer() const;
 	void SetIRCServer(const CString& s);
 	bool SetNextServer(const CServer* pServer);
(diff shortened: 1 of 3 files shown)


include/znc/IRCNetwork.h
```diff
@@ -120,7 +120,7 @@ class CIRCNetwork {
 	bool DelServer(const CString& sName, unsigned short uPort, const CString& sPass);
 	bool AddServer(const CString& sName);
 	bool AddServer(const CString& sName, unsigned short uPort, const CString& sPass = "", bool bSSL = false);
-	CServer* GetNextServer();
+	CServer* GetNextServer(const bool pAdvance = true);
```
DarthGandalf: It's bool, not pointer. bAdvance.
const doesn't change anything here, but ok

src/IRCNetwork.cpp
```diff
@@ -1151,7 +1151,11 @@ CServer* CIRCNetwork::GetNextServer() {
 		m_uServerIdx = 0;
```
DarthGandalf: Sometimes it still advances?

DarthGandalf:
Also please replace "Fix #1147" in commit message with "See #1147", as it's not a full fix of that bug.

src/IRCNetwork.cpp
jreese: Since this only moves the pointer when it already points past the end of the list, it's technically not advancing the index, just setting it to where it "should be".  Unless I'm missing something, the next call to GetNextServer with `bAdvance == true` would have just set the index to zero anyways, and this way we can return what that next server would have been without having to do a special check.

PR #1166, opened 2015-10-24
znc.service.in: add datadir & instructions

### LINKED ISSUE #1165: znc.service.in only works with ~/.znc
...which will fail on all systems where znc USER has its HOME set to /dev/null. of course one could override this with /etc/systemd/system/znc.service/override.conf but that's a just a workaround. that is "works out of box" for systemd is broken.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27617215-znc-service-in-only-works-with-znc?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mikaela:
Closes #1165

### DIFF
--- a/znc.service.in
+++ b/znc.service.in
@@ -3,7 +3,11 @@ Description=ZNC, an advanced IRC bouncer
 After=network-online.target
 
 [Service]
-ExecStart=@bindir@/znc -f
+# To create user znc:
+# `sudo useradd -m -d /var/lib/znc -r -s /sbin/nologin -U znc`
+# To create the config file:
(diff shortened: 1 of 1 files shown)


Mikaela:
It seems that AppVeyor fixed itself after `git pull --rebase upstream master`. It was broken before even if this PR doesn't touch anything Windows uses as far as I am aware.

PR #1171, opened 2015-10-26
Make $USER and $NETWORK case sensitive in logs

walliski:
@JackHarley noticed this while playing around with the znc-backlog module. (https://github.com/FruitieX/znc-backlog/issues/13), and old PR: #590 

While it is indeed true that IRC is not case sensitive, ZNC still thinks that users and the configured names for networks are. This means that the log path is different depending on which scope the module is loaded in. As a global module both $USER and $NETWORK would be lowercase, but loaded as Network module they would be whatever casing they are in ZNC configs. Issue being difficulty to handle the different scenarios in other modules (znc-backlog) or external scripts.

This will break log folders for people once again.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -233,8 +233,8 @@ void CLogMod::PutLog(const CString& sLine, const CString& sWindow /*= "Status"*/
 
 	// TODO: Properly handle IRC case mapping
 	// $WINDOW has to be handled last, since it can contain %
-	sPath.Replace("$USER", CString((GetUser() ? GetUser()->GetUserName() : "UNKNOWN")).AsLower());
-	sPath.Replace("$NETWORK", CString((GetNetwork() ? GetNetwork()->GetName() : "znc")).AsLower());
+	sPath.Replace("$USER", CString((GetUser() ? GetUser()->GetUserName() : "UNKNOWN")));
+	sPath.Replace("$NETWORK", CString((GetNetwork() ? GetNetwork()->GetName() : "znc")));
 	sPath.Replace("$WINDOW", CString(sWindow.Replace_n("/", "-").Replace_n("\\", "-")).AsLower());
 
 	// Check if it's allowed to write in this specific path


Mikaela:
It sounds like this will just break logging for everyone and annoy people instead of fixing the real issue.

> However ZNC seems to allow creation of multiple users with the same name, different case, (Walliski and walliski), which means the logs would go into the wrong folder if they are always lower case, such as it is now. Oh man this is a mess... It also seems to break with the Network name in case the log module is loaded on user level. 

https://github.com/FruitieX/znc-backlog/issues/13#issuecomment-151297951

This is the bug. I see no reason why ZNC would allow creating of two users/networks with the same name, but different capitalization. 

If I used identfile and created two users with the same name, but different capitalization and one of them was abusive, they would both get banned from the network as the IRCd doesn't care about case difference (or is stupid if it does).

JackHarley:
@Mikaela Whilst this is correct, right now there is inconsistency in the ZNC codebase, the log module creates all lowercase folders, whilst ZNC core creates users and networks with their correct capitalisations.

Unless you want to start lowercasing every user and network folder that ZNC stores (which will require an update script in order to not completely break everyone's installations), the best solution is to just remove the completely unnecessary network and user lowercase conversions that the log module performs. It's only logs so whilst it is slightly irritating for the folders to switch over, it won't actually break anything since the vast majority of people either load the log module at the network level (in which case they are not effected at all) or read logs manually (in which case the logs they need will still be there).

Mikaela:
> it won't actually break anything since the vast majority of people either load the log module at the network level (in which case they are not effected at all) or read logs manually (in which case the logs they need will still be there).

there are also log parses such as jsis and pisg.

walliski:
I'd still say that the log modules inconcistency causes more problems for people that want to do stuff with the logs, and it should be fixed. Runnig Pisg myself so I'll have to move around files and stuff like that, but it makes it easier in the long run.

Also agree that ZNC should block making users with similar names except from the capitalisation. How the folders then are implemented is another question, to lowercase everything or to keep it as is...

But I'd also say that

JackHarley:
Yeah I've got to agree with @walliski on this, the inconsistency isn't sustainable going forward. It will need to be addressed at some point so we may as well do it now.

walliski:
We getting anywhere with this?
@DarthGandalf @jpnurmi @kylef Anyone wanna comment?

PR #1172, opened 2015-10-27
Update add_edit_network.tmpl

Roobarb:
Explicitly mention SHA-256 certificate fingerprints, since for example SHA-1 fingerprints are not accepted and will cause verification to fail.


dgw:
s/certficate/certificate/

Might as well include a & before CA as long as the string is being edited anyway, for grammar's sake.

Otherwise, :+1:

DarthGandalf:
Months without the pointed out fix.

PR #1176, opened 2015-11-03
Add test for controlpanel module

KindOne-:


DarthGandalf:
> Connection refused

That's why other tests wait until ZNC connects to the IRCd before trying to connect to ZNC as a client.

test/Integration.cpp
```diff
@@ -383,6 +383,867 @@ TEST_F(ZNCTest, InvalidConfigInChan) {
...
+	client.ReadUntil(response + "User [KindOne] added!");Z;
+
+	client.Write(request + "AddUser KindOne hunter2");
+	client.ReadUntil(response + "Error: User [KindOne] already exists!");Z;
+
+	client.Write(request + "AddCTCP user VERSION Test");
+	client.ReadUntil(response + "Added!");Z;
+
+	client.Write("PRIVMSG nick :\x01VERSION\x01");
+	ircd.Write(":NOTICE nick :mIRC 7.42");Z;
```
DarthGandalf: What's this?

KindOne-: Looks like a mistake. I'll remove it.

DarthGandalf: This doesn't test anything, be it "Test" or "mIRC" or something else. If it did test something, the test would probably fail. Good test should fail when the thing which it tests behaves differently from what the test expects.

This whole test now tests only interaction with user, but it doesn't test that the settings are actually applied and used somewhere. Please add a TODO to add that.
Then a real test that CTCP VERSION does what it's supposed to do can be added later.

KindOne-: Done.

PR #1183, opened 2015-11-15
Add test for autoop and autovoice modules

KindOne-:


DarthGandalf:
May you test that it actually autoops people, not just UI?

PR #1192, opened 2015-11-24
support for separate SSLKeyFile configuration

nospampleasemam:
This pull request adds backwards compatible support for a new configuration variable, `SSLKeyFile`, allowing separation of SSL keys and cert files. It depends on https://github.com/jimloco/Csocket/pull/58 (please note that CI builds are expected to fail for this reason). Documentation and `--makecert` support should still be added, but I wanted to make this request first for feedback.

Eventually I would like to add support for an initial persistent read of the files into memory, which will allow ZNC to run initially as a root service, read protected keys and immediately drop privileges to the znc user. This will be particularly helpful to support https://letsencrypt.org certificates without a wrapper.

### DIFF
--- a/include/znc/znc.h
+++ b/include/znc/znc.h
@@ -128,6 +128,8 @@ public:
 	CString GetUserPath() const;
 	CString GetModPath() const;
 	CString GetPemLocation() const;
+	CString GetKeyLocation() const;
+	CString GetDHParamLocation() const;
 	const CString& GetConfigFile() const { return m_sConfigFile; }
 	bool WritePemFile();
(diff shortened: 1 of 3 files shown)


DarthGandalf:
I think --makepem should be updated too.

nospampleasemam:
I've added support for separate dhparams, and added support for this configuration to `--makeconf`. However, I don't see the need to update `--makepem`, as the old format is still supported; a user that wants to autogenerate a cert likely does not care about the layout.

DarthGandalf:
> However, I don't see the need to update --makepem, as the old format is still supported; a user that wants to autogenerate a cert likely does not care about the layout.

That's true.
I thought about which layout is the "recommended" one, single file or separate files.

PR #1195, opened 2015-11-26
Show client identification in notify_connect module

BlaXpirit:
If someone connects to ZNC as `username@identifier/network`, the module will now show that *identifier*. When the identifier is not provided, the behavior is unchanged.

> [Notice] -\*status- \*\*\* username@identifier detached (from 127.0.0.1)

### DIFF
--- a/modules/notify_connect.cpp
+++ b/modules/notify_connect.cpp
@@ -22,16 +22,27 @@ public:
 	MODCONSTRUCTOR(CNotifyConnectMod) {}
 
 	void OnClientLogin() override {
-		SendAdmins(GetUser()->GetUserName() + " attached (from " + GetClient()->GetRemoteIP() + ")");
+		NotifyAdmins("attached");
 	}
 
 	void OnClientDisconnect() override {
-		SendAdmins(GetUser()->GetUserName() + " detached (from " + GetClient()->GetRemoteIP() + ")");
+		NotifyAdmins("detached");
 	}
 
 private:
 	void SendAdmins(const CString &msg) {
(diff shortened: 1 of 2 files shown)


DarthGandalf:
May you add a test for it?
At this point we have enough infrastructure to make this easy (`make test2` and `test/Integration.cpp`)

BlaXpirit:
If this module had tests, I would gladly extend them. But I have no idea how to add new ones.

Mikaela:
[Clientnotify](http://wiki.znc.in/Clientnotify) could probably also use this change as I don't think it has identifier either.

BlaXpirit:
@Mikaela Interesting. This change is even more relevant to *clientnotify*, and the module is more relevant to me, but I wasn't aware of it!

Mikaela:
I believe that the two modules  should be merged and same with automode/autoop. They do so similar thing and better than two separate modules would be just having options "notify me (admin) when someone else connects" or "notify me when other clients login as me". But that is a separate issue.

BlaXpirit:
I investigated the testing facilities.
So `make test2` requires `sudo make install`? No, thanks :neutral_face:

DarthGandalf:
why sudo?
./configure --prefix=/tmp/znc

2015-11-27 9:44 GMT+00:00 Oleh Prypin <notifications@github.com>:

DarthGandalf:
Oh, I've noticed you added the test...
Please comment in such cases; github doesn't notify about new commits in PR, but does about new comments.

DarthGandalf:
Merged with this change: d7363971946c17e17aaac2fdd5a50020ded87270

Issue #948, opened 2015-04-21
Support for SNI


rcombs:
ZNC should use `SSL_set_tlsext_host_name` and the related OpenSSL APIs to indicate the expected server name, so TLS can work as expected with DNS round robins (assuming properly-configured servers, which is a big assumption… but hey, can't hurt to be some of the first ones to be well-behaved)

DarthGandalf:
What would be the use of it in IRC?
With HTTP, this is needed, because HTTP servers support multiple sites one the same IP.

rcombs:
An IRC server might accept connections on multiple hostnames. For instance, one very common practice is to have one master hostname (`irc.<network>.net` or the like) that routes to any of a network's servers, and also have each individual server have its own hostname. SNI support would allow servers to return either a cert for the round-robin domain, or for their individual name, depending on what the client's asking for.

DarthGandalf:
Why would client want to sometimes connect to any server in the network, and sometimes to some specific server, which has a different hostname, so that wildcard cert doesn't work for such different hostname?

teward:
> On Apr 21, 2015, at 14:31, Alexey Sokolov <notifications@github.com> wrote:
> 
> What would be the use of it in IRC?
> With HTTP, this is needed, because HTTP servers support multiple sites one the same IP.
> 
> —
> Reply to this email directly or view it on GitHub.
> 

There is no need for it.

SNI is a web server thing.  IRC servers are only permitted to serve the one hostname/cert combo if I remember right.

And for ZNC webadmin there is no need for it as znc.pem is a singular file, and it could only serve one very (if I am wrong on that let me know).

teward:
And to answer the parts that only just hit my email, @DarthGandalf, there are some IRC networks with disjointed hostname a which end up having different certs, or private networks which have no wildcard, and other 'edge cases'.

I still don't see a value in enabling SNI which is a web server thing usually.

Mikaela:
I think this should possibly be reconsidered based on https://github.com/ircv3/ircv3-specifications/pull/155.

glyph:
[...]

PR #1200, opened 2015-12-04
Add SNI client support

Adam-:


src/Socket.cpp
```diff
@@ -126,6 +126,12 @@ void CZNCSock::SSLHandShakeFinished() {
 	CallSockError(errnoBadSSLCert, sErrorMsg);
 	Close();
 }
+
+bool CZNCSock::SNIConfigureClient(CS_STRING & sHostname)
```
DarthGandalf: It's `CString&` in ZNC.
Also please follow the style around.

DarthGandalf:
#948 

What happens if `m_HostToVerifySSL` is IP instead of hostname?

Adam-:
> It's CString& in ZNC.

Several of the related declarations in CZNCSock use the CS_STRING & format but sure I'll change it.

> What happens if m_HostToVerifySSL is IP instead of hostname?

That doesn't matter, the server name sent would just be the IP.

DarthGandalf:
> Several of the related declarations in CZNCSock use the CS_STRING & format but sure I'll change it.

I meant not just the space, but CString vs CS_STRING, and placement of {
The only other instance of CS_STRING should be fixed though, thanks for pointing it out.

PR #1205, opened 2015-12-13
Add support for CIDR notation in allowed hosts list.

### LINKED ISSUE #207: Make Allowed IPs to work with CIDR subnets instead of string wildcards [$5]
<bountysource-plugin>
---
There is a **[$5 open bounty](https://www.bountysource.com/issues/654-make-allowed-ips-to-work-with-cidr-subnets-instead-of-string-wildcards-5?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DinCahill:
Fixes #207 

This is my first time writing C++. Please let me know if anything looks dodgy.
A non-exhaustive bunch of test cases all seem to work. I've also left wildcard matching in there.

I also have a patch for 1.6.x, but I don't know if that's desired.


DinCahill:
Oops. Fixed those memory leaks. I hadn't tried building with clang.

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+    };
+    for (int i = 0; i < (int)(sizeof(aHostTests) / sizeof(aHostTests[0]));
+         ++i) {
+        m_pTestUser = new CUser("user");
+        hostTest* h = aHostTests + i;
+        m_pTestUser->AddAllowedHost(h->sTestHost);
+        CString should = h->bExpectedResult ? "" : " not";
+        CString error = "Allow-host string \"" + h->sTestHost + "\" should" +
+                        should + " allow host \"" + h->sIP + "\"";
+        EXPECT_EQ(m_pTestUser->IsHostAllowed(aHostTests[i].sIP),
```
DarthGandalf: Unfortunately, EXPECT_EQ accepts expected first, then actual, otherwise the error message is broken (reverse)

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+        {"::0/0", "::g", false},
+
+        {"::0/0/0", "::0", false},
+        {"2001:db8::/33", "2001:db9:0::", false},
+        {"2001:db8::/32", "2001:db8:8000::", true},
+        {"2001:db8::/33", "2001:db8:8000::", false},
+    };
+    for (int i = 0; i < (int)(sizeof(aHostTests) / sizeof(aHostTests[0]));
+         ++i) {
+        m_pTestUser = new CUser("user");
```
DarthGandalf: Just create it on stack, or at least with unique_ptr

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+        {"127.0.0.1/0", "127.0.0.a", false},
+        {"::0/0", "::g", false},
+
+        {"::0/0/0", "::0", false},
+        {"2001:db8::/33", "2001:db9:0::", false},
+        {"2001:db8::/32", "2001:db8:8000::", true},
+        {"2001:db8::/33", "2001:db8:8000::", false},
+    };
+    for (int i = 0; i < (int)(sizeof(aHostTests) / sizeof(aHostTests[0]));
+         ++i) {
```
DarthGandalf: `for (const hostTest& ht : aHostTests)`

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+        m_pTestUser = new CUser("user");
+        hostTest* h = aHostTests + i;
+        m_pTestUser->AddAllowedHost(h->sTestHost);
+        CString should = h->bExpectedResult ? "" : " not";
+        CString error = "Allow-host string \"" + h->sTestHost + "\" should" +
+                        should + " allow host \"" + h->sIP + "\"";
+        EXPECT_EQ(m_pTestUser->IsHostAllowed(aHostTests[i].sIP),
+                  aHostTests[i].bExpectedResult)
+            << error;
```
DarthGandalf: `<< "Allow-host is " << h->sTestHost;`

Built-in error of EXPECT_EQ contains the rest. (just try to switch one of values and see it)

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <znc/znc.h>
+
+class UserTest : public ::testing::Test {
+  protected:
+    UserTest() { CZNC::CreateInstance(); }
+    ~UserTest() { CZNC::DestroyInstance(); }
+};
+
+TEST_F(UserTest, IsHostAllowed) {
+    CUser* m_pTestUser;
```
DarthGandalf: It's not a member, no `m_`

test/UserTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <gtest/gtest.h>
+#include <znc/User.h>
+#include <znc/znc.h>
+
+class UserTest : public ::testing::Test {
+  protected:
+    UserTest() { CZNC::CreateInstance(); }
+    ~UserTest() { CZNC::DestroyInstance(); }
+};
```
DarthGandalf: Is CZNC even used in this test?

src/User.cpp
```diff
@@ -852,6 +852,99 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
...
+            memset(&in6aBitmask, 0, sizeof(in6aBitmask));
+            memset(&aiHints, 0, sizeof(addrinfo));
+
+            aiHints.ai_family = AF_UNSPEC;  // Allow any address family
+            aiHints.ai_socktype = 0;        // Any socket
+            aiHints.ai_flags = AI_NUMERICHOST;
+            aiHints.ai_protocol = 0;  // Any protocol
+            aiHints.ai_canonname = NULL;
+            aiHints.ai_addr = NULL;
+            aiHints.ai_next = NULL;
```
DarthGandalf: s/NULL/nullptr/

However, I don't see value in this; you already nulled it via memset. So only important fields (non-null) need to be set.

src/User.cpp
```diff
@@ -852,6 +852,99 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
     for (const CString& sHost : m_ssAllowedHosts) {
         if (sHostMask.WildCmp(sHost)) {
             return true;
+        } else {
+            // CIDR notation checker, e.g. "192.0.2.0/24" or "2001:db8::/32"
+            addrinfo *aiHost, *aiRange;
+            addrinfo aiHints;
+            in6_addr in6aBitmask;
```
DarthGandalf: Please declare variable closer to their usage

DarthGandalf:
> Oops. Fixed those memory leaks. I hadn't tried building with clang.

Yeah, AddressSanitizer is awesome

src/User.cpp
```diff
@@ -852,6 +852,99 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
...
+                }
+
+                // Compare masked IPv6s
+                bIsHostInRange = true;
+                const sockaddr_in6* sa6Host = (sockaddr_in6*)(aiHost->ai_addr);
+                const sockaddr_in6* sa6Range =
+                    (sockaddr_in6*)(aiRange->ai_addr);
+
+                for (int i = 0; i < 16; ++i) {
+                    if (!((in6aBitmask.s6_addr[i] &
```
DarthGandalf: `!(a == b)` is the same as `a != b`, but the latter reads a bit easier.

DinCahill:
Thanks for all the help! I think I've fixed everything you mentioned, except CZNC.

test/UserTest.cpp
DinCahill: If I remove the CZNC calls, it looks like the unit test segfaults when instantiating a CUser.

DarthGandalf: Oh! I see.
Please add a comment though.

src/User.cpp
```diff
@@ -852,6 +852,94 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
...
+
+            addrinfo* aiRange;
+            int iIsRangeValid = getaddrinfo(vsSplitCIDR.front().c_str(), NULL,
+                                            &aiHints, &aiRange);
+            if (iIsRangeValid != 0) {
+                freeaddrinfo(aiHost);
+                continue;
+            }
+
+            // "/0" allows all IPv[4|6] addresses
```
DarthGandalf: Does 0.0.0.0/0 allow ::1?

DarthGandalf:
> Thanks for all the help! 

No, thanks to you :-P For implementing this

src/User.cpp
DinCahill: 0.0.0.0/0 doesn't allow ::1 because they're different address families. ::0/0 matches ::1, and 0.0.0.0/0 matches 127.0.0.1. I will add test cases for these.

I decided not to match "0.0.0.0" with "127.0.0.1" or "::0" with "::1". Should I implement this matching?

src/User.cpp
```diff
@@ -852,6 +852,94 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
     for (const CString& sHost : m_ssAllowedHosts) {
         if (sHostMask.WildCmp(sHost)) {
             return true;
+        } else {
+            // CIDR notation checker, e.g. "192.0.2.0/24" or "2001:db8::/32"
+
+            // Try to split the string into an IP and routing prefix
+            VCString vsSplitCIDR;
+            const int iSplitCIDRCount = sHost.Split("/", vsSplitCIDR, false);
+            const int iRoutingPrefix = vsSplitCIDR.back().ToInt();
```
psychon: Don't you have to check `iSplitCIDRCount == 2` before calling `back()`? `back()` on an empty vector is unspecified behavior.

src/User.cpp
DarthGandalf: > 0.0.0.0/0 doesn't allow ::1 because they're different address families

Ah.

> I decided not to match "0.0.0.0" with "127.0.0.1" or "::0" with "::1". Should I implement this matching?

Does such match make sense?

src/User.cpp
```diff
@@ -852,6 +852,94 @@ bool CUser::IsHostAllowed(const CString& sHostMask) const {
...
+            // from the routing prefix, AND it to the host and range, and see if
+            // they match
+            bool bIsHostInRange = false;
+            if (aiHost->ai_family == AF_INET) {
+                const sockaddr_in* saHost = (sockaddr_in*)(aiHost->ai_addr);
+                const sockaddr_in* saRange = (sockaddr_in*)(aiRange->ai_addr);
+
+                // Make IPv4 bitmask
+                const in_addr_t inBitmask =
+                    htonl((~0u) << (32 - iRoutingPrefix));
```
psychon: Doesn't this need a check for `iRoutingPrefix == 0`? Shifting a 32 bit variable by 32 is undefined behavior, I think (but I'm not completely sure and might be wrong)

DinCahill: I think it's OK because it's unsigned (the u in `~0u`). The check for `iRoutingPrefix == 0` is on L887.

psychon: > Oversized Shift Amounts: Shifting a uint32_t by 32 or more bits is undefined. My guess is that this originated because the underlying shift operations on various CPUs do different things with this: [...]

http://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html

DinCahill: Ah, I had checks for that at some point, but they got lost :P

I've also noticed that `CString::ToInt()` returns 0 on failure, so a "/a" is treated like a "/0". If the int is 0, i'll check that the string is all 0s.

src/User.cpp
DinCahill: Indeed! Thanks.

src/User.cpp
DinCahill: It could be useful to allow only local connections, on either IPv4, IPv6, or both. If you think no one's weird enough to do that: I do!

DarthGandalf: I think it's fine to write explicitly 127.0.0.1 and ::1 if one wants to allow only local connections.

DinCahill: Alrighty!

DinCahill:
Updated.

DinCahill:
Updated.

DarthGandalf:
LGTM from me. @psychon ?

psychon:
Looks good to me, too.

PR #1222, opened 2016-01-02
Default SSL ciphers updated to new Mozilla recommendation

onovy:


src/Socket.cpp
```diff
@@ -30,7 +30,7 @@
 #ifdef HAVE_LIBSSL
 // Copypasted from
 // https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29
-// at 22 Dec 2014
+// at 2 Jan 2016
```
Mikaela: Could it be possible to change this to the ISO 8601 format (2016-01-02)? I know the previous one wasn't, but I think it would be positive change and comments would be welcome to it.

src/Socket.cpp
```diff
@@ -30,7 +30,7 @@
 #ifdef HAVE_LIBSSL
 // Copypasted from
 // https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29
```
Mikaela: Is there some reason why we couldn't go to https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility ? (Offtopic:) This was for both client & server, wasn't it?

I don't think anyone uses old OpenSSL/GnuTLS/whatever that wouldn't have those and IRCds which are actively developed has them and XChat support already dropped with SSLv3 support.

src/Socket.cpp
onovy: done

src/Socket.cpp
onovy: i think this change is for "server" only. so no znc<->ircd connections, only client<->znc connections.

but good question. modern compatibilty will break:
* Java J2ME clients (really? :)
* webadmin in "old" browsers

onovy: btw: just tried "modern compatibility", and my hexchat on mac didn't connect:

* Connection failed ((336032784) error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure)

Mikaela: Discussing at #IRCv3 it depends on IRCd configs etc. but we cannot be fully sure.

@DarthGandalf Could we put git users to test the modern compatibility and see if we have any issues and if not, continue with the modern compatibility?

Mikaela: Apparently keeping this page open without refreshing didn't show your comments, thanks for testing, breaking HexChat is undesirable AFAIK.

teward: Let's just stick with Intermediate for out of the box compatibility - if someone wants to apply Modern they only have to add it to their configuration... out of the box I don't think Intermediate is bad, provided SSLv3 remains disabled in the protocols supported.

onovy: +1 teward

src/Socket.cpp
DarthGandalf: Does it matter here?... It's not for machine, but for human anyway. And by spelling month with letters there's no ambuiguity of order of day and month

Mikaela: There is ambiguity when you don't remember names and orders of months in English.

src/Socket.cpp
Mikaela: Apparently it's fault of OS X apps building with ancient OS X-shipped OpenSSL versions causing trouble.

PR #1228, opened 2016-01-05
Fixed few misspellings

onovy:


Mikaela:
The changelog misspellings should probably also be converted at the wiki.

* http://wiki.znc.in/Category:ChangeLog

PR #1237, opened 2016-01-24
WIP: Add support for i18n

DarthGandalf:


psychon:
Interesting stuff, nice!

Mikaela:
Is how this works from translator point of view documented anywhere yet?

DarthGandalf:
Nope, they can use usual tools to edit .po files
Lots of strings are still not translateable.

2016-03-08 8:39 GMT+00:00 Mikaela Suomalainen <notifications@github.com>:

PR #1245, opened 2016-02-26
Patch set from Gentoo

Whissi:
This is the updated patch set we are using in Gentoo. I rebased the patches against current znc git master.

While I have noticed that you have recently added support for ENV var _ZNC_NO_LAUNCH_AFTER_MAKECONF_ (https://github.com/znc/znc/pull/257) I'd like to take a different road:

Yes, wen can set an environment variable in our _pkg_config_ phase (`emerge --config znc`) but isn't it cleaner to have a parameter which is also documented in help?

If you don't like a hardcoded pidfile I can provide a different path using configure.

### DIFF
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -629,6 +629,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
     VCString vsLines;
 
     vsLines.push_back(MakeConfigHeader());
+    vsLines.push_back("PidFile = /run/znc/znc.pid");
     vsLines.push_back("Version = " + CString(VERSION_STR));
 
     m_sConfigFile = ExpandConfigPath(sConfigFile);
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -75,6 +75,7 @@ CZNC::CZNC()
       m_uiForceEncoding(0),
       m_sConnectThrottle(),
       m_bProtectWebSessions(true),
+      m_bSystemWideConfig(false),
       m_bHideVersion(false),
       m_Translation("znc") {
     if (!InitCsocket()) {
@@ -969,7 +970,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
         // installed.
         // See https://github.com/znc/znc/pull/257
         char* szNoLaunch = getenv("ZNC_NO_LAUNCH_AFTER_MAKECONF");
-        if (szNoLaunch && *szNoLaunch == '1') {
+        if (szNoLaunch && *szNoLaunch == '1' || m_bSystemWideConfig) {
             bWantLaunch = false;
         }
     }
@@ -2121,3 +2122,7 @@ void CZNC::LeakConnectQueueTimer(CConnectQueueTimer* pTimer) {
 }
 
 bool CZNC::WaitForChildLock() { return m_pLockFile && m_pLockFile->ExLock(); }
+
+void CZNC::SetSystemWideConfig(bool systemWideConfig) {
+    m_bSystemWideConfig = systemWideConfig;
+}
--- a/include/znc/znc.h
+++ b/include/znc/znc.h
@@ -244,6 +244,8 @@ class CZNC {
 
     static void DumpConfig(const CConfig* Config);
 
+    void SetSystemWideConfig(bool systemWideConfig);
+
   private:
     CFile* InitPidFile();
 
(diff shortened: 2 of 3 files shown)


src/znc.cpp
```diff
@@ -628,6 +629,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
     VCString vsLines;
 
     vsLines.push_back(MakeConfigHeader());
+    vsLines.push_back("PidFile = /run/znc/znc.pid");
```
teward: Isn't the standard to use `/run/process.pid`?  Why do we have to place it in a sub folder?  We may also wish to make this a config level item which can pull from the conf file if someone needs a customized setup.

Whissi: We are using a sub folder because per default we are creating a dedicated *znc* user to run znc. But normal user cannot write to */run/* per default. That's why we need a sub folder.

psychon: So this works for Gentoo, but breaks if I want to have my own ZNC installation in `~/znc-install`? That sounds bad.

Whissi: Well, remember: This only sets the default value. In your config you would change that.

However that's why I added the comment in the PR that I can provide a mechanism to specify default pidfile via configure.

teward: I don't have a real say, but I'm going to -1 your entire pull req.  The default pidfile should be able to be configured *without* having to do `./configure` and then compiling to do it, if we want a customized PIDFile location, it would be prudent to put it into a config-level-defined option, rather than as a compile option (precedent: Apache, nginx, lighttpd, many other services which permit pidfile paths to be defined at the configuration file level, rather than at the configure command).

As psychon identified, this will break as-is if we want an installation in `~/znc-install`.  If the intent is to run this as a system service, then this will work for that; but as there is zero guarantee the system will have ZNC running as a system service; with many people using user-space installations, I don't see this being a 'compatible' change being introduced.

We may also have cases where people take ZIP files from one system, and move them to another with different user/folder paths; we have to make sure that won't break, especially if we need users to recompile each and every time...

Whissi: While I do not really understand your concerns regarding picking up compiled znc and move to another system (these people shouldn't set the option per default so they don't end up with a default PidFile option) let me tell you why I initially added a default PidFile option:

If you are using a runscript/unit file you need a known resource you can check for status. I.e. if znc is running there must be a pidfile and this pid must be alive (just checking for znc processes isn't enough when you want to support multiple instances).

When user are installing znc, the configuration tool don't set such an option per default. So this is a problem. Telling the user "Now that you called the config tool you still need to add something manually" is not a good user experience, is it? That's why I added the patch.

We could only set PidFile option per default when `bSystemWideConfig` is true. However you would still disagree because this could be a problem if people decide to copy the compiled binary to another system where they want to use a different path.

Another solution would be to add a pidfile command line option. This way we could set the pidfile via runscript to a known value and wouldn't care if the config option is set and matches (cli parameter overrides).

What do you think?

DarthGandalf:
Are you picking up the package from @wired? If the package is going to be up to date again, cool.

DarthGandalf:
> Yes, wen can set an environment variable in our pkg_config phase (emerge --config znc) but isn't it cleaner to have a parameter which is also documented in help?

I don't see how this feature can be useful outside of Gentoo...
Usual users don't need to create config and actually start it separately from each other.

src/znc.cpp
DarthGandalf: There is another use case, even on Gentoo: to not use the provided init scripts, but just run `znc --makeconf` as your own user.

> If you are using a runscript/unit file you need a known resource you can check for status. I.e. if znc is running there must be a pidfile and this pid must be alive (just checking for znc processes isn't enough when you want to support multiple instances).

With `start-stop-daemon --background --make-pidfile` you can run `znc --foreground`, and then start-stop-daemon will put ZNC to background itself, and write its pidfile.

sbraz:
I will be co-maintaining the package with @wired as he seems to be AWOL.
With the environment variable, we should be able to drop one of our patches but the PID one will remain.
It's probably best we rely on znc's ability to daemonise instead of using the `--foreground` option and `--make-pidfile`.
I think the best way to configure the PID file would be via a command-line option. This way, we could have our init script specify it. This is something several applications do and it sounds like a sensible choice.

DarthGandalf:
> It's probably best we rely on znc's ability to daemonise instead of using the --foreground option and --make-pidfile.

May you explain why? Start-stop-daemon is used in much more places than ZNC, and should be reliable enough...

sbraz:
@DarthGandalf I'm not really sure what all the downsides are but I think start-stop-daemon wasn't initially made to start programmes in the background, most daemonise on their own.

There is one major drawback to the s-s-d approach: when you rely on it to save logs, you can't rotate without a service restart. However, I don't think this applies to znc as it doesn't write any system logs.

DarthGandalf:
Superceded by https://github.com/gentoo/gentoo/pull/3236

PR #1250, opened 2016-03-08
Add config write delay configuration setting

Adam-:
This prevents chansaver from causing configuration file writes so frequently.

### DIFF
--- a/include/znc/znc.h
+++ b/include/znc/znc.h
@@ -31,6 +31,7 @@ class CListener;
 class CUser;
 class CIRCNetwork;
 class CConnectQueueTimer;
+class CConfigWriteTimer;
 class CConfig;
 class CFile;
 
@@ -47,6 +48,7 @@ class CZNC {
         ECONFIG_NEED_REHASH,
         ECONFIG_NEED_WRITE,
         ECONFIG_NEED_VERBOSE_WRITE,
+        ECONFIG_DELAYED_WRITE,
         ECONFIG_NEED_QUIT,  // Not really config...
     };
 
@@ -125,6 +127,7 @@ class CZNC {
     void SetSSLCiphers(const CString& sCiphers) { m_sSSLCiphers = sCiphers; }
     bool SetSSLProtocols(const CString& sProtocols);
     void SetSSLCertFile(const CString& sFile) { m_sSSLCertFile = sFile; }
+    void SetConfigWriteDelay(unsigned int i) { m_uiConfigWriteDelay = i; }
     // !Setters
 
     // Getters
@@ -170,6 +173,7 @@ class CZNC {
     }
     CString GetSSLCertFile() const { return m_sSSLCertFile; }
     static VCString GetAvailableSSLProtocols();
+    unsigned int GetConfigWriteDelay() const { return m_uiConfigWriteDelay; }
     // !Getters
 
     // Static allocator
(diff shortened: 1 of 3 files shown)


psychon:
"want write" and "need write" sound quite similar and the difference is not at all clear from the name.

Also, why? What bad effects does writing the config have that mean it should only be done sometimes?

@DarthGandalf Could it be that the new shell module test fails randomly? Any idea why and what can be done about it?

Adam-:
Currently the config is written with chansaver anytime a user joins or parts a channel, and if you have a large config with a lot of users it can amount to a lot of disk i/o.I was trying to valgrind-debug znc some earlier and ended up with the application spending the majority of its time writing config files to disk due to this, which was frustrating.

DarthGandalf:
I don't know why that test fails :(
Probably the best action would be to just revert it...

src/znc.cpp
```diff
@@ -2121,3 +2152,10 @@ void CZNC::LeakConnectQueueTimer(CConnectQueueTimer* pTimer) {
 }
 
 bool CZNC::WaitForChildLock() { return m_pLockFile && m_pLockFile->ExLock(); }
+
+void CZNC::DisableConfigTimer() {
+  if (m_pConfigTimer) {
+    m_pConfigTimer->Stop();
+    m_pConfigTimer = nullptr;
```
DarthGandalf: No need to recreate the timer every time... `StartMaxCycles(delay, 1)`
Also, manager still has pointer to this timer, which never gets deleted until program shutdown. So number of these times will increase over time.

src/znc.cpp
```diff
@@ -227,6 +244,17 @@ void CZNC::Loop() {
...
+                if (GetConfigWriteDelay() > 0) {
+                  if (m_pConfigTimer == nullptr) {
+                    m_pConfigTimer = new CConfigWriteTimer(GetConfigWriteDelay());
+                    GetManager().AddCron(m_pConfigTimer);
+                  }
+                  break;
+                }
+		/* Fall through */
             case ECONFIG_NEED_WRITE:
             case ECONFIG_NEED_VERBOSE_WRITE:
```
DarthGandalf: I think, need to stop timer here, if there was request to write config without delay. Otherwise, after the timer runs, config will be written again.

include/znc/znc.h
```diff
@@ -47,6 +48,7 @@ class CZNC {
         ECONFIG_NEED_REHASH,
         ECONFIG_NEED_WRITE,
         ECONFIG_NEED_VERBOSE_WRITE,
+        ECONFIG_WANT_WRITE,
```
DarthGandalf: Something like `..._DELAYED_WRITE` probably would be better name

DarthGandalf:
Please reformat number of spaces to look more similar to code around

DarthGandalf:
@psychon I've enabled github's new feature of protecting branches to protect branches from me pushing broken code to them...

src/znc.cpp
Adam-: "manager"? CSockCommon::Cron automatically deletes timers when it detects them as being invalid.

DarthGandalf: I've had a closer look at Csocket, it looks you're right... `Pause()` could also help, but not very well in this case either. While I don't like this approach too much, it's better than some alternatives.

But please reformat the code. ZNC doesn't use 2 spaces for indentation.

DarthGandalf:
> I've enabled github's new feature of protecting branches to protect branches from me pushing broken code to them...

It's too strict though ("This branch is out-of-date with the base branch"), as rebasing every PR every time a commit is added to master is probably not an option. Disabling again.

PR #1260, opened 2016-04-11
Strip background colours when we reset foreground 

kylef:
The IRC message `\x03,03test` represents resetting foreground colour, following by setting a background colour. This currently wasn't correctly stripped in ZNC's StripControls.


kylef:
Sorry, I realised I made my branch on ZNC and recall the policy here was to use our own forks. Moved to https://github.com/znc/znc/pull/1261.

PR #1261, opened 2016-04-11
Strip background colours when we reset foreground

kylef:
The IRC message `\x03,03test` represents resetting foreground colour, following by setting a background colour. This currently wasn't correctly stripped in ZNC's StripControls.

I've also added test coverage for the rest of StripControls.


psychon:
:+1:

Issue #866, opened 2015-02-17
Add per-network option to disable TLS certificate validation


dgw:
Validating TLS certificates is great, but let's face it: A lot of networks use self-signed or otherwise "invalid" certificates, and for networks with many servers it's probably not worth maintaining the list of fingerprints in the majority of cases. Lots of networks that do provide their certificate fingerprints do so as SHA-1, not SHA-256, which further adds to the inconvenience.

For most of us, chances are it's not *that* critical to validate the certificates. Therefore, I propose that each network allow disabling certificate checking, so users can focus on chatting if they want.

dgw:
Note that this could be implemented as a special `*` value, rather than a separate option altogether, as was mentioned on IRC.

Mikaela:
This looks very insecure to me. Even if you add the first certificate blindly, with *luck* you will get to the server directly and be protected from MITM.

KindOne-:
Why not something like  ```/msg *status AddTrustedServerFingerprint __ACCEPT_INVALID_CERTIFICATES__``` and just let znc accept them without having to use a script or whatever to automatically add all the invalid ones.

rewbycraft:
Not sure if it's relevant, but I created a patch that adds a network option to just disable the SSL/TLS cerificate check on a per-network basis for a friend's BNC. For some networks it is impractical to manually trust the cerificates. I know it's not the most secure thing (I wouldn't use it), but most people aren't techsavvy enough to wanna deal with this. (Here's a link: https://github.com/YourBNC/znc/commit/5c74759840e68ef6262aef61dc4f73730f32d74f)

DarthGandalf:
And by "dealing with this" you mean copypasting one line from ZNC's output
in *status about bad certificate back to *status?

2015-02-18 11:12 GMT+00:00 Roelf Wichertjes <notifications@github.com>:

rewbycraft:
A friend requested I write that patch for him. Those were his exact words. Someone pointed me to this thread and said that the patch may be relevant.

I want this to be clear: Personally, I agree with you 100%.

But he (correctly) pointed out that most people aren't tech-savvy enough to know what to do (they usually stop reading when the numbers start popping up).
I'm not asking the patch be included in znc main, just noting the patch for people who are interested.

DarthGandalf:
Are they tech-savvy enough to be able to select the relevant setting?

2015-02-18 11:22 GMT+00:00 Roelf Wichertjes <notifications@github.com>:

[...]

PR #1277, opened 2016-05-19
Add network-specific settings for cert validation

xnrand:
Added the following two network-specific configuration options that can
be changed via controlpanel or webadmin:

* TrustAllCerts: Will trust ALL certificates when enabled, effectively
  disabling TLS certificate validation. This has precedence over TrustPKI.  
  Default value: false
* TrustPKI: Whether or not to trust PKI-valid certificates. Setting this
  to false will make znc trust only trusted certificates added by the
  user.  
  Default value: true

With default values, behavior is exactly the same as before.

This is based on the work of Roelf Wichertjes. See YourBNC/znc@5c747598.

See znc/znc#866.

----

This would resolve znc/znc#866.

See [my comment there](https://github.com/znc/znc/issues/866#issuecomment-220449552) for my reasoning why I implemented it this way.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,12 @@
     * it is not called during ZNC startup any more
 * Fix build on Solaris 10
 * Fix build with LibreSSL
+* Added two network-specific config options to control TLS certificate
+  validation
+    * `TrustAllCerts`: Disables certificate validation. Will take precedence
+      over TrustPKI
+    * `TrustPKI`: Whether or not to trust PKI-valid Certificates. Setting this
+      to false will make znc trust only trusted certificates added by the user.
 
 
 
--- a/include/znc/IRCNetwork.h
+++ b/include/znc/IRCNetwork.h
@@ -266,6 +266,12 @@ class CIRCNetwork {
         m_uJoinDelay = uJoinDelay;
     }
 
+    void SetTrustAllCerts(const bool bTrustAll = false) { m_bTrustAllCerts = bTrustAll; }
+    bool GetTrustAllCerts() const { return m_bTrustAllCerts; }
+
+    void SetTrustPKI(const bool bTrustPKI = true) { m_bTrustPKI = bTrustPKI; }
+    bool GetTrustPKI() const { return m_bTrustPKI; }
+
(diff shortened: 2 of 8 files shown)


xnrand:
## Security implications

There are IRC networks with TLS Certificates that are not signed by a CA that's publicly trusted. These were the motivation for this change. This change allows the user to adjust the level of scrutiny that a TLS Certificate is validated with. This comment outlines the security implications of this change by going over the possible values of the introduced settings.

### TrustAllCerts=false, TrustPKI=true (Default values)  
  With this setting, behavior is the same as _status quo_, certificates are regarded as valid if they are signed by a publicly trusted Certificate Authority or if a used has added them to the list of trusted fingerprints.
### TrustAllCerts=true, TrustPKI doesn't matter  
  With this setting, certificates are not validated and all certificates are trusted. This might leave user _vulnerable_ to MITM attacks.  
  However, it is considerably better than using a plain connection without any encryption. While MITM attacks are possible, these require sophisticated means to realize. This configuration value still protects against passive sniffing of the network traffic.  
  Not providing this configuration value will cause some users that connect to IRC Networks with invalid certificates to just use a plain connection, because doing so is easier than adding the fingerprints of TLS Certificates to the list of trusted fingerprints.  
  This value provides a more secure but equally easy alternative to using a plain connection. This is why I believe providing this configuration value to the user is important.
### TrustAllCerts=false, TrustPKI=false
  With this setting, certificates are trusted only if they are on the list of trusted fingerprints. This can be considerably _more_ secure than the status quo, because not as many parties need to be trusted:  
  The Public Key Infrastructure is huge. There are [more than 600 publicly trusted CAs from all over the world](https://www.eff.org/observatory) holding [over 1000 CA certificates](https://www.eff.org/files/defconssliverse.pdf#19). By trusting the PKI, you trust _all_ of these CAs. The interesting point about this is that the trustworthiness of the CA you choose for your service does not matter. What matters is the trustworthiness of the least trustworthy CA out of that set, because if any one CA signs a certificate that should not be trusted, all clients that rely on the PKI will trust that certificate.  
  For that reason, some might choose not to trust the PKI and instead manually verify their certificates. This is why I believe providing this configuration value to the user is important.

PR #1279, opened 2016-05-22
Added the reload action to the systemd service

jgeboski:


DarthGandalf:
This feature isn't really supported, and is broken. Adding more ways to trigger it won't help.

PR #1285, opened 2016-06-29
adds option to join after identification to nickserv module

jkhsjdhjs:
Adds an option to join after successful nickserv identification or immediately like how it is currently.
Useful if you want to have your vhost activated before joining channels and SASL or CertFP is not available.
Can be configured via query with *nickserv:
SetJoinAfterIdentified true/false

### DIFF
--- a/modules/nickserv.cpp
+++ b/modules/nickserv.cpp
@@ -113,7 +113,7 @@ class CNickServ : public CModule {
         }
 
         CString sTmp;
-		m_bJoinAfterIdentified = (sTmp = GetNV("JoinAfterIdentified")).empty() ? true : sTmp.ToBool();
+        m_bJoinAfterIdentified = (sTmp = GetNV("JoinAfterIdentified")).empty() ? true : sTmp.ToBool();
 
         return true;
     }
@@ -148,10 +148,10 @@ class CNickServ : public CModule {
     }
 
     EModRet OnJoining(CChan& Channel) override {
-		if (!m_bIdentified && m_bJoinAfterIdentified)
-			return HALT;
-		return CONTINUE;
-	}
+        if (!m_bIdentified && m_bJoinAfterIdentified)
+            return HALT;
+        return CONTINUE;
+    }
 
     EModRet OnPrivMsg(CNick& Nick, CString& sMessage) override {
         HandleMessage(Nick, sMessage);
@@ -165,7 +165,7 @@ class CNickServ : public CModule {
 
     void OnIRCDisconnected() override {
         m_bIdentified = false;
-	}
+    }
   private:
     bool m_bIdentified{};
     bool m_bJoinAfterIdentified{};
--- a/modules/nickserv.cpp
+++ b/modules/nickserv.cpp
@@ -16,6 +16,7 @@
 
 #include <znc/Modules.h>
 #include <znc/User.h>
+#include <znc/IRCNetwork.h>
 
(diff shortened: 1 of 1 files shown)


Zarthus:
Your code seems to suffer from mixed indentation (tabs/spaces used inconsistently.)

I'd use whatever the module currently is using, and maybe event convert it to one or the other depending on if it has mixed content itself too.

jkhsjdhjs:
@Zarthus 
yes, didn't notice that, thanks!
just fixed it

modules/nickserv.cpp
```diff
@@ -119,9 +141,18 @@ class CNickServ : public CModule {
             MCString msValues;
             msValues["password"] = GetNV("Password");
             PutIRC(CString::NamedFormat(GetNV("IdentifyCmd"), msValues));
+        } else if(Nick.NickEquals(sNickServName) && sMessage.find("Password accepted") != CString::npos && m_bJoinAfterIdentified) {
```
DarthGandalf: This will break on every services which show some different string instead of this one

modules/nickserv.cpp
```diff
@@ -93,6 +112,9 @@ class CNickServ : public CModule {
             SetNV("IdentifyCmd", "NICKSERV IDENTIFY {password}");
         }
 
+        CString sTmp;
+        m_bJoinAfterIdentified = (sTmp = GetNV("JoinAfterIdentified")).empty() ? true : sTmp.ToBool();
```
DarthGandalf: I don't think default should be true...
Nickserv module is too fragile, and if user just stops joining any channel, how is he supposed to know what happened?

modules/nickserv.cpp
```diff
@@ -56,6 +57,20 @@ class CNickServ : public CModule {
         PutModule("Ok");
     }
 
+    void SetJoinAfterIdentifiedCommand(const CString& sLine) {
+        if(sLine.Token(1, true).find("true") != CString::npos) {
```
DarthGandalf: Just use `ToBool()`?

modules/nickserv.cpp
jkhsjdhjs: @DarthGandalf 
yes, you're right. but is there any other way to do it?
one way would be that the user has to set keywords for example like "Password accepted", but for each network specifically
or is there a better alternative?

DarthGandalf: SASL is the alternative.
For this module, I suggest checking how it looks like on various networks which use various services packages, but be prepared for breakage anyway.

dgw: At least, be prepared for issue reports that will require adding a large list of possible strings from various networks, as seen in this module's `HandleMessage` function.

jkhsjdhjs: Ok, i took a look around and saw that every irc network I visited which uses nickserv has "Password accepted" in their message and it's also default in unrealircd
But you're right, there are for sure networks that don't do that so I'm prepared if someone has problems with this option.

kylef: Atheme uses "You are now identified" (https://github.com/atheme/atheme/blob/4fd9583b7b8498fb5379b2c07ec4de50f877d85e/modules/operserv/identify.c#L85)

jkhsjdhjs: Sorry, I didn't notice your comment till now. I added support for atheme, but atheme and anope are the only IRC services I know of. Are there any other I should support?

PR #1286, opened 2016-06-29
Add Dockerfile

torarnv:


torarnv:
Well, @coveralls seems to be on crack 😄

Zarthus:
Is this somehow configurable to permit changing things and still have git support? I'd maybe rename it to `Dockerfile.example` and let the user manually `mv` it here, adding it to gitignore.

I personally don't care about python and perl support, my addons are written in C++. 

I also think this adds too much things to the README, perhaps it is better suited as a page on the Wiki. (you may even just place the full dockerfile in the wiki and not have it in the root repository. But I personally don't mind.)

torarnv:
> Is this somehow configurable to permit changing things and still have git support?

Depends what you mean? You can change code and rebuild znc from git, effectively using docker as an isolated build environment. That's what the last part of the README shows an example of.

> I personally don't care about python and perl support, my addons are written in C++.

You can pass `config_flags=` to your liking. The default is there for people who don't know they need modpython/perl until they want to run a module that requires it, and then are out of luck. Many of the prebuilt znc packages I've tried don't include modpython/perl, which leaves the user to build znc from source. A prebuilt docker image with those options enabled simplifies a lot of things.

> I also think this adds too much things to the README

I agree, the wiki is a good alternative, I'll move the instructions there.

> I'd maybe rename it to Dockerfile.example

Part of the benefit of having it in the upstream repo is being a default, but I see your point. An extension of this would be having official docker images of znc, for the latest version, and master. Would simplify testing and deployment.

Zarthus:
We talked over IRC and smoothened some things out: Essentially my suggestion for disabling python and perl support were already implemented, just unclearly phrased in the README additions. It appears to be configurable after all.

torarnv:
Instructions have now moved to the wiki 👍

AKPWebDesign:
I've been trying this for a couple of hours now, and simply cannot get it to work. Whether I'm trying to migrate my existing ZNC over to a Docker container or just trying to generate a fresh config, I can't get a ZNC instance running.

Main problem: file permissions. Setting up a linked volume as the instructions specify leaves me getting "Can not open config file" errors when providing an already existing config, and file save errors when saving a new config file.

I solved this by chmodding my entire ZNC data directory to 777, but then ZNC simply isn't connecting to anything once it does start running.

DarthGandalf:
If I understand it correctly, with this file in repo we can provide an "approved" Docker image with ZNC?

torarnv:
That's the idea yes. 

I'm back from vacation now and will try to find some cycles to have a look at the file permission issue.

DarthGandalf:
Looking at https://hub.docker.com/_/irssi/, they download the latest tarball, instead of relying on . to contain sources of the right version. Maybe we should do the same? What are pros/cons?

I had to learn basics of docker when saw this pull request, so not very familiar with it yet.

DarthGandalf:
Another question: is https://github.com/znc/znc the right place for the Dockerfile? Would https://github.com/znc/znc-docker be more appropriate? Similar to existing https://github.com/znc/znc-cygwin

pixie79:
Hi,

This is a great start but can i make a couple of suggestions.

1, I would add and apk update first to ensure all the packages installed are the latest as this saves from creating containers with known bugs which are already fixed.

2, I would not use "/home/znc" as the install path - generally it would be better to either follow the unix standard and install under "/opt/znc" (then localise bin and everything else to their as well) as it is not a system package or follow the common docker method of installing under "/app" which is probably the cleanest option.

Regards

Mark

jyaworski:
I second @pixie79's suggestions about updating the apks first and using /app. Also, we should consider alpine:latest instead of pinning on 3.3. They're remarkably good at compatibility, since it's so minimal.

@DarthGandalf lots of projects treat Dockerfiles as a top-level file in the main project, since it's just one file. It's like requirements.txt in Python, or Gemfile in Ruby.

sixcorners:
@pixie79 
I don't think doing additional things to update stuff is necessary.
https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache
> As of Alpine Linux 3.3 there exists a new --no-cache option for apk. It allows users to install packages with an index that is updated and used on-the-fly and not cached locally

Zoddo:
@torarnv Any update? I would like to see this PR merged :)

xavier83:
Any luck here?

nicolai86:
Just tried to set up znc with docker using this PR plus the suggested changes from @pixie79  (`/home/znc` -> `/opt/znc`), and it works without any issues!

DarthGandalf:
I submitted this to https://github.com/znc/znc-docker with some changes, which are mostly inspired by @jimeh
Does anyone wish to review it?

Zarthus:
- port 6667 is used by ircds, I'd change it to something else
- I'd say a git-clone is preferable over http download via znc wiki and verifying the signature.

DarthGandalf:
> I'd say a git-clone is preferable over http download via znc wiki and verifying the signature.

Why, and what do you mean by wiki? Wiki doesn't store tarballs.

> you might want to upload it to docker hub at some point

Yeh, I intended to send pull request to https://github.com/docker-library/official-images after initial sanity check.

Zarthus:
https://github.com/znc/znc-docker/blob/master/small/Dockerfile#L31-L32

```
    && wget "http://znc.in/releases/archive/znc-${ZNC_VERSION}.tar.gz" \
    && wget "http://znc.in/releases/archive/znc-${ZNC_VERSION}.tar.gz.sig" \
```

I referred to it as the wiki because it was on the same site, but yes, I'm mistaken on that front. The wiki just links to it. It's still HTTP traffic and I haven't heard any progression of them wishing to move to https any time soon.

DarthGandalf:
https://github.com/docker-library/official-images#security says that PGP check is better than HTTPS.

Zarthus:
you can do both, but clone over https.

When you download the signature from the same source as the tarball, it probably loses a significant amount of meaning.

DarthGandalf:
"Releases" on github are different tarballs (e.g. they don't have pregenerated SWIG files or configure script), and they are not signed, are they?

> When you download the signature from the same source as the tarball, it probably loses a significant amount of meaning.

No, it doesn't. It still needs to match D5823CACB477191CAC0075555AE420CC0209989E hardcoded to Dockerfile.
It's different from downloading checksum from the same place as tarball. That would lose its meaning over HTTP.

Zarthus:
Where is the `D5823CACB477191CAC0075555AE420CC0209989E` hardcoded? I might be missing something obvious.

Zarthus:
Yep - I'm mistaken, sorry. I completely glossed over the --recv-keys part - I was eyeballing for a statement like `ENV RUBY_DOWNLOAD_SHA` as per the link you provided.

DarthGandalf:
https://hub.docker.com/_/znc/ is there now.
Thanks!

PR #1289, opened 2016-07-05
Don't strip timestamp when sending echo-messages

torarnv:
We preserve the time of the original message by making a copy of it,
and just changing the nick, which we then pass on to the CMessage
overload of PutClient, which converts the timestamp into a @time tag.


DarthGandalf:
A test for echo-message would help here...

PR #1290, opened 2016-07-05
Fix inverted gettimeofday() return value handling

torarnv:
The gettimeofday function returns 0 for success, not for failure. As a
result of the inverted logic we were losing millisecond precision when
parsing incoming messages on non-HAVE_CLOCK_GETTIME systems (macOS).

A followup commit does a bit of cleanup to prevent this kind of error in the future, by centralising the way we get the current time.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -418,6 +418,31 @@ inline CString FixGMT(CString sTZ) {
 }
 }  // namespace
 
+timeval CUtils::GetTime() {
+#ifdef HAVE_CLOCK_GETTIME
+    timespec ts;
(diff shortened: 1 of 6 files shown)


src/Utils.cpp
```diff
@@ -418,6 +418,33 @@ inline CString FixGMT(CString sTZ) {
 }
 }  // namespace
 
+timeval CUtils::GetTime()
+{
```
DarthGandalf: Please put { on the previous line (or clang-format can do that for you :) )

torarnv: 👍  Done

PR #1294, opened 2016-07-13
Add OnPrivBufferStarting/Ending signals, similar to Chan buffers

torarnv:
Allows hooking into query buffer playback before and after all the
lines of the query buffer are replayed. The EModRet return value
has no effect at the moment, but may be used in the future to e.g.
prevent playback. The Chan version of these signals use EModRet to
skip emitting the status message, but not the whole playback.

### DIFF
--- a/include/znc/Modules.h
+++ b/include/znc/Modules.h
@@ -32,6 +32,7 @@
 // Forward Declarations
 class CAuthBase;
 class CChan;
+class CQuery;
 class CIRCNetwork;
 class CClient;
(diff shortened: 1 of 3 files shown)


torarnv:
CI build failures seem unrelated.

psychon:
I restarted the failed travis build. Perhaps "leader" will this time wait successfully for his "minions", whatever that means...

torarnv:
Kicking Travis seems to have fixed it 😄

Issue #1180, opened 2015-11-08
Override CSRF protection for POST requests [$73]


lol768:
Is it possible to disable/override the CSRF protection token verification requirement for a module's OnWebRequest listener?

I believe the check is performed [here](https://github.com/znc/znc/blob/d66cb36cc3cef65f472f4c6535128b9f785c94b4/src/WebModules.cpp#L608) - there doesn't appear to be a way to disable it.

psychon:
Am I understanding you correctly: You are writing a module `foo` and it offers a web page `/bar` and you want to disable the CSRF protection for this specific page?
Why would you want/need that?

lol768:
Hi @psychon,

That's correct. In my case, it's not a webpage but it's a route in a RESTful API I'm working on. For one of the routes, a POST action makes the most sense and there's no real way that the API client would be able to know the CSRF token (nor should it be expected to).

Cheers!

DarthGandalf:
Is HTTP-Basic auth an option for you?

lol768:
Not in this case, the API is designed to be accessible to anyone (no authentication should be required).

psychon:
So, are you looking for `WebRequiresLogin() { return false; }`?
https://github.com/znc/znc/blob/73d5f382f83eb1cd77d3ea4afb98e132fb6f4705/include/znc/Modules.h#L422-L425

lol768:
> Edit: Ohhhhh, now I get it!
> You already have WebRequireLogin() returning false, but of course the CSRF check is still run.

Yup. I've worked around it so far by forcing use of GET requests but it's not ideal, especially when you have a lot of data (gets messy putting it in the query string).

lol768:
I've added a bounty on this since it was annoying me and I'd like to do some further API work in my module.

Ideally I think there'd be a `virtual EModRet OnCsrfCheck(CWebSock& WebSock, const CString& sPageName);` (and then I could return HALT) or perhaps a  `virtual bool RequiresCsrfCheck(CWebSock& WebSock, const CString& sPageName);` where I could just return `false` for API requests.

PR #1296, opened 2016-07-15
Allow modules to override CSRF protection.

tzvetkoff:
Allow modules to override CSRF protection.
Useful for Web APIs and all other kinds of things.

API changes:
	- Added public CHTTPSock::GetURI() method
	- Added public CModule::ValidateWebRequestCSRFCheck() method
	- Made CWebSock::GetCSRFCheck() method public so it can be accessed
	  from CModule
	- Added public CWebSock::ValidateCSRFCheck() method

Other changes:
	- Added a Sample Web API module (modules/samplewebapi.cpp) and a
	  simple web form with no CSRF check.

Implements feature request #1180.

### DIFF
--- a/src/WebModules.cpp
+++ b/src/WebModules.cpp
@@ -655,8 +655,8 @@ CWebSock::EPageReqResult CWebSock::OnPageRequestInternal(const CString& sURI,
     // 1. they obviously know the password,
     // 2. it's easier to automate some tasks e.g. user creation, without need to
     // care about cookies and csrf
-    if (IsPost() && !m_bBasicAuth &&
-        GetParam("_CSRF_Check") != GetCSRFCheck() && sURI != "/login") {
+    if (IsPost() && !m_bBasicAuth && !sURI.StartsWith("/mods/") &&
+        !ValidateCSRFCheck(sURI)) {
         DEBUG("Expected _CSRF_Check: " << GetCSRFCheck());
         DEBUG("Actual _CSRF_Check:   " << GetParam("_CSRF_Check"));
         PrintErrorPage(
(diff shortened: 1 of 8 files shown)


lol768:
Cheers @tzvetkoff

src/WebModules.cpp
```diff
@@ -655,8 +655,8 @@ CWebSock::EPageReqResult CWebSock::OnPageRequestInternal(const CString& sURI,
     // 1. they obviously know the password,
```
DarthGandalf: This comment becomes outdated.

DarthGandalf:
Please add some tests.

I don't quite understand why existing `WebRequiresLogin()` is not enough. Shouldn't that function be used to disable protection?

lol768:
> I don't quite understand why existing WebRequiresLogin() is not enough. Shouldn't that function be used to disable protection?

I can still think of use-cases where I want to require cookie-based auth but not require a CSRF token. For example, a JSON-based API used with frontend JavaScript. CSRF isn't necessarily a worry if a `Content-Type` of `application/x-www-form-urlencoded` isn't accepted since a form is only going to send data using that content type - and without CORS, the same-origin policy will prevent other-origin requests to the same API.

In this circumstance, requiring a CSRF value to be submitted is just additional hassle.

DarthGandalf:
Yes, that's a good example, thanks.

lol768:
@DarthGandalf @tzvetkoff I've updated this PR against the latest `master`, clarified some of the source code comments and added an integration test.

https://github.com/lol768/znc/commits/module_csrf_override

What's the best way forward, can you merge my commits in @tzvetkoff?

DarthGandalf:
@lol768 you can create PR yourself if @tzvetkoff doesn't merge your commits to this PR.

lol768:
Cheers @DarthGandalf, will do so if the commits aren't added in a few days.

lol768:
PR added to #1327

PR #1302, opened 2016-08-11
Add my own query messages to the buffer

kgraefe:
This patch adds my own query messages to the buffer when no client is
connected.

My setup is `IRC <--> irssi with irssi-proxy <--> ZNC <--> Pidgin` . That way I have a complete backlog in both clients. Whoever, when I have a query chat within `irssi` while `Pidgin` is not connected , I only get my partners messages in `Pidgin`. This patch fixes this.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -758,8 +758,16 @@ void CIRCSock::ReadLine(const CString& sData) {
 					if (OnPrivMsg(Nick, sMsg)) {
 						return;
 					}
-				} else {
+				} else if(m_pNetwork->IsChan(sTarget)) {
 					if (OnChanMsg(Nick, sTarget, sMsg)) {
+						return;
+					}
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Only clients which requested `znc.in/self-message` or `echo-message` should receive this.

kgraefe:
Do you see a clean way to implement that? I think there should be some flags in AddBuffer() but that would be a change I would feel comfortable with because I don't know the source well enough to estimate the impact.

Is filtering the replay done somewhere else in the source code?

kgraefe:
Do I have to worry about the Travis check failure?

RealKindOne:
No, Travis is failing due to other reasons. #1306

kgraefe:
According to the Change Log,  `self-message`  support has been there since ZNC 1.6.0. I'm not sure why I wasn't able to figure out how to use it back then. I can now, so this PR is (and probably has always been) obsolete. Closing.

Issue #1310, opened 2016-08-30
OpenSSL 1.1 build issues


welwood08:
I don't know if anyone is actively working on compatibility with the recently-released OpenSSL 1.1, I can't find anything public at least. Here follows some snippets from my own attempts to build znc against OpenSSL 1.1 so as to document many of the problems in one place and provide visibility of the issue for anybody else attempting the same thing.

```
$> /usr/local/bin/openssl version -a -o

OpenSSL 1.1.0  25 Aug 2016
built on: reproducible build, date unspecified
platform: linux-elf
options:  bn(64,32) rc4(1x,char) des(long) idea(int) blowfish(ptr)
compiler: /usr/local/bin/gcc62 -DDSO_DLFCN -DHAVE_DLFCN_H -DNDEBUG -DOPENSSL_THREADS -DOPENSSL_NO_STATIC_ENGINE -DOPENSSL_PIC -DOPENSSL_BN_ASM_PART_WORDS -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DMD5_ASM -DRMD160_ASM -DAES_ASM -DVPAES_ASM -DWHIRLPOOL_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DPOLY1305_ASM -D_FORTIFY_SOURCE=2 -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-1.1\""  -O3 -gdwarf-2 -gstrict-dwarf -fstack-protector-all -Wformat -Wformat-security -Werror=format-security -I/usr/local/lib/gcc/i686-pc-linux-gnu/6.2.0/include -I/usr/local/lib/gcc/i686-pc-linux-gnu/6.2.0/include-fixed -I/usr/local/include/c++ -Wa,--noexecstack
OPENSSLDIR: "/usr/local/ssl"
ENGINESDIR: "/usr/local/lib/engines-1.1"

[...]

PR #1311, opened 2016-09-03
OpenSSL 1.1 compatibility

welwood08:
As promised in #1310, this deals with making znc code compile against OpenSSL 1.1 with no errors or relevant deprecation warnings.

I have compiled these changes with OpenSSL versions 1.1.0, 1.0.2h and 0.9.8g on a Debian system - all successfully. I've also run `make test` (after fixing it for the 1.6.x branch, let me know if you're interested) on each version - pass every time, although I'm not sure how much that's really testing. I have not yet used the built znc (other than `znc -v`) or attempted to exercise any of the features these changes affect.

Please make sure I haven't done anything obviously stupid and then I'll happily "eat my own dog food" with OpenSSL 1.1 - hopefully others can test with older OpenSSLs. I'm not a user of the schat module, so also hope someone can test that with 1.1.

### DIFF
--- a/modules/schat.cpp
+++ b/modules/schat.cpp
@@ -25,6 +25,12 @@
 #include <znc/User.h>
 #include <znc/IRCNetwork.h>
 
+#if !defined(OPENSSL_VERSION_NUMBER) || defined(LIBRESSL_VERSION_NUMBER) || OPENSSL_VERSION_NUMBER < 0x10100007
+/* SSL_SESSION was made opaque in OpenSSL 1.1.0, cipher accessor was added 2 weeks before the public release.
+See openssl/openssl@e92813234318635639dba0168c7ef5568757449b. */
+# define SSL_SESSION_get0_cipher(pSession) ((pSession)->cipher)
+#endif
+
 using std::pair;
 using std::stringstream;
 using std::map;
@@ -226,9 +232,7 @@ public:
 					Table.SetCell("Host", pSock->GetRemoteIP());
 					Table.SetCell("Port", CString(pSock->GetRemotePort()));
 					SSL_SESSION *pSession = pSock->GetSSLSession();
-					if (pSession && pSession->cipher && pSession->cipher->name)
-						Table.SetCell("Cipher", pSession->cipher->name);
-
+					Table.SetCell("Cipher", SSL_CIPHER_get_name(pSession ? SSL_SESSION_get0_cipher(pSession) : NULL));
 				} else {
 					Table.SetCell("Status", "Waiting");
 					Table.SetCell("Port", CString(pSock->GetLocalPort()));
@@ -286,11 +290,7 @@ public:
 					Table.SetCell("RemoteIP:Port", pSock->GetRemoteIP() + ":" +
 							CString(pSock->GetRemotePort()));
 					SSL_SESSION *pSession = pSock->GetSSLSession();
-					if (pSession && pSession->cipher && pSession->cipher->name)
-						Table.SetCell("Cipher", pSession->cipher->name);
-					else
-						Table.SetCell("Cipher", "None");
-
+					Table.SetCell("Cipher", SSL_CIPHER_get_name(pSession ? SSL_SESSION_get0_cipher(pSession) : NULL));
 				} else {
 					Table.SetCell("Type", "Listener");
 					Table.SetCell("LocalIP:Port", pSock->GetLocalIP() +
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -26,6 +26,7 @@
 #include <znc/FileUtils.h>
 #ifdef HAVE_LIBSSL
 #include <openssl/ssl.h>
+#include <memory>
 #endif /* HAVE_LIBSSL */
 #include <unistd.h>
 #include <time.h>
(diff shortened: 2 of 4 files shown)


DarthGandalf:
No changes to Csocket required?

welwood08:
Csocket changes are in jimloco/Csocket#64.

welwood08:
Looking at the travis logs, it seems at least the osx run failed on `make test` due to the problem I encountered with the same download - shall I add the fix in here or another pull req or not bothered?

DarthGandalf:
Not here. It doesn't seem to be related to OpenSSL.

modules/schat.cpp
```diff
@@ -226,9 +230,15 @@ class CSChat : public CModule
...
+#ifdef HAVE_OPAQUE_SSL_SESSION
+					if (pSession)
+						Table.SetCell("Cipher", SSL_CIPHER_get_name(SSL_SESSION_get0_cipher(pSession)));
+#else
 					if (pSession && pSession->cipher && pSession->cipher->name)
 						Table.SetCell("Cipher", pSession->cipher->name);
-
+#endif
+					else
+						Table.SetCell("Cipher", "None");
```
psychon: How about defining your own (`static inline`) `SSL_CIPHER_get_name` if none is provided by OpenSSL? That way you don't need an `#ifdef'  in the code, but only in the "global" part.

src/main.cpp
```diff
@@ -17,7 +17,8 @@
 #include <znc/znc.h>
 #include <signal.h>
 
-#if defined(HAVE_LIBSSL) && defined(HAVE_PTHREAD)
+#if defined(HAVE_LIBSSL) && defined(HAVE_PTHREAD) && \
+( !defined(OPENSSL_VERSION_NUMBER) || OPENSSL_VERSION_NUMBER < 0x10100004 ) /* 1.1.0-pre4: openssl/openssl@2e52e7df518d80188c865ea3f7bb3526d14b0c08 */
```
psychon: Doesn't this break thread-safety?
https://www.openssl.org/docs/faq.html#PROG1

modules/schat.cpp
welwood08: After reading a bit about what that means, it does indeed sound like a better idea. I'll see what I can come up with, thanks.

src/main.cpp
welwood08: To quote the commit message that I pointed to in the comment about 1.1.0-pre4: "There is now no longer a need to set locking callbacks!!". The `CRYPTO_set_locking_callback` and `CRYPTO_THREADID_set...` APIs mentioned in the FAQ were all replaced with no-ops in that commit. Therefore, as far as I understand it, 1.1 now doesn't need any of this stuff and perhaps just the FAQ hasn't been updated after the recent release.

modules/schat.cpp
psychon: Alternatively, you could also do `#define SSL_CIPHER_get_name(session) ((session)->cipher) ? (session)->cipher->name : "None")` instead of a static inline method.

src/main.cpp
psychon: Ah, I understand.

I guess that should be made more clear. Some comment saying so explicitly so. Something like "Starting with version 1.1.0-pre4, OpenSSL automatically uses pthreads / the windows locking API without needing help. See commit  openssl/openssl@2e52e7df518d80188c865ea3f7bb3526d14b0c08."

modules/schat.cpp
welwood08: I looked into it a bit, `SSL_CIPHER_get_name` seems to have been available in 0.9.8 (does znc still support older than that?) - what's changed is that 1.1 now requires it to extract the name from the opaque cipher.

It's `SSL_SESSION_get0_cipher` that was only implemented a few weeks before 1.1's release and is needed to get the cipher out of the opaque session, so this is what I'd implement a shim for (unless pre-0.9.8 support is also still needed).

psychon: @DarthGandalf gets to decide the minimum supported version. However, I think that pre-0.9.8 support isn't much of a problem and we could at least try to work there (dunno if it has to be tested explicitly for if it works).

welwood08:
Rebased on latest 1.6.x branch, first commit ammended to include @psychon's suggestions (subsequent commits unchanged) and forced push to here. As before, I have compiled successfully against OpenSSL 1.1.0, 1.0.2h and 0.9.8g.

Currently I don't think this will work with OpenSSL versions older than 0.9.8 because they probably don't have `SSL_CIPHER_get_name`. If support is still desired, a similar macro shim in schat.cpp should not be a problem, but I don't have an easy way to test it.

modules/schat.cpp
DarthGandalf: According to https://www.openssl.org/news/changelog.html 0.9.7 was in 2005 and earlier. I think that's long enough to not care too much about it.

welwood08: Having looked at the implementation of `SSL_CIPHER_get_name` and digging through blame/history, it seems the function existed in the original SSLeay back in 1998 and has only had minor changes (adding consts to the signature) since then so my concern is no longer valid anyway. Additionally, it has always returned "(NONE)" when a NULL cipher is given, so slightly less code needed in schat.cpp.

src/Utils.cpp
```diff
@@ -60,70 +65,69 @@ void CUtils::GenerateCert(FILE *pOut, const CString& sHost) {
 	unsigned int uSeed = (unsigned int)time(NULL);
 	int serial = (rand_r(&uSeed) % 9999);
 
-	RSA *pRSA = RSA_generate_key(2048, 0x10001, NULL, NULL);
-	if ((pKey = EVP_PKEY_new())) {
-		if (!EVP_PKEY_assign_RSA(pKey, pRSA)) {
-			EVP_PKEY_free(pKey);
-			return;
-		}
+#ifdef HAVE_RSA_GENERATE_KEY_EX
```
DarthGandalf: Does this work with all 0.9.8 and later? If so, let's remove the "else" part and the whole ifdef.

src/Utils.cpp
```diff
@@ -60,70 +65,69 @@ void CUtils::GenerateCert(FILE *pOut, const CString& sHost) {
...
-			return;
 		}
-
-		PEM_write_X509(pOut, pCert);
-		X509_free(pCert);
 		EVP_PKEY_free(pKey);
-
-		fprintf(pOut, "%s", szDefaultDH2048);
 	}
+	RSA_free(pRSA);
```
DarthGandalf:
```c++
std::unique_ptr<RSA, void (*)(RSA*)> pRSA(RSA_new(), ::RSA_free);
```

It prevents memory leaks while still allowing early return.

src/Utils.cpp
welwood08: The function was added in 0.9.8 betas. I'll remove the "else" code path, dropping support here for older OpenSSL.

src/Utils.cpp
welwood08: I'm not familiar with that syntax, but it looks fairly self explanatory. Assuming I can simply replace `RSA *pRSA = NULL;` with that and get rid of the later `RSA_free(pRSA);` line, should I do the same for other pointers that need freeing while I'm here?

DarthGandalf: The same idea would work for other types too, of course.

welwood08: After some more reading I agree smart pointers are the right solution here, allowing early return without risk of memory leaks. Annoyingly, all the OpenSSL functions want raw pointers so it's not quite as simple as replacing the definitions - the price paid for eliminating the pyramid of doom.

I'll push changes once I've made sure it compiles against all 3 of my OpenSSLs.

DarthGandalf: `.get()` is not a big price for this benefit.

DarthGandalf: Or, `&*p` which looks strange, but 4 characters shorter.

welwood08:
Rebased on latest 1.6.x again, 2nd commit amended to include @DarthGandalf's suggestions (other commits unchanged since last rebase). Compiles successfully with the aforementioned OpenSSL versions.

DarthGandalf:
Have you tested `znc --makepem`? Which contains the biggest change.
```
znc -d /tmp --makepem
[ .. ] Writing Pem file [/tmp/znc.pem]...
=================================================================
==29873==ERROR: AddressSanitizer: attempting double-free on 0x60b000015540 in thread T0:
    #0 0x61b720 in free /mnt/3tbhome/portage-tmp/portage/sys-devel/llvm-3.8.0-r1/work/llvm-3.8.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:38
    #1 0x7fb42c77db54 in CRYPTO_free /mnt/3tbhome/portage-tmp/portage/dev-libs/openssl-1.0.2h-r2/work/openssl-1.0.2h-abi_x86_64.amd64/crypto/mem.c:442
    #2 0x7fb42c6e9651 in RSA_eay_finish /mnt/3tbhome/portage-tmp/portage/dev-libs/openssl-1.0.2h-r2/work/openssl-1.0.2h-abi_x86_64.amd64/crypto/rsa/rsa_eay.c:896
    #3 0x7fb42c6eba61 in RSA_free /mnt/3tbhome/portage-tmp/portage/dev-libs/openssl-1.0.2h-r2/work/openssl-1.0.2h-abi_x86_64.amd64/crypto/rsa/rsa_lib.c:229
    #4 0x7f375d in std::unique_ptr<rsa_st, void (*)(rsa_st*)>::~unique_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/include/g++-v4/bits/unique_ptr.h:236:4
    #5 0x7ec6f9 in CUtils::GenerateCert(_IO_FILE*, CString const&) /home/somebody/znc/git/build-1.6/../src/Utils.cpp:125:1
    #6 0x6acad5 in CZNC::WritePemFile() /home/somebody/znc/git/build-1.6/../src/znc.cpp:298:2
    #7 0x651b55 in main /home/somebody/znc/git/build-1.6/../src/main.cpp:270:3
    #8 0x7fb4293ab60f in __libc_start_main /mnt/3tbhome/portage-tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/csu/libc-start.c:289
    #9 0x576bf8 in _start (/home/somebody/znc/git/build-1.6/znc+0x576bf8)

0x60b000015540 is located 0 bytes inside of 104-byte region [0x60b000015540,0x60b0000155a8)
freed by thread T0 here:
    #0 0x61b720 in free /mnt/3tbhome/portage-tmp/portage/sys-devel/llvm-3.8.0-r1/work/llvm-3.8.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:38
    #1 0x7fb42c77db54 in CRYPTO_free /mnt/3tbhome/portage-tmp/portage/dev-libs/openssl-1.0.2h-r2/work/openssl-1.0.2h-abi_x86_64.amd64/crypto/mem.c:442
    #2 0x7ec6ed in CUtils::GenerateCert(_IO_FILE*, CString const&) /home/somebody/znc/git/build-1.6/../src/Utils.cpp:125:1
    #3 0x6acad5 in CZNC::WritePemFile() /home/somebody/znc/git/build-1.6/../src/znc.cpp:298:2
    #4 0x651b55 in main /home/somebody/znc/git/build-1.6/../src/main.cpp:270:3
    #5 0x7fb4293ab60f in __libc_start_main /mnt/3tbhome/portage-tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/csu/libc-start.c:289

previously allocated by thread T0 here:
    #0 0x61ba28 in __interceptor_malloc /mnt/3tbhome/portage-tmp/portage/sys-devel/llvm-3.8.0-r1/work/llvm-3.8.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:52
    #1 0x7fb42c77d8df in CRYPTO_malloc /mnt/3tbhome/portage-tmp/portage/dev-libs/openssl-1.0.2h-r2/work/openssl-1.0.2h-abi_x86_64.amd64/crypto/mem.c:342

SUMMARY: AddressSanitizer: double-free /mnt/3tbhome/portage-tmp/portage/sys-devel/llvm-3.8.0-r1/work/llvm-3.8.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:38 in free
==29873==ABORTING
```

Integration test of master branch includes --makeconf which also generates certificate, but 1.6 branch doesn't have that test...

DarthGandalf:
Probably some of the new `.get()`s should be `.release()` instead.

welwood08:
I hadn't, but you're right it does fail. Are the changes needed to compile with asan simple enough (I haven't looked into it yet)?

`src/Utils.cpp:125` isn't inside a function I've modified in the code I have so that's not particularly helpful yet.

src/Utils.cpp
```diff
@@ -59,79 +60,68 @@ constexpr const char* szDefaultDH2048 =
...
-			EVP_PKEY_free(pKey);
-			return;
-		}
+	std::unique_ptr<BIGNUM, void (*)(BIGNUM*)> pExponent(BN_new(), ::BN_free);
+	if (!pExponent || !BN_set_word(pExponent.get(), 0x10001))
+		return;
 
-		PEM_write_RSAPrivateKey(pOut, pRSA, NULL, NULL, 0, NULL, NULL);
+	std::unique_ptr<RSA, void (*)(RSA*)> pRSA(RSA_new(), ::RSA_free);
+	if (!pRSA || !RSA_generate_key_ex(pRSA.get(), 2048, pExponent.get(), NULL))
```
psychon: This should be `pRSA.release()` instead of `.get()`:
http://linux.die.net/man/3/evp_pkey_assign_rsa
> EVP_PKEY_assign_RSA() EVP_PKEY_assign_DSA(), EVP_PKEY_assign_DH() and EVP_PKEY_assign_EC_KEY() also set the referenced key to key however these use the supplied key internally and so key will be freed when the parent pkey is freed.

src/Utils.cpp
```diff
@@ -59,79 +60,68 @@ constexpr const char* szDefaultDH2048 =
...
+		return;
 
-		if (!(pCert = X509_new())) {
-			EVP_PKEY_free(pKey);
-			return;
-		}
+	std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)> pKey(EVP_PKEY_new(), ::EVP_PKEY_free);
+	if (!pKey || !EVP_PKEY_assign_RSA(pKey.get(), pRSA.get()))
```
psychon: Whoops, the above comment should be here instead.

Alternatively, use `EVP_PKEY_set1_RSA` instead. The docs make it sound like it creates a copy of the key.

psychon:
Another comment, just for completeness:

> src/Utils.cpp:125 isn't inside a function I've modified in the code I have so that's not particularly helpful yet.

It's at the end of the function with all the `std::unique_ptr`s. It's where they are destroyed and thus also free their contained pointer.

src/Utils.cpp
welwood08: Interesting, using `release` here would have worked if I hadn't moved `PEM_write_RSAPrivateKey` later (was originally writing the private key using pRSA even though the function could still return early before writing the rest of the certificate, which is a bit pointless). It's my understanding that `get` returns null after `release`, so if I want to use that I have to move `PEM_write_RSAPrivateKey` back before that line.

Since we're going to the trouble of having a unique_ptr that makes sure pRSA is freed, I think it makes sense to change this to use `EVP_PKEY_set1_RSA` instead (which internally just calls `EVP_PKEY_assign_RSA` and ups the ref count).

welwood08:
> It's at the end of the function with all the std::unique_ptrs. It's where they are destroyed and thus also free their contained pointer.

Of course that makes perfect sense and is obvious now I think about it, I was looking at my changes in isolation on top of 1.6.3 release code (which is what I'm testing currently) rather than the result of rebasing on latest 1.6.x (which seems to have added some defines in this file).

welwood08:
Rebased with 2nd commit amended to include @psychon's `EVP_PKEY_set1_RSA` suggestion, no other changes. Built with asan on without error and successful `--makepem` against all 3 tested OpenSSLs.

DarthGandalf:
@welwood08 may you take a look what caused this on OS X? 
```
Undefined symbols for architecture x86_64:
  "_SSL_CIPHER_get_name", referenced from:
      CSChat::OnModCommand(CString const&) in schat.cpp.o
ld: symbol(s) not found for architecture x86_64
```
https://travis-ci.org/znc/znc/builds/159726400
The result differs between autoconf and CMake builds.
OpenSSL version is 1.0.2h from homebrew, according to logs.

psychon:
@DarthGandalf To me that seems to be an already pre-existing bug. If I read `configure.ac` correctly, the OpenSSL flags end up being used for modules. However, in CMake I do not see anything like this. Thus, CMake just doesn't link modules against libssl.
The reason it only breaks now likely is that previously no module used any OpenSSL function?

welwood08:
Is the prefixed underscore significant in any way? `SSL_CIPHER_get_name` has been in OpenSSL since 2005, `_SSL_CIPHER_get_name` does not exist as far as I'm aware.

psychon:
Ignore the underscore. It's an implementation detail. Everything called `foo` in C becomes `_foo` as a symbol (at least on some systems and apparently Mac OS is one of them).
The `-undefined suppress` that autoconf adds means "ignore undefined symbols and just continue". CMake doesn't use this flag and thus the linker complains about the undefined symbol. A proper fix would likely be to link modules which have `REQUIRE_SSL` defined against OpenSSL, but I'll leave this task to someone else. (Or just unconditionally link all modules against OpenSSL, that shouldn't do any harm)

DarthGandalf:
That fixed it, thanks.

Issue #1299, opened 2016-07-28
route_replies + irssi localhost'd into znc will cause irssi to disconnect under certain conditions. 


RealKindOne:
Not sure if its irssi bug or not, but route_replies said report it...

irssi <---> localhost <---> znc 

irssi will stay connected to znc while attempting to connect to ```chat.fake.net``` since it does not exist.
https://dl.dropboxusercontent.com/u/16148324/znc/irssi-route-replies-attempting-server.txt

irssi will lose connection to znc while attempting to connect to``` irc.freenude.net``` (domain squatting?) after a few minutes.
https://dl.dropboxusercontent.com/u/16148324/znc/irssi-route-replies-attempting-server2.txt

mIRC <---> localhost <---> znc

mIRC just stays connected in both cases
https://dl.dropboxusercontent.com/u/16148324/znc/mirc-route-replies-attempting-server.txt
https://dl.dropboxusercontent.com/u/16148324/znc/mirc-route-replies-attempting-server2.txt

znc.conf
https://dl.dropboxusercontent.com/u/16148324/znc/znc-2016-07-28.conf

user@localhost ~ $ znc --version
ZNC 1.7.x-git-623-f1ef8c7 - http://znc.in
IPv6: yes, SSL: yes, DNS: threads, charset: yes, build: autoconf, i18n: no**

```./configure --prefix=/usr --enable-perl --enable-python --enable-debug --enable-tcl --enable-cyrus```

user@localhost ~ $ irssi --version
irssi 0.8.20-g3e28bf7 (20160717 1237)

```./configure --prefix=/usr --enable-silent-rules --enable-true-color --with-socks --with-bot --with-proxy```

mIRC 7.45.769 Beta.

```
user@localhost ~ $ ping chat.fake.net
ping: chat.fake.net: No address associated with hostname
user@localhost ~ $ ping irc.freenude.net
PING irc.freenude.net (199.59.243.120) 56(84) bytes of data.
^C
--- irc.freenude.net ping statistics ---
6 packets transmitted, 0 received, 100% packet loss, time 5039ms
```

psychon:
ZNC:

> [2016-07-28 08:24:08.386000] (KindOne/freenode) CLI -> ZNC [PING irc.znc.in]
> [2016-07-28 08:24:08.386231] (KindOne/freenode) ZNC -> IRC [PING irc.znc.in]
[...]

PR #1314, opened 2016-09-12
route_replies: fix module to check if we are not connected to a network.

RealKindOne:
psychon gets credit for this code change.


psychon:
Looks fine to me. Did I "fix" this on IRC, or what?

RealKindOne:
You suggested it, it works, so it fixed it.
https://github.com/znc/znc/issues/1299#issuecomment-236127513

PR #1322, opened 2016-09-30
nickserv: support messages from X3 services

Zimmedon:
This patch will enable the NickServ module to handle messages from the X3 services daemon

### DIFF
--- a/modules/nickserv.cpp
+++ b/modules/nickserv.cpp
@@ -110,6 +110,8 @@ class CNickServ : public CModule {
              sMessage.find("If this is your nick, type") != CString::npos ||
              sMessage.find("This is a registered nickname, please identify") !=
                  CString::npos ||
+             sMessage.find("is a registered nick - you must auth to account") !=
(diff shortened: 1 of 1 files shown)


rubinlinux:
X3 maintainer here (and znc user) - thanks for proposing this zimmedon.  I wonder, though if it should not be made configurable like the nickserv nick, and identify command.  While this would work in most cases today, x3 has a mechanism for being translated, and for users to individually choose their supported language. If this message were in French it would break ZNC again.

Maybe if no custom 'nick in use' message is found it uses the code here, but if one is defined it uses it instead?

DarthGandalf:
@rubinlinux the whole nickserv module is a big hack. The proper way to auth is SASL, and it doesn't depend on nick of nickserv, nor on language selected by user.

Zimmedon:
@rubinlinux You have a point in regards to language translation. Perhaps being able to configure the detection string would be a better approach. That or using SASL like @DarthGandalf suggests.

PR #1324, opened 2016-10-01
keepnick: improve behaviour by listening to ircd-side numeric errors

### LINKED ISSUE #945: keepnick: change nicks in +N channels
#### Problem
I've encountered a problem that keepnick will just blindly keep trying to get the nick it's supposed to even if the server replies to all such requests with information that the user is in a channel which disallows nick changes. It's a particularly inconvenient thing when joined into many channels which do that.

#### Proposed solution
What keepnick could do is listen out for such failed nickchange replies and temporarily part from the channels in question. Then do the nickchange and rejoin. What should be, however, also accounted for is if how will ZNC be able to rejoin the channel (passworded/invite-only channels). It might not be particularly elegant so maybe keepnick should have an option whether do be this agressive or not.

#### Details
The types of reply vary per IRCd making this harder to nicely implement.
Reply on **UnrealIRCd**: (only sent for one channel when in multiple channels with +N)
`:server 447 nick1 :Can not change nickname while on #chan (+N)`
Reply on **Freenode**: (doesn't even have +N but has same kind of block possible)
`:server 435 nick1 nick2 ##chan :Cannot change nickname while banned on channel`
There might be more different ways for other IRCd-s.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11101115-keepnick-change-nicks-in-n-channels?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Zarthus:
keepnick will now abort if it encounters an unrecoverable error rather than flooding the ircd every 30 seconds with a NICK. The user will be notified that keepnick cannot change nicks due to a numeric from the ircd. 

What the user will have to do depends on the error, which the ircd provides a description of. In the case of charybdis an example would be "Cannot change nickname while banned on channel". While it's possible to catch this, part the channel, change nick, and rejoin -- this causes noise, may not work, you may be affected by multiple bans (which becomes more complex and is more error prone), and it may be a collision between numerics.

In short, if this is nothing keepnick can solve, it stops bothering and alerts the user they need to fix it.

In the backend we've also modernized the module to use OnNumericMessage instead of OnRaw, making the usage easier, more maintainable, and easier to read.

----

Tested on charybdis. 
```
(19:24:33) <*keepnick> Unable to obtain nick: ####z-test: :Cannot change nickname while banned on channel
```

----
Fixes #945

### DIFF
--- a/modules/keepnick.cpp
+++ b/modules/keepnick.cpp
@@ -174,7 +174,7 @@ class CKeepNickMod : public CModule {
                 return HALT;
             }
 
-            PutModule("Unable to obtain nick: " + numeric.GetParam(2) + ": " + numeric.GetParams(3));
+            PutModule("Unable to obtain nick: " + numeric.GetParam(2) + ": " + numeric.GetParam(3));
             Disable();
         }
 
--- a/modules/keepnick.cpp
+++ b/modules/keepnick.cpp
@@ -164,17 +164,17 @@ class CKeepNickMod : public CModule {
     }
 
     EModRet OnNumericMessage(CNumericMessage& numeric) override {
-        //              bad nick                    chan is +N (unrealircd)     banned on chan (charybdis)
-        if (m_pTimer && (numeric.GetCode() == 433 || numeric.GetCode() == 447 || numeric.GetCode() == 435)) {
-            CString sLine = numeric.ToString();
-
+        if (m_pTimer &&
+                (numeric.GetCode() == 433 || /* bad nick */
+                 numeric.GetCode() == 435 || /* user is banned on channel (charybdis) */
+                 numeric.GetCode() == 447)) { /* channel is +N (unrealircd) */
             // Are we trying to get our primary nick and we caused this error?
             // :irc.server.net 433 mynick badnick :Nickname is already in use.
-            if (numeric.GetCode() == 433 && sLine.Token(3).Equals(GetNick())) {
+            if (numeric.GetCode() == 433 && numeric.GetParam(1).Equals(GetNick())) {
                 return HALT;
             }
 
-            PutModule("Unable to obtain nick: " + sLine.Token(4) + ": " + sLine.Token(5, true, " "));
+            PutModule("Unable to obtain nick: " + numeric.GetParam(2) + ": " + numeric.GetParams(3));
             Disable();
         }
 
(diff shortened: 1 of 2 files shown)


modules/keepnick.cpp
```diff
@@ -163,12 +163,20 @@ class CKeepNickMod : public CModule {
...
-    EModRet OnRaw(CString& sLine) override {
-        // Are we trying to get our primary nick and we caused this error?
-        // :irc.server.net 433 mynick badnick :Nickname is already in use.
-        if (m_pTimer && sLine.Token(1) == "433" &&
-            sLine.Token(3).Equals(GetNick()))
-            return HALT;
+    EModRet OnNumericMessage(CNumericMessage& numeric) override {
+        //              bad nick                    chan is +N (unrealircd)     banned on chan (charybdis)
+        if (m_pTimer && (numeric.GetCode() == 433 || numeric.GetCode() == 447 || numeric.GetCode() == 435)) {
+            CString sLine = numeric.ToString();
```
DarthGandalf: CNumericMessage already has the message parsed. Why do you need to stringify it, and then parse with Token() again?
Also, by doing this you're adding a bogus `:` in the beginning of error message.

modules/keepnick.cpp
```diff
@@ -163,12 +163,20 @@ class CKeepNickMod : public CModule {
...
     }
 
-    EModRet OnRaw(CString& sLine) override {
-        // Are we trying to get our primary nick and we caused this error?
-        // :irc.server.net 433 mynick badnick :Nickname is already in use.
-        if (m_pTimer && sLine.Token(1) == "433" &&
-            sLine.Token(3).Equals(GetNick()))
-            return HALT;
+    EModRet OnNumericMessage(CNumericMessage& numeric) override {
+        //              bad nick                    chan is +N (unrealircd)     banned on chan (charybdis)
```
DarthGandalf: I'd prefer inline comments (e.g. with new lines near `||`) rather than such formatting with spaces.

Zarthus:
The requested changes are in, but it does not get rid of the leading `:`, is there a better method than GetParams?

DarthGandalf:
No real docs for that yet... Just follow the examples?
It's in test/integration/main.cpp, first several helper functions are defined, then they are used in actual tests.

PR #1325, opened 2016-10-01
Message.h: Add a comment briefly explaining how to use the class

Zarthus:
and exactly a message is.

It annoyed me there was no real way of knowing what each index meant, I spent a few minutes printing everything out with DEBUG() and this should be pretty accurate.

### DIFF
--- a/include/znc/Message.h
+++ b/include/znc/Message.h
@@ -39,6 +39,30 @@ class CChan;
 class CClient;
 class CIRCNetwork;
 
+/**
+ * Here is a small explanation of how messages on IRC work, and how you can use
+ * this class to get useful information from the parsed message. The output varies
+ * greatly and this advice may not apply to every message type, but this will hopefully
+ * help you understand how it works more accurately.
+ *
+ * @time=some-timestamp :server.network.net 366 something #channel :End of /NAMES list.
+ * tags                  nick               0   1         2        3
+ *
+ * `tags` is the IRCv3 tags associated with the message, obtained with GetTag(s).
+ *   the @time tag can also be obtained with GetTime(), some messages have other tags 
+ *   with them. Tags may not always be present. Refer to IRCv3 for documentation on tags.
+ * `nick` is the sender, which can be obtained with GetNick()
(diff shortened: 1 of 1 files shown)


include/znc/Message.h
```diff
@@ -39,6 +39,25 @@ class CChan;
...
+ * this class to get useful information from the parsed message.
+ *
+ * @time=some-timestamp :server.network.net 366 something #channel :End of /NAMES list.
+ * tags                                     0   1         2        3
+ *
+ * `tags` is the IRCv3 tags associated with the message, obtained with GetTag(s).
+ * `0` is the command, which is obtained via GetCommand, or with GetParam(0)
+ * `1` is the first parameter index, which usually includes a sender or a recipient.
+ * `2` is the target or element the message is about.
+ * `3` is the message.
```
DarthGandalf: This depends on the message type. What you describe here, is true only for 366 numeric, but it sounds like parameters 0-3 have this exact meaning always. Not every message has exactly this number of parameters.

include/znc/Message.h
```diff
@@ -39,6 +39,25 @@ class CChan;
 class CClient;
 class CIRCNetwork;
 
+/**
+ * Here is a small explanation of how messages on IRC work, and how you can use
+ * this class to get useful information from the parsed message.
+ *
+ * @time=some-timestamp :server.network.net 366 something #channel :End of /NAMES list.
+ * tags                                     0   1         2        3
```
DarthGandalf: `:server.network.net` can be retrieved with `GetNick()`

include/znc/Message.h
```diff
@@ -39,6 +39,25 @@ class CChan;
...
+ *
+ * `tags` is the IRCv3 tags associated with the message, obtained with GetTag(s).
```
DarthGandalf: `GetTag("time")`, but time is probably not the best example, because there is also `GetTime()`, which works even if the message didn't have a `time` tag.

Zarthus: Time is probably the most common tag, I've added some more elaboration on tags, though.

PR #1326, opened 2016-10-04
Fix newest additions to keepnick module and write the test for it.

DarthGandalf:
Fix the documentation.

See #1324 and #1325 

@Zarthus

### DIFF
--- a/modules/keepnick.cpp
+++ b/modules/keepnick.cpp
@@ -163,19 +163,28 @@ class CKeepNickMod : public CModule {
         return CONTINUE;
     }
 
-    EModRet OnNumericMessage(CNumericMessage& numeric) override {
-        if (m_pTimer &&
-                (numeric.GetCode() == 433 || /* bad nick */
-                 numeric.GetCode() == 435 || /* user is banned on channel (charybdis) */
-                 numeric.GetCode() == 447)) { /* channel is +N (unrealircd) */
+    EModRet OnNumericMessage(CNumericMessage& msg) override {
+        if (m_pTimer) {
             // Are we trying to get our primary nick and we caused this error?
             // :irc.server.net 433 mynick badnick :Nickname is already in use.
-            if (numeric.GetCode() == 433 && numeric.GetParam(1).Equals(GetNick())) {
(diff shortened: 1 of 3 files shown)


Zarthus:
that's better, thanks! I agree an enumeration of params is better than rather than statically explaining which one is which with ambiguous words like "maybe contains this".

The keepnick getparam() worked for me when I tested it on freenode, I will check if this is still the case with your fix.

modules/keepnick.cpp
```diff
@@ -174,7 +174,7 @@ class CKeepNickMod : public CModule {
                 return HALT;
             }
 
-            PutModule("Unable to obtain nick: " + numeric.GetParam(2) + ": " + numeric.GetParam(3));
+            PutModule("Unable to obtain nick: " + numeric.GetParam(1) + ": " + numeric.GetParam(2));
```
Zarthus: The GetParam calls were definitely correct, although I can see where you were coming from.

By changing GetParam's ints around, the string changed from the following:

```
Unable to obtain nick: ####z-test: Cannot change nickname while banned on channel
Unable to obtain nick: #channel: ReasonByIrcd
```

To:

```
Unable to obtain nick: zarthue: ####z-test
Unable to obtain nick: NickInQuestion: TargetChannel
```

The idea was to include the ircd provided reason so they know *why* the nick change was not going through. I omitted the nick (Param 1) because your primary nickname is unlikely to change, and you are likely to know it already. Inclusion is not a bad thing (Changing the string to "Unable to obtain nick $nick: $channel: $reason"), but I personally omitted it because I did not think it was relevant.

DarthGandalf: I just took your test from your PR (you deleted that test), and made it work. I don't know the actual format of 435 and 447. The test now matches the behavior.
That's one of the points of tests, after all.
If the format is wrong, may you suggest fixes to test and to this string?

DarthGandalf: "message" is just a last parameter, there's nothing special about number of parameters.
`GetParam(2)` will be empty for 447, because `GetParam(1)` has leading colon and therefore is the last parameter.
Is `goodstuff` the current nick?

Zarthus: Yes.

PR #1327, opened 2016-10-04
Module CSRF override support (mark II)

lol768:
This is essentially a duplicate of tzvetkoff's work in #1296 with fixes per the comments on that PR.


lol768:
Rebased following commits.

PR #1333, opened 2016-10-21
stickychan: stop checking so often, increase delay to once every 3 mi…

Zarthus:
…nutes

once every 15 seconds is too often to be useful.

If the usecase is to prevent parting, StickyChan should halt ONRAW PART.

### DIFF
--- a/modules/stickychan.cpp
+++ b/modules/stickychan.cpp
@@ -17,6 +17,8 @@
 #include <znc/Chan.h>
 #include <znc/IRCNetwork.h>
 
+#define STICKYCHAN_TIMER_INTERVAL 180
+
 using std::vector;
 
 class CStickyChan : public CModule {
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Is there any particular reason why 15 seconds is too small?

Zarthus:
If you can't join the channel the first time the timer hits, chances are you won't be able the next 15 seconds after either. StickyChan should listen to numerics/kicks(configurable)/etc and determine why it can't join if it wants to help the user join the channel, not "brute force" it's way in.

It already stops you from parting accidentally, the timer is just a "in case I get kicked or whatever", in which case *kickrejoin exists if you want to be annoying. With a 180 second delay stickychan does not function as autorejoin on kick and the chance of it(the warning) slipping by you is felt better.

Zarthus:
I'm okay with either, I just went with 180 because it functions the same as kickrejoin with anything less (and as a channel operator I don't want stickychan to be a hard-to-detect kickrejoin)

Zarthus:
Modified timer to 60 seconds.

PR #1338, opened 2016-11-05
sasl: Make the first requirement for SET actually mandatory, return i…

Zarthus:
…nformation about settings if no input for SET

This changes SASL to have the first parameter of `set` actually be required, by responding with information about the current configuration when no parameters are passed.

I come across it from time to time where I try to look up my info, and with `/msg *sasl set` I accidentally overwrite my existing settings. Oops!

---

Now returns the following message:

```
Username is currently set to 'Zarthus', a password was supplied.
```

### DIFF
--- a/modules/sasl.cpp
+++ b/modules/sasl.cpp
@@ -101,6 +101,15 @@ class CSASLMod : public CModule {
     }
 
     void Set(const CString& sLine) {
+        if (sLine.Token(1).empty()) {
+            CString sUsername = GetNV("username");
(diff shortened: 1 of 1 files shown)


GLolol:
To extend upon this, it would be even better if this behaviour was actually mentioned in `/msg *sasl help`. A `status` command with similar output could also make this more user friendly.

Zarthus:
@GLolol 

Thanks for pointing that out, I've made another PR to update this.

PR #1343, opened 2016-11-16
Fix version.sh for ancient git

welwood08:
I'm not sure when the `format` argument was added to `git log`, but it seems to be an alias for `pretty` (ie both exist in current versions of git). I noticed this was failing on my ancient system that's stuck with git 1.5, this makes it work.

Also, `ignore-submodules` is not understood by ancient `git status` - I believe the expected output would be achieved by simply removing this argument, but it's not important since it fails gracefully enough and fixing it so newer git doesn't fail seems more complicated.

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -16,7 +16,7 @@ else
 
 	# Figure out the information we need
 	LATEST_TAG=`${GIT} describe --abbrev=0 HEAD`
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Is C++ compiler new enough on your system with ancient git?

welwood08:
I've been using my own build of latest gcc release for ages, just changed my script today to build znc from git instead of release and working out teething problems like this.

PR #1345, opened 2016-11-27
Strip Control Codes from various settings.

ChasedSpade:
The controlpanel module and /znc commands do not strip control codes from essential ZNC and IRC server settings. This commit strips color codes on important settings.
- Password (A hash is generated with the control code characters allowing login with a colored password.)
- Nick/ident (Most IRCds do not allow special characters in nick and ident.)
- Integers (When setting an integer value, the new value is set to 0 rather than the intended value.)
- Addchan (ZNC will add the channel with the color code, allowing a color code channel with the same name as a non-color coded channel.)
- Addserver (ZNC will add the server which ZNC will fail to resolve.)
- Bindhost (ZNC cannot bind to colored addresses as they do not exist.)

### DIFF
--- a/modules/controlpanel.cpp
+++ b/modules/controlpanel.cpp
@@ -286,7 +286,7 @@ class CAdminMod : public CModule {
     void Set(const CString& sLine) {
         const CString sVar = sLine.Token(1).AsLower();
         CString sUserName = sLine.Token(2);
-        CString sValue = sLine.Token(3, true);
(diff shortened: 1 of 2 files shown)


Zarthus:
Problems with pesky mIRC users that have scripts that always send in colours? :P

ChasedSpade:
Been gone for a while and never got the chance to respond. I'll take another look at this.

Issue #1274, opened 2016-05-11
ZNC Crashes on start


Flare183:
After upgrading my Ubuntu Server, my new ZNC (Version ZNC 1.6.3+deb1 - http://znc.in
IPv6: yes, SSL: yes, DNS: threads, charset: yes), now crashes with the following message:

[879:878 - 0:503] 05:48:42 [jesse@quartz:/dev/pts/0 +1] ~ 
$ znc
[ .. ] Checking for list of available modules...
*** Error in `znc': free(): invalid pointer: 0xbfb87bf4 ***
======= Backtrace: =========
/lib/i386-linux-gnu/libc.so.6(+0x67257)[0xb7088257]
/lib/i386-linux-gnu/libc.so.6(+0x6d577)[0xb708e577]
/lib/i386-linux-gnu/libc.so.6(+0x6dd31)[0xb708ed31]
/usr/lib/i386-linux-gnu/libstdc++.so.6(_ZdlPv+0x18)[0xb7280d98]
/usr/lib/i386-linux-gnu/libstdc++.so.6(_ZNSs4_Rep10_M_destroyERKSaIcE+0x18)[0xb72c6ee8]
/usr/lib/i386-linux-gnu/libstdc++.so.6(_ZNSs6assignERKSs+0xac)[0xb72c880c]
/home/jesse/.znc/modules/push.so(ZNCModInfo+0x76)[0xb56ec5f6]
znc(_ZN8CModules10OpenModuleERK7CStringS2_RbR8CModInfoRS0_+0x1cd)[0x810182d]
znc(_ZN8CModules14GetModPathInfoER8CModInfoRK7CStringS4_RS2_+0x31)[0x8101da1]
znc(_ZN8CModules16GetAvailableModsERSt3setI8CModInfoSt4lessIS1_ESaIS1_EENS1_11EModuleTypeE+0x59b)[0x810ed5b]
znc(main+0xb60)[0x8070e10]
/lib/i386-linux-gnu/libc.so.6(__libc_start_main+0xf7)[0xb7039637]
znc[0x8072819]
======= Memory map: ========
08048000-081bc000 r-xp 00000000 08:01 1452369    /usr/bin/znc
081bd000-081be000 r--p 00174000 08:01 1452369    /usr/bin/znc
081be000-081bf000 rw-p 00175000 08:01 1452369    /usr/bin/znc
08fbf000-08fe4000 rw-p 00000000 00:00 0          [heap]
[...]

Issue #1255, opened 2016-03-30
Crash after updating to 1.6.3 from 1.6.2


thepace:
Im runing znc on ubuntu 14.04 lts , compiled it from source. 1.6.2 was working fine compiled from source too using official releases both times. After upgrading to 1.6.3 znc is crashing. Steps to reproduce: start znc from command line, top lists it on process list, webui works. Then i try to connect to bounces using mirc from local machine (ZNC runs on remote server) and it dies. Guys from #znc told me that this info might be useful for investigating whats going on http://pastebin.com/A3HbQKrB
##

thepace:
http://pastebin.com/9W54Gw9P this is what i got via gdb before typing bt full

DarthGandalf:
May you show last lines of `znc -D` before the crash?
What modules do you have loaded?

thepace:
Can u give me step by step for that , i start znc load mirc = crash , should i start znc type znc -d and crash it or what ?

psychon:
Yup, you should start znc with `znc -D` and then make it crash. You will get lots of output from ZNC (which could also include passwords, so don't copy everything blindly). We are interested in this output (at least the last part of it, let's say... the last 100 lines?)

DarthGandalf:
From discussion on IRC:
There was an external module built for 1.6.2. The crash was fixed after rebuilding that module.
Though update between 1.6.x versions should require rebuild...

PR #1353, opened 2016-12-21
Refactor the way how modules are loaded.

DarthGandalf:
Make version checks more strict.

This finishes attempt to preserve ABI between patch versions. That
didn't work well, and the people who could make it work, left the
project already.

### DIFF
--- a/include/znc/Modules.h
+++ b/include/znc/Modules.h
@@ -60,25 +60,53 @@ class CModInfo;
 #define ZNC_EXPORT_LIB_EXPORT
 #endif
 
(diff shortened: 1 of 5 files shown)


psychon:
Have there been more problems of this kind lately? I never thought about it, but are likely right that ssl and ipv6 affects the ABI as well...
Dunno if ABI-comatibility could ever work well. It's quite hard and even harder when "everything is API". :-)

DarthGandalf:
Yep, #1255 #1274 
I've also noticed that the maintainer-defined part (`VERSION_EXTRA`, `+deb1`) needs to be compared.

PR #1361, opened 2017-01-11
Fix block_motd incorrectly tracking state on which user requested MOTD

kylef:
There were some problems with the existing implementation:

- If a server has no MOTD `m_bTemporaryAcceptMotd` would be set to true and then subsequent MOTDs would be sent to the current user or another user who has the same instance of the block_motd loaded.
- If the module is loaded globally then the `m_bTemporaryAcceptMotd` could be changed by other users and this could lead to timing issues where one user asks for an MOTD and another user may get a full or even partial MOTD. This could lead to sending a start motd or end motd separately and potentially breaking clients.
- Lastly, we should send "MOTD" instead of "motd" as the IRC command to a server as per RFC1459.

This is my first PR in ZNC in some years so please review with caution incase I did something wrong or there is a better way.

### DIFF
--- a/modules/block_motd.cpp
+++ b/modules/block_motd.cpp
@@ -42,9 +42,9 @@ class CBlockMotd : public CModule {
         const CString sServer = sLine.Token(1);
 
         if (sServer.empty()) {
-            PutIRC("motd");
+            PutIRC("MOTD");
         } else {
-            PutIRC("motd " + sServer);
+            PutIRC("MOTD " + sServer);
         }
     }
 
--- a/modules/block_motd.cpp
+++ b/modules/block_motd.cpp
@@ -15,6 +15,9 @@
  */
 
 #include <znc/Modules.h>
+#include <znc/IRCNetwork.h>
+
+using std::set;
 
 class CBlockMotd : public CModule {
   public:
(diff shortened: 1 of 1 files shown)


modules/block_motd.cpp
```diff
@@ -44,21 +52,54 @@ class CBlockMotd : public CModule {
...
+            if (*it == GetNetwork()->GetIRCSock()) {
+                m_vTemporaryAcceptedMotdSocks.erase(it);
+                return;
+            }
+        }
+    }
+
   private:
-    bool m_bTemporaryAcceptMotd = false;
+    vector<CIRCSock *> m_vTemporaryAcceptedMotdSocks;
```
psychon: How about turning this into a `set` and using `count(GetNetwork()->GetIRCSock() > 0` and `erase(GetNetwork()->GetIRCSock()` instead of iterating through the vector?

psychon:
Note that this is my first review in several years, so... ;-)

modules/block_motd.cpp
kylef: That would simplify this a lot, I've updated PR to use set.

Issue #1240, opened 2016-02-07
support for listening on a unix socket, not a port


mauke:
This is a feature request.

It would be nice if znc supported listening on a unix socket, not a tcp port. I would then use ssh to forward a local port (on my client machine) to the unix socket znc is listening on (on a shared server). I don't want other others of the shared server to be able to connect() to my znc. (I know znc supports passwords but why let them get that far?)
##

Wolf480pl:
This would also be useful when reverse-proxying the web panel through some webserver (like nginx).
Doing it with unix sockets would be cleaner, I think, and would allow setting permissions on the socket so that only the webserver can access it.

ilmaisin:
Another use case would be port access control. Most operating systems lack an easy way to restrict access to bind a specific TCP port to a specific user. That means that another user on the same machine can "steal" your port accidentally or intentionally.

With domain sockets, one can use file-system based access control for the sockets and let xinetd to bind the ports on boot time and do the redirection from the port to the domain socket.

cheako:
What can be done about this?
https://github.com/jimloco/Csocket/pull/69

psychon:
@cheako Uhm... what do you mean? That's a merged PR that adds the necessary support to Csocket.

cheako:
I see that Csocket has support.  I didn't realize that means I can use UDS from the ZNC config file, I think there is some glue missing after reading the WIP PR and this "open" issue.

bastien-roucaries:
Any news about this ?

cheako:
I'm out, this project didn't do what I needed and am still using the irccloud platform.

PR #1363, opened 2017-01-12
WIP: Add support for listening on an UNIX socket

psychon:
This implements https://github.com/znc/znc/issues/1240. So far, I did not do much testing of this (feel free to test this!). Also, a little more work is needed so that listing listeners and adding/removing listeners supports this feature properly. However, I'm out of time for this and lazy.
So: Feel free to pick this up and continue working on it.

### DIFF
--- a/include/znc/Listener.h
+++ b/include/znc/Listener.h
@@ -22,6 +22,7 @@
 
 // Forward Declarations
 class CRealListener;
+class CConfig;
 // !Forward Declarations
 
 class CListener {
@@ -60,6 +61,7 @@ class CListener {
 
     bool Listen();
     void ResetRealListener();
+    CConfig ToConfig() const;
 
   private:
   protected:
--- a/include/znc/Listener.h
+++ b/include/znc/Listener.h
@@ -57,6 +57,8 @@ class CListener {
 
   private:
   protected:
+    void setupSSL(CRealListener* listener) const;
+
     bool m_bSSL;
     CString m_sURIPrefix;
     CRealListener* m_pListener;
--- a/include/znc/Listener.h
+++ b/include/znc/Listener.h
@@ -29,27 +29,19 @@ class CListener {
   public:
     typedef enum { ACCEPT_IRC, ACCEPT_HTTP, ACCEPT_ALL } EAcceptType;
 
-    CListener(unsigned short uPort, const CString& sBindHost,
-              const CString& sURIPrefix, bool bSSL, EAddrType eAddr,
-              EAcceptType eAccept)
+    CListener(const CString& sURIPrefix, bool bSSL, EAcceptType eAccept)
         : m_bSSL(bSSL),
-          m_eAddr(eAddr),
-          m_uPort(uPort),
-          m_sBindHost(sBindHost),
           m_sURIPrefix(sURIPrefix),
           m_pListener(nullptr),
           m_eAcceptType(eAccept) {}
 
-    ~CListener();
+    virtual ~CListener();
 
     CListener(const CListener&) = delete;
     CListener& operator=(const CListener&) = delete;
 
     // Getters
     bool IsSSL() const { return m_bSSL; }
-    EAddrType GetAddrType() const { return m_eAddr; }
-    unsigned short GetPort() const { return m_uPort; }
-    const CString& GetBindHost() const { return m_sBindHost; }
     CRealListener* GetRealListener() const { return m_pListener; }
     const CString& GetURIPrefix() const { return m_sURIPrefix; }
     EAcceptType GetAcceptType() const { return m_eAcceptType; }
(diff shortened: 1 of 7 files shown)


include/znc/Listener.h
```diff
@@ -92,6 +92,28 @@ class CTCPListener : public CListener {
...
+    ~CUnixListener();
+
+    CUnixListener(const CUnixListener&) = delete;
+    CUnixListener& operator=(const CUnixListener&) = delete;
+
+    // Getters
+    const CString& GetPath() const { return m_sPath; }
+    // !Getters
+
+    virtual bool Listen();
```
DarthGandalf: s/virtual/override/

src/Listener.cpp
```diff
@@ -83,6 +85,60 @@ CConfig CTCPListener::ToConfig() const {
...
+    memset(&addr, 0, sizeof(addr));
+    addr.sun_family = AF_UNIX;
+    strncpy(addr.sun_path, m_sPath.c_str(), sizeof(addr.sun_path)-1);
+
+    CFile::Delete(m_sPath);
+    if (bind(fd, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr)) < 0) {
+        close(fd);
+        return false;
+    }
+    if (listen(fd, 42) < 0) {
```
DarthGandalf: This loses the error message, which is only at errno at this point, and gets overwritten by close()

src/znc.cpp
```diff
@@ -1717,9 +1717,70 @@ bool CZNC::AddListener(unsigned short uPort, const CString& sBindHost,
...
+    CUtils::PrintAction("Binding to path [" + sPath + "]");
+
+#ifndef HAVE_LIBSSL
+    if (bSSL) {
+        sError = "SSL is not enabled";
+        CUtils::PrintStatus(false, sError);
+        return false;
+    }
+#else
+    CString sPemFile = GetPemLocation();
```
DarthGandalf: SSL code should be shared between TCP and Unix sockets...

src/znc.cpp
```diff
@@ -1730,32 +1791,22 @@ bool CZNC::AddListener(CConfig* pConfig, CString& sError) {
     bool bIRC;
     bool bWeb;
     unsigned short uPort;
+    bool tcpListener = true;
```
DarthGandalf: bTcpListener

psychon:
> there is socket API for ZNC modules, it supports only TCP sockets now, I'd expect it to also get support of Unix sockets

So the unix-socket-listening should be implemented in `CSocket` instead (or whatever that class was called...)?

DarthGandalf:
Yeh, I think so. It still needs some support on the UI/configs side, of course.

DarthGandalf:
Well, either Csock or CZNCSock, whatever makes more sense.

psychon:
> Well, either Csock or CZNCSock, whatever makes more sense.

Apparently Csock: https://github.com/jimloco/Csocket/pull/69

psychon:
Updated this to address comments above. Still lots of WIP left.

src/znc.cpp
```diff
@@ -1788,8 +1820,28 @@ bool CZNC::AddListener(CConfig* pConfig, CString& sError) {
...
+        } else {
+            sError = "No address family given";
+            CUtils::PrintError(sError);
+            return false;
+        }
+
+        return AddListener(uPort, sBindHost, sURIPrefix, bSSL, eAddr, eAccept,
+                           sError);
+    }
+    return AddListener(sPath, sURIPrefix, bSSL, eAccept, sError);
```
DarthGandalf: I think this should be AddUnixListener() vs AddTCPListener()

test/integration/main.cpp
```diff
@@ -901,4 +903,44 @@ TEST_F(ZNCTest, ModuleCrypt) {
...
+    QTemporaryDir tempDir;
+    QDir dir(tempDir.path());
+    QString socketName = dir.filePath("socket");
+    dir.mkdir("configs");
+    QFile config(dir.filePath("configs/znc.conf"));
+    ASSERT_TRUE(config.open(QIODevice::WriteOnly | QIODevice::Text));
+    Z;
+
+    QTextStream out(&config);
+    // FIXME: Hardcoding a version is bad
```
DarthGandalf: Is version necessary there?
If it is, to avoid hardcoding it, you can also use --makeconf, and append stuff to config, like in ModuleCrypt test above.

cheako:
What effect does Csocket support have?
https://github.com/jimloco/Csocket/pull/69

psychon:
> What effect does Csocket support have?

Well... it makes this PR possible? IIRC, this PR already uses that support in Csocket.

cheako:
Thanks for that clarification, from the sound of this PR it uses an unrelated API.  Given that, I'll take a close look at the code here.

psychon:
Yeah, originally this added the needed code in ZNC directly. However, "it was decided" that this is better done in Csocket. Thus, the PR you found. The state of this PR (as far as I remember) is: everything works, but it so far can not be configured everywhere, e.g. webadmin support might be missing. Feel free to pick this up and finish things. You can see DarthGandalf's review from three years ago above.

src/znc.cpp
cheako: The arguments to PrintAction like 1749 and 1683 are diffrent.

PR #1366, opened 2017-01-19
znc makeconf: secure by default

Zarthus:
Default bListenSSL to true when SSL is enabled.

### DIFF
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -653,7 +653,14 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
 #endif
     CString sListenHost;
     CString sURIPrefix;
+
+// Default to true when we have SSL bundled (secure by default).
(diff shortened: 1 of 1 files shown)


psychon:
Wow. I haven't seen so much red in a while. :-)
The tests need to be fixed. Right now things go with the default value for SSL during the --makeconf part of the tests, so after this change the tests try to connect plain-text to a SSL listener.

Zarthus:
Yeah, Travis is just being incredibly slow and I'm waiting for the rest of them to complete to see if anything breaks.

DarthGandalf:
It'll only scare all browsers away.
Probably need to offer creation of letsencrypt cert, instead of self-signed. Of course, this doesn't apply to tests, which should use self-signed cert.

Zarthus:
if we're gonna go the route of having LE do this automatically in the future that's fine, but for now I'lll close the PR then.

PR #1377, opened 2017-02-09
Fixes for the nick-prefix in the crypt module

Cizzle:
Don't use a key which is a valid nickname, now changed to "@nick-prefix@" instead.
If you had loaded the crypt module before, the old nick-prefix will still remain in your config but can safely be removed.

Also check if the nick-prefix and status-prefix aren't the same or overlapping, "*" or "." are then used depending on either value. If they would be the same, you would be messaging a module instead of a user.
And if they'd overlap, depending on the nick, you could be messaging a module as well.

### DIFF
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -45,6 +45,8 @@
 #include <znc/IRCNetwork.h>
 
 #define REQUIRESSL 1
+// To be removed in future versions
+#define NICK_PREFIX_OLD_KEY "[nick-prefix]"
 #define NICK_PREFIX_KEY "@nick-prefix@"
 
 class CCryptMod : public CModule {
@@ -80,6 +82,19 @@ class CCryptMod : public CModule {
     }
 
     ~CCryptMod() override {}
+
+    bool OnLoad(const CString& sArgsi, CString& sMessage) override {
+        MCString::iterator it = FindNV(NICK_PREFIX_KEY);
+        if (it == EndNV()) {
+            /* Don't have the new prefix key yet */
+            it = FindNV(NICK_PREFIX_OLD_KEY);
+            if (it != EndNV()) {
+                SetNV(NICK_PREFIX_KEY, it->second);
+                DelNV(NICK_PREFIX_OLD_KEY);
+            }
+        }
+        return true;
+    }
 
     EModRet OnUserMsg(CString& sTarget, CString& sMessage) override {
         sTarget.TrimPrefix(NickPrefix());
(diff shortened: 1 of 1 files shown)


modules/crypt.cpp
```diff
@@ -45,12 +45,24 @@
...
+         * SP = "*"; NP = "*s"; "tatus" sends an encrypted message appearing at "*status"
+         */
+        CString sStatusPrefix = GetUser()->GetStatusPrefix();
+        if (it != EndNV()) {
+            size_t sp = sStatusPrefix.size();
+            size_t np = it->second.size();
+            if (sStatusPrefix.CaseCmp(it->second, (np > sp) ? sp : np ))
+                return it->second;
+        }
+        return sStatusPrefix.StrCmp("*", 1) ? "*" : ".";
```
DarthGandalf: `.StartsWith("*")`

modules/crypt.cpp
```diff
@@ -45,12 +45,24 @@
...
+        /*
+         * Check for different Prefixes to not confuse modules with nicknames
+         * Also check for overlap for rare cases like:
+         * SP = "*"; NP = "*s"; "tatus" sends an encrypted message appearing at "*status"
+         */
+        CString sStatusPrefix = GetUser()->GetStatusPrefix();
+        if (it != EndNV()) {
+            size_t sp = sStatusPrefix.size();
+            size_t np = it->second.size();
+            if (sStatusPrefix.CaseCmp(it->second, (np > sp) ? sp : np ))
```
DarthGandalf: `std::min`
`.Equals()`

modules/crypt.cpp
```diff
@@ -45,12 +45,24 @@
 #include <znc/IRCNetwork.h>
 
 #define REQUIRESSL 1
-#define NICK_PREFIX_KEY "[nick-prefix]"
+#define NICK_PREFIX_KEY "@nick-prefix@"
```
DarthGandalf: Existing configuration will break...

Cizzle: Could add code to the NickPrefix() function that first checks NICK_PREFIX_KEY (with @), if it doesn't find that to check NICK_PREFIX_KEY_OLD (with []) and to set that value under NICK_PREFIX_KEY then?

modules/crypt.cpp
Cizzle: But the Equals() with length argument is listed as deprecated?

DarthGandalf: Ah, you're right, I didn't notice. Then let's make it more explicit with `!= 0`.

modules/crypt.cpp
DarthGandalf: Yeah, or in OnLoad().
After a while (how long?) that code can be removed.

PR #1378, opened 2017-02-11
Add DH1080 keyexchange to the crypt module.

Cizzle:
Partly based on http://wiki.znc.in/Fish.
Your private and public key are generated once per this module load upon first need (init sent or received).

### DIFF
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -26,10 +26,9 @@
 // TODO:
 //
 // 1) Encrypt key storage file
-// 2) Secure key exchange using pub/priv keys and the DH algorithm
-// 3) Some way of notifying the user that the current channel is in "encryption
+// 2) Some way of notifying the user that the current channel is in "encryption
 // mode" verses plain text
-// 4) Temporarily disable a target (nick/chan)
+// 3) Temporarily disable a target (nick/chan)
 //
 // NOTE: This module is currently NOT intended to secure you from your shell
 // admin.
@@ -43,6 +42,9 @@
 #include <znc/Chan.h>
 #include <znc/User.h>
 #include <znc/IRCNetwork.h>
+#include <openssl/dh.h>
+#include <openssl/bn.h>
+#include <znc/SHA256.h>
 
 #define REQUIRESSL 1
 // To be removed in future versions
@@ -50,6 +52,137 @@
 #define NICK_PREFIX_KEY "@nick-prefix@"
 
 class CCryptMod : public CModule {
+  private:
+    const char *prime1080="FBE1022E23D213E8ACFA9AE8B9DFADA3EA6B7AC7A7B7E95AB5EB2DF858921FEADE95E6AC7BE7DE6ADBAB8A783E7AF7A7FA6A2B7BEB1E72EAE2B72F9FA2BFB2A2EFBEFAC868BADB3E828FA8BADFADA3E4CC1BE7E8AFE85E9698A783EB68FA07A77AB6AD7BEB618ACF9CA2897EB28A6189EFA07AB99A8A7FA9AE299EFA7BA66DEAFEFBEFBF0B7D8B";
+    /* Generate our keys once and reuse, just like ssh keys */
+    DH *dh = NULL;
+    CString sPrivKey;
+    CString sPubKey;
+
+#if OPENSSL_VERSION_NUMBER < 0X10100000L
+    int DH_set0_pqg(DH *dh, BIGNUM *p, BIGNUM *q, BIGNUM *g) {
+        /* If the fields p and g in dh are NULL, the corresponding input
+         * parameters MUST be non-NULL.  q may remain NULL.
+         */
+        if (dh == NULL || (dh->p == NULL && p == NULL) || (dh->g == NULL && g == NULL))
+            return 0;
+
+        if (p != NULL) {
+            BN_free(dh->p);
+            dh->p = p;
+        }
+        if (g != NULL) {
+            BN_free(dh->g);
+            dh->g = g;
+        }
+        if (q != NULL) {
+            BN_free(dh->q);
+            dh->q = q;
+            dh->length = BN_num_bits(q);
+        }
+
+        return 1;
+    }
+
+    void DH_get0_key(const DH *dh, const BIGNUM **pub_key, const BIGNUM **priv_key) {
+        if (dh != NULL) {
+            if (pub_key != NULL)
+                *pub_key = dh->pub_key;
+            if (priv_key != NULL)
+                *priv_key = dh->priv_key;
+        }
+    }
+
+#endif
+
+    bool DH1080_gen() {
+        /* Generate our keys on first call */
+        if (!dh) {
+            int len;
+            unsigned char *buf;
+            const BIGNUM *bPrivKey = NULL;
+            const BIGNUM *bPubKey = NULL;
+            BIGNUM *bPrime = NULL;
+            BIGNUM *bGen = NULL;
+            dh = DH_new();
+
+            if (!BN_hex2bn(&bPrime, prime1080) || !BN_dec2bn(&bGen, "2") || !DH_set0_pqg(dh, bPrime, NULL, bGen) || !DH_generate_key(dh)) {
+                /* one of them failed */
+                if (bPrime)
+                    BN_clear_free(bPrime);
+                if (bGen)
+                    BN_clear_free(bGen);
+                if (dh) {
+                    DH_free(dh);
+                    dh = NULL;
+                }
+                return false;
+            }
+
+            /* Get our keys */
+            DH_get0_key(dh, &bPubKey, &bPrivKey);
+
+            /* Get our private key */
+            len = BN_num_bytes(bPrivKey);
+            buf = (unsigned char *)calloc(len, 1);
+            BN_bn2bin(bPrivKey, buf);
+            /*
+             * Use basic_string constructor with given start and end,
+             *  as we might have \0 and don't want to cast
+             */
+            sPrivKey = CString::basic_string(buf, &buf[len]);
+            sPrivKey.Base64Encode();
+
+            /* Get our public key */
+            len = BN_num_bytes(bPubKey);
+            buf = (unsigned char *)realloc(buf, len);
+            BN_bn2bin(bPubKey, buf);
+            sPubKey = CString::basic_string(buf, &buf[len]);
+            sPubKey.Base64Encode();
+
+            if (buf)
+                free(buf);
+        }
+
+        return true;
+    }
+
+
+    bool DH1080_comp(CString& sOtherPubKey, CString& sSecretKey) {
+        unsigned long len;
+        unsigned char *key;
+        unsigned char digest[SHA256_DIGEST_SIZE];
+        BIGNUM *bOtherPubKey = NULL;
+
+        /* Prepare other public key */
+        len = sOtherPubKey.Base64Decode();
+        bOtherPubKey = BN_bin2bn((unsigned char *)sOtherPubKey.data(), len, NULL);
+
+        /* Generate secret key */
+        key = (unsigned char *)calloc(DH_size(dh), 1);
+        if ((len = DH_compute_key(key, bOtherPubKey, dh)) == -1) {
+            sSecretKey = "";
+            if (bOtherPubKey)
+                BN_clear_free(bOtherPubKey);
+            if (key)
+                free(key);
+            return false;
+        }
+
+        /* Get our secret key */
+        sha256(key, len, digest);
+        sSecretKey = CString::basic_string(digest, &(digest[sizeof digest]));
+        sSecretKey.Base64Encode();
+        sSecretKey.TrimRight("=");
+
+        if (bOtherPubKey)
+            BN_clear_free(bOtherPubKey);
+        if (key)
+            free(key);
+
+        return true;
+    }
+
     CString NickPrefix() {
         MCString::iterator it = FindNV(NICK_PREFIX_KEY);
         /*
@@ -67,6 +200,7 @@ class CCryptMod : public CModule {
         return sStatusPrefix.StartsWith("*") ? "." : "*";
     }
 
+
   public:
     MODCONSTRUCTOR(CCryptMod) {
         AddHelpCommand();
@@ -79,9 +213,17 @@ class CCryptMod : public CModule {
         AddCommand("ListKeys", static_cast<CModCommand::ModCmdFunc>(
                                    &CCryptMod::OnListKeysCommand),
                    "", "List all keys");
+        AddCommand("KeyX", static_cast<CModCommand::ModCmdFunc>(
+                                 &CCryptMod::OnKeyXCommand),
+                   "<Nick>", "Start a DH1080 key exchange with nick");
     }
 
-    ~CCryptMod() override {}
+    ~CCryptMod() override {
+        if (dh) {
+            DH_free(dh);
+            dh = NULL;
+        }
+    }
 
     bool OnLoad(const CString& sArgsi, CString& sMessage) override {
         MCString::iterator it = FindNV(NICK_PREFIX_KEY);
@@ -227,6 +369,41 @@ class CCryptMod : public CModule {
     }
 
     EModRet OnPrivNotice(CNick& Nick, CString& sMessage) override {
+        CString sCommand = sMessage.Token(0);
+        CString sOtherPubKey = sMessage.Token(1);
+
+        if ((sCommand.Equals("DH1080_INIT") || sCommand.Equals("DH1080_INIT_CBC")) && !sOtherPubKey.empty()) {
+            CString sSecretKey;
+            CString sTail = sMessage.Token(2); /* For fish10 */
+
+            /* remove trailing A */
+            if (sOtherPubKey.TrimSuffix("A") && DH1080_gen() && DH1080_comp(sOtherPubKey, sSecretKey)) {
+                PutModule("Received DH1080 public key from " + Nick.GetNick() + ", sending mine...");
+                PutIRC("NOTICE " + Nick.GetNick() + " :DH1080_FINISH " + sPubKey + "A" + (sTail.empty()?"":(" " + sTail)));
+                SetNV(Nick.GetNick().AsLower(), sSecretKey);
+                PutModule("Key for " + Nick.GetNick() + " successfully set.");
+                return HALT;
+            }
+            PutModule("Error in " + sCommand + " with " + Nick.GetNick() + ": " + (sSecretKey.empty()?"no secret key computed":sSecretKey));
+            return CONTINUE;
+
+        } else if (sCommand.Equals("DH1080_FINISH") && !sOtherPubKey.empty()) {
+            /*
+             * In theory we could get a DH1080_FINISH without us having sent a DH1080_INIT first,
+             * but then to have any use for the other user, they'd already have our pub key
+             */
+            CString sSecretKey;
+
+            /* remove trailing A */
+            if (sOtherPubKey.TrimSuffix("A") && DH1080_gen() && DH1080_comp(sOtherPubKey, sSecretKey)) {
+                SetNV(Nick.GetNick().AsLower(), sSecretKey);
+                PutModule("Key for " + Nick.GetNick() + " successfully set.");
+                return HALT;
+            }
+            PutModule("Error in " + sCommand + " with " + Nick.GetNick() + ": " + (sSecretKey.empty()?"no secret key computed":sSecretKey));
+            return CONTINUE;
+        }
+
         FilterIncoming(Nick.GetNick(), Nick, sMessage);
         return CONTINUE;
     }
@@ -320,6 +497,21 @@ class CCryptMod : public CModule {
                       "]");
         } else {
             PutModule("Usage: SetKey <#chan|Nick> <Key>");
+        }
+    }
+
+    void OnKeyXCommand(const CString& sCommand) {
+        CString sTarget = sCommand.Token(1);
+
+        if (!sTarget.empty()) {
+            if (DH1080_gen()) {
+                PutIRC("NOTICE " + sTarget + " :DH1080_INIT " + sPubKey + "A");
+                PutModule("Sent my DH1080 public key to " + sTarget + ", waiting for reply ...");
+            } else {
+                PutModule("Error generating our keys, nothing sent.");
+            }
+        } else {
+            PutModule("Usage: KeyX <Nick>");
         }
     }
 
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -99,7 +99,6 @@ class CCryptMod : public CModule {
         /* Generate our keys on first call */
         if (!dh) {
             int len;
-            unsigned char *buf;
             const BIGNUM *bPrivKey = NULL;
             const BIGNUM *bPubKey = NULL;
             BIGNUM *bPrime = NULL;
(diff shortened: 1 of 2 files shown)


modules/crypt.cpp
```diff
@@ -43,16 +42,151 @@
...
+        if (!dh) {
+            int len;
+            unsigned char *buf;
+            const BIGNUM *bPrivKey = NULL;
+            const BIGNUM *bPubKey = NULL;
+            BIGNUM *bPrime = NULL;
+            BIGNUM *bGen = NULL;
+            dh = DH_new();
+
+            if (!BN_hex2bn(&bPrime, prime1080) || !BN_dec2bn(&bGen, "2") || !DH_set0_pqg(dh, bPrime, NULL, bGen) || !DH_generate_key(dh)) {
```
psychon: `DH_generate_key` will freeze ZNC, won't it?

modules/crypt.cpp
```diff
@@ -43,16 +42,151 @@
...
+             */
+            sPrivKey = CString::basic_string(buf, &buf[len]);
+            sPrivKey.Base64Encode();
+
+            /* Get our public key */
+            len = BN_num_bytes(bPubKey);
+            buf = (unsigned char *)realloc(buf, len);
+            BN_bn2bin(bPubKey, buf);
+            sPubKey = CString::basic_string(buf, &buf[len]);
+            sPubKey.Base64Encode();
```
psychon: Not that it matters much, but you could just `string.resize(len)` a (previously empty) string to the right size and then call `BN_bn2bin(foo, string.c_str())` to directly write the binary representation there. That would save a copy (not important and all) and would avoid manual memory management.

modules/crypt.cpp
```diff
@@ -43,16 +42,151 @@
...
+        BIGNUM *bOtherPubKey = NULL;
+
+        /* Prepare other public key */
+        len = sOtherPubKey.Base64Decode();
+        bOtherPubKey = BN_bin2bn((unsigned char *)sOtherPubKey.data(), len, NULL);
+
+        /* Generate secret key */
+        key = (unsigned char *)calloc(DH_size(dh), 1);
+        if ((len = DH_compute_key(key, bOtherPubKey, dh)) == -1) {
+            sSecretKey = "";
```
psychon: ...why?

modules/crypt.cpp
```diff
@@ -65,9 +199,17 @@ class CCryptMod : public CModule {
...
+        AddCommand("KeyX", static_cast<CModCommand::ModCmdFunc>(
+                                 &CCryptMod::OnKeyXCommand),
+                   "<Nick>", "Start a DH1080 key exchange with nick");
     }
 
-    ~CCryptMod() override {}
+    ~CCryptMod() override {
+        if (dh) {
+            DH_free(dh);
+            dh = NULL;
```
psychon: How about `std::unique_ptr` to avoid a lot of manual memory management?

modules/crypt.cpp
```diff
@@ -200,6 +342,41 @@ class CCryptMod : public CModule {
...
+
+        } else if (sCommand.Equals("DH1080_FINISH") && !sOtherPubKey.empty()) {
+            /*
+             * In theory we could get a DH1080_FINISH without us having sent a DH1080_INIT first,
+             * but then to have any use for the other user, they'd already have our pub key
+             */
+            CString sSecretKey;
+
+            /* remove trailing A */
+            if (sOtherPubKey.TrimSuffix("A") && DH1080_gen() && DH1080_comp(sOtherPubKey, sSecretKey)) {
```
psychon: The weird trailing A is... weird. So, this stuff is Base64 encoded, but trailing = are dropped and afterwards trialing A are dropped?

modules/crypt.cpp
```diff
@@ -43,16 +42,151 @@
...
+#include <openssl/dh.h>
+#include <openssl/bn.h>
+#include <znc/SHA256.h>
 
 #define REQUIRESSL 1
 #define NICK_PREFIX_KEY "[nick-prefix]"
 
 class CCryptMod : public CModule {
+  private:
+    const char *prime1080="FBE1022E23D213E8ACFA9AE8B9DFADA3EA6B7AC7A7B7E95AB5EB2DF858921FEADE95E6AC7BE7DE6ADBAB8A783E7AF7A7FA6A2B7BEB1E72EAE2B72F9FA2BFB2A2EFBEFAC868BADB3E828FA8BADFADA3E4CC1BE7E8AFE85E9698A783EB68FA07A77AB6AD7BEB618ACF9CA2897EB28A6189EFA07AB99A8A7FA9AE299EFA7BA66DEAFEFBEFBF0B7D8B";
```
psychon: Any pointers for where this prime comes from?

modules/crypt.cpp
Cizzle: Not that I have experienced, it doesn't generate a prime or anything... It's also only ran once, though could move all this stuff to the contructor?

modules/crypt.cpp
Cizzle: Yes, that's how this "standard" is setup... just following it so it works with other implementations.

modules/crypt.cpp
Cizzle: Used in all other DH1080 fish implementations I could find (in either hex, dec or base64 encoded form): [KVIrc](https://github.com/Cizzle/KVIrc/blob/master/src/modules/fish/libkvifish.cpp#L70), [fish10](https://github.com/flakes/mirc_fish_10/blob/master/fish_10/src/dh1080.cpp#L57), ZNC FiSH [1 L118](http://pastebin.com/K8zvJ89v) and [2 L114](http://slexy.org/view/s2poFq0BaF), [FiSH-irssi](https://github.com/falsovsky/FiSH-irssi/blob/master/src/DH1080.c#L15), [Quassel IRC](https://github.com/quassel/quassel/blob/master/src/core/cipher.cpp#L20), [XChat-FiSH-AES](https://github.com/emigonza/XChat-FiSH-AES/blob/master/XChat_AES.py#L272) and so on...
Also mentioned [here](http://blog.bjrn.se/2009/01/proposal-for-better-irc-encryption.html).

modules/crypt.cpp
Cizzle: Will try this out, sounds like a better idea.

modules/crypt.cpp
psychon: Sorry, you are right. `DH_generate_key` doesn't actually sound as expensive as I previously thought.

modules/crypt.cpp
psychon: I'm afraid to ask, but: Is there any kind of docs or specification for this?

Cizzle: It's of course based on DH key exchange but the oldest implementation I could find dates back to [April 2004](http://mircryption.sourceforge.net/Extras/McpsFishDH.zip) and contains the closest thing to technical notes and documents making me believe that is the starting point. See [here](http://web.archive.org/web/20040503044610/http://fish.sekure.us/) as well.

modules/crypt.cpp
DarthGandalf: Please add a comment in code

modules/crypt.cpp
DarthGandalf: +1

modules/crypt.cpp
```diff
@@ -43,16 +42,141 @@
...
```
DarthGandalf: `*` is near type, not near variable. Also spaces missing around `=`. Just use clang-format, there is its configuration in this repo.

modules/crypt.cpp
```diff
@@ -43,16 +42,141 @@
...
+    /* Generate our keys once and reuse, just like ssh keys */
+    DH *dh = NULL;
+    CString sPrivKey;
+    CString sPubKey;
+
+#if OPENSSL_VERSION_NUMBER < 0X10100000L
+    int DH_set0_pqg(DH *dh, BIGNUM *p, BIGNUM *q, BIGNUM *g) {
```
DarthGandalf: It doesn't use `this`? Make it `static`.

modules/crypt.cpp
```diff
@@ -43,16 +42,141 @@
...
+        /* If the fields p and g in dh are NULL, the corresponding input
+         * parameters MUST be non-NULL.  q may remain NULL.
+         */
+        if (dh == NULL || (dh->p == NULL && p == NULL) || (dh->g == NULL && g == NULL))
```
DarthGandalf: nullptr

modules/crypt.cpp
```diff
@@ -43,16 +42,141 @@
...
 #define REQUIRESSL 1
 #define NICK_PREFIX_KEY "[nick-prefix]"
 
 class CCryptMod : public CModule {
+  private:
+    const char *prime1080="FBE1022E23D213E8ACFA9AE8B9DFADA3EA6B7AC7A7B7E95AB5EB2DF858921FEADE95E6AC7BE7DE6ADBAB8A783E7AF7A7FA6A2B7BEB1E72EAE2B72F9FA2BFB2A2EFBEFAC868BADB3E828FA8BADFADA3E4CC1BE7E8AFE85E9698A783EB68FA07A77AB6AD7BEB618ACF9CA2897EB28A6189EFA07AB99A8A7FA9AE299EFA7BA66DEAFEFBEFBF0B7D8B";
+    /* Generate our keys once and reuse, just like ssh keys */
+    DH *dh = NULL;
+    CString sPrivKey;
+    CString sPubKey;
```
DarthGandalf: members start with m_

modules/crypt.cpp
```diff
@@ -65,9 +189,17 @@ class CCryptMod : public CModule {
         AddCommand("ListKeys", static_cast<CModCommand::ModCmdFunc>(
                                    &CCryptMod::OnListKeysCommand),
                    "", "List all keys");
+        AddCommand("KeyX", static_cast<CModCommand::ModCmdFunc>(
```
DarthGandalf: What does X mean?

psychon: Key eXchange, for some reason I feel like this abbreviation is common in crypto-related stuff, but I agree that it is not self-explanatory.

modules/crypt.cpp
```diff
@@ -200,6 +332,41 @@ class CCryptMod : public CModule {
     }
 
     EModRet OnPrivNotice(CNick& Nick, CString& sMessage) override {
+        CString sCommand = sMessage.Token(0);
+        CString sOtherPubKey = sMessage.Token(1);
+
+        if ((sCommand.Equals("DH1080_INIT") || sCommand.Equals("DH1080_INIT_CBC")) && !sOtherPubKey.empty()) {
```
psychon: Are you sure that both the normal INIT and the INIT_CBC should be handled the same? Whoever added (or removed) a CBC suffix must have had a reason for it (CBC: cypher block chaining, a not-really-secure but way too common scheme)

psychon:
> I don't know much crypto. @psychon does crypto part look fine for you?

No. However, if this is some kind of "almost standard"... And it will be better than transfering a shared secret over IRC directly.

All that this code does is add a possibility for a diffie-hellman(-merkle) key exchange. That crypto primitive is good (however: unauthenticated! A man-in-the-middle can do whatever he wants). I am not sure if the hardcoded prime is a good one (if I read this correctly, it has 1080 bits? That might be a bit few these days), but whatever.

All together: The existing crypto in here is bad enough already that I am not really worried here.

psychon:
In TLS-speak: This might be DHE_1080_Blowfish_CBC (I made up that acronym).
TLS wouldn't pick a 1080 bit prime (well, it could only pick a 1024 one, I think). That's quite few bits and not really recommended. TLS wouldn't pick Blowfish. AES is prefered, I think. CBC is also not recommended.

modules/crypt.cpp
Cizzle: I've chosen to stick with what's being used in all other implementations.

modules/crypt.cpp
Cizzle: History is like this:
- first there was DH1080_INIT and the keys generated with this were used for blowfish in ECB mode.
- as this ECB was deemed to insecure (always the same encoded word for the same unencoded word) a transition was made to use CBC.
- Mircryption started using this with DH1080_INIT_CBC.
- fish10 (and/or others) kept using DH1080_INIT but added a " CBC" suffix after the pubkey.
- KVIrc (and/or others) use CBC by default for all.

So DH1080_INIT and DH1080_INIT_CBC (and "DH1080_INIT pubkey CBC") all use the same method, but just differ in setting the secretkey: for CBC they mostly get a "cbc:" prefix, ECB gets nothing or "old:" or "ecb:". Then the client knows what decryption and encryption scheme to use, though most have fallback to either one depending on the received input (ECB encoded messages start with "+OK ", CBC with "+OK *").

As ECB is more insecure than CBC and thus to push CBC usage, I've chosen to use CBC by default even for the original DH1080_INIT, also because the crypt module only supports CBC mode.

Cizzle:
> I am not sure if the hardcoded prime is a good one

Can be checked with [DH_check()](https://www.openssl.org/docs/man1.1.0/crypto/DH_generate_parameters.html).

modules/crypt.cpp
psychon: Thanks for this information :-)

psychon:
>>  I am not sure if the hardcoded prime is a good one

>Can be checked with DH_check().

That's more meant towards "only 1080(?)" bits. Anyway, the general idea here is fine with me and I leave reviewing the implementation to @DarthGandalf

modules/crypt.cpp
Cizzle: But does std::unique_ptr know it should use DH_new() to create a new DH object and DH_free() to delete it?

psychon: I didn't actually test this, but I imagine something like this:
```c++
std::unique_ptr<DH, decltype(&DH_free)> m_dh;
```
This would then require the constructor of CCryptMod to do something like `m_dh(nullptr, DH_free)` to initalize this object. To construct a new DH instance, use `m_dh.reset(DH_new())`. In other places, it should just be possible to use `m_dh` as if it had type `DH*`.

@DarthGandalf Please correct me if I am wrong (I don't think I ever used this before).

psychon: Actually... I do not think you have to do anything in `CCryptMod`'s constructor. Instead, just use the following for assignment: `m_dh = std::unique_ptr<DH, void(*)(DH*)>(DH_new(), DH_free);`

psychon: Google found: http://stackoverflow.com/a/19054467/436275

DarthGandalf:
Do you have any idea how to add a test to it? https://github.com/znc/znc/blob/master/test/integration/main.cpp tests some modules, but crypt would require reimplementing the same logic in test or connecting a different client which already implements it

Cizzle:
Have the test start two znc servers and have two clients connect, then have one client keyx to the other and check the secret key set for both clients? Could parse the notices sent too for validity?

modules/crypt.cpp
```diff
@@ -43,16 +42,145 @@
...
+
+            if (!BN_hex2bn(&bPrime, prime1080) || !BN_dec2bn(&bGen, "2") || !DH_set0_pqg(m_pDH, bPrime, nullptr, bGen) || !DH_generate_key(m_pDH)) {
+                /* one of them failed */
+                if (bPrime)
+                    BN_clear_free(bPrime);
+                if (bGen)
+                    BN_clear_free(bGen);
+                if (m_pDH) {
+                    DH_free(m_pDH);
+                    m_pDH = nullptr;
```
Cizzle: Will std::unique_ptr handle this correctly? m_pDH needs to be put back to nullptr if any of those commands failed.

DarthGandalf:
Yeah, that might work, I didn't think about starting 2 ZNCs. May you add such test?

DarthGandalf:
Sure, I don't see why not.

2017-02-17 0:06 GMT+00:00 Cizzle <notifications@github.com>:

Cizzle:
If someone could take a look at the test and why the module's OnPrivNotice isn't activated for the primary user upon receiving the DH1080_FINISH? Have tested it with just about any NOTICE and extra PutModule lines before any actual checking in the OnPrivNotice override.

psychon:
@DarthGandalf What do you think about something like the following so people who run into this spent less time to learn that ZNC should be installed before running these tests?
```diff
diff --git a/test/integration/main.cpp b/test/integration/main.cpp
index 4a725a7..d453aaf 100644
--- a/test/integration/main.cpp
+++ b/test/integration/main.cpp
@@ -141,6 +141,9 @@ class Process : public IO<QProcess> {
         m_proc.setProcessEnvironment(env);
         setup(&m_proc);
         m_proc.start(cmd, args);
+        [this]() {
+            ASSERT_TRUE(m_proc.waitForStarted());
+        }();
     }
     ~Process() override {
         if (m_kill) m_proc.terminate();
```

psychon:
By the way, Travis fails also due to (but I don't remember which Travis build I copied this from):
```
/home/travis/build/znc/znc/test/integration/main.cpp:88:58: error: non-const

      lvalue reference to type 'QByteArray' cannot bind to a value of unrelated

      type 'const char [1]'

    void ReadUntilAndGet(QByteArray pattern, QByteArray& match = "") {

                                                         ^       ~~

/home/travis/build/znc/znc/test/integration/main.cpp:88:58: note: passing

      argument to parameter 'match' here
```

psychon:
And another @DarthGandalf 
```diff
diff --git a/test/integration/main.cpp b/test/integration/main.cpp
index 1fe3a7d..0c63bdb 100644
--- a/test/integration/main.cpp
+++ b/test/integration/main.cpp
@@ -456,6 +456,8 @@ TEST_F(ZNCTest, HTTP) {
     EXPECT_THAT(reply->rawHeader("Server").toStdString(), HasSubstr("ZNC"));
 }
 
+// This test takes way too long...
+#if 0
 TEST_F(ZNCTest, FixCVE20149403) {
     auto znc = Run();
     Z;
@@ -492,6 +494,7 @@ TEST_F(ZNCTest, FixCVE20149403) {
                     .toStdString(),
                 HasSubstr("Channel [#znc] already exists"));
 }
+#endif
 
 TEST_F(ZNCTest, FixFixOfCVE20149403) {
     auto znc = Run();
@@ -819,7 +822,7 @@ TEST_F(ZNCTest, ControlpanelModule) {
     Z;
 
     client.Write(request + "ListCTCPs");
-    client.ReadUntil(response + "Request: VERSION");
+    client.ReadUntil(response + "| VERSION | Test  |");
     Z;
 
     client.Write(request + "ListMods");
@@ -831,11 +834,11 @@ TEST_F(ZNCTest, ControlpanelModule) {
     Z;
 
     client.Write(request + "ListNetworks");
-    client.ReadUntil(response + "Network: test");
+    client.ReadUntil(response + "| test ");
     Z;
 
     client.Write(request + "ListUsers");
-    client.ReadUntil(response + "Username: user");
+    client.ReadUntil(response + "| user ");
     Z;
 
     client.Write(request + "LoadModule");
@@ -1406,7 +1409,7 @@ TEST_F(ZNCTest, ControlpanelModule) {
     Z;
 
     client.Write(request + "ListCTCPs KindOne");
-    client.ReadUntil(response + "Request: VERSION");
+    client.ReadUntil(response + "| VERSION | Test");
     Z;
 
     client.Write(request + "ListMods KindOne");
@@ -1424,7 +1427,7 @@ TEST_F(ZNCTest, ControlpanelModule) {
     Z;
 
     client.Write(request + "ListMods KindOne");
-    client.ReadUntil(response + "Name: perform");
+    client.ReadUntil(response + "| perform |");
     Z;
 
     client.Write(request + "ListNetMods KindOne");
@@ -1432,11 +1435,11 @@ TEST_F(ZNCTest, ControlpanelModule) {
     Z;
 
     client.Write(request + "ListNetworks KindOne");
-    client.ReadUntil(response + "Network: freenode");
+    client.ReadUntil(response + "| freenode | No ");
     Z;
 
     client.Write(request + "ListUsers");
-    client.ReadUntil(response + "Username: user");
+    client.ReadUntil(response + "| user ");
     Z;
 
     client.Write(request + "LoadModule");
```

psychon:
@Cizzle ZNC does not recognize the NOTICE as a private notice, because of a kind of desync: You expect ZNC to think its nick is "nick", but ZNC thinks its nick is "user". Possible patch (also "fixes" the rest of the test):
```diff
diff --git a/test/integration/main.cpp b/test/integration/main.cpp
index 99f106b..9897dbe 100644
--- a/test/integration/main.cpp
+++ b/test/integration/main.cpp
@@ -2079,6 +2079,7 @@ TEST_F(ZNCTest, ModuleCrypt) {
 
     QByteArray pub2("");
     ircd2.ReadUntilAndGet("NOTICE nick :DH1080_FINISH ", pub2);
+    pub2.replace(" nick ", " user ");
     ircd1.Write(":nick2!user2@user2/test " + pub2);
     Z;
 
@@ -2086,7 +2087,16 @@ TEST_F(ZNCTest, ModuleCrypt) {
     Z;
 
     client1.Write("PRIVMSG *crypt :listkeys");
-    client1.ReadUntil("Target: nick2");
+    /* FIXME: This part does not work with the "new old" table layout. This
+     * needs to parse the following correctly:
+[2017-02-28 09:58:50.026045] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :+---------------+---------------------------------------------+]
+[2017-02-28 09:58:50.026074] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :| Target        | Key                                         |]
+[2017-02-28 09:58:50.026110] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :+---------------+---------------------------------------------+]
+[2017-02-28 09:58:50.026130] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :| nick2         | Z3dFNiK8ojJr77x4y6TIcctpamhY75hHCZMVg1w7sX0 |]
+[2017-02-28 09:58:50.026150] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :| @nick-prefix@ | .                                           |]
+[2017-02-28 09:58:50.026171] (user/test) ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :+---------------+---------------------------------------------+]
+
+    client1.ReadUntil("| nick2 ");
     Z;
     QByteArray key1("");
     client1.ReadUntilAndGet("Key: ", key1);
@@ -2098,6 +2108,7 @@ TEST_F(ZNCTest, ModuleCrypt) {
     client2.ReadUntilAndGet("Key: ", key2);
     Z;
     ASSERT_EQ(key1, key2);
+*/
 }
 
 }  // namespace
```

DarthGandalf:
@psychon I have a local patch which partially fixes the test (basically removes the controlpanel test), I've been trying to fix another test (building modules) before pushing it.
You can push both your patches. I don't know what to do with FixCVE20149403.

Cizzle:
> By the way, Travis fails also due to (but I don't remember which Travis build I copied this from):

Would be my fault, though compiling didn't throw any warnings? Use QByteArray("") then?

> ZNC does not recognize the NOTICE as a private notice, because of a kind of desync

Not better to sync it then if this is possible? And how come ZNC does send a PRIVMSG to the client's nick "nick" after keyx?
> ZNC -> CLI [:*crypt!znc@znc.in PRIVMSG nick :Sent my DH1080 public key to nick2, waiting for reply ...]

psychon:
> You can push both your patches.

I'll add it to my to do list (might take a while). Also, I prefer going through a PR instead of pushing this directly.

> I don't know what to do with FixCVE20149403.

This test is slow because it waits until ZNC joins the channel after it was added. This joining is triggered by `CIRCNetworkJoinTimer` after 30 seconds. However, I do not see why it waits for the JOIN. That does not cause any internal state to change, does it? And if I understand/remember the original issue correctly (adding "znc" through webadmin while "#znc" already exists), the following would still trigger the original issue, wouldn't it?
```diff
diff --git a/test/integration/main.cpp b/test/integration/main.cpp
index de8e693..81a0dcd 100644
--- a/test/integration/main.cpp
+++ b/test/integration/main.cpp
@@ -481,8 +481,6 @@ TEST_F(ZNCTest, FixCVE20149403) {
                           {"name", "znc"},
                           {"enabled", "1"},
                       });
-    ircd.ReadUntil("JOIN #znc");
-    Z;
     EXPECT_THAT(HttpPost(request,
                          {
                              {"user", "user"},
```
Now... sorry for the noise and let's get back to this pull request.

PR #1380, opened 2017-02-26
Adding some comments to block module

Phansa:
I noticed a lot of the modules don't have comments/it's hard to figure out what is going on so I thought it would be a good idea to add some comments to help other people trying to develop modules understand what is going on.

### DIFF
--- a/modules/blockuser.cpp
+++ b/modules/blockuser.cpp
@@ -24,6 +24,7 @@ using std::vector;
 class CBlockUser : public CModule {
   public:
     MODCONSTRUCTOR(CBlockUser) {
+        //Creating all commands
         AddHelpCommand();
         AddCommand("List", static_cast<CModCommand::ModCmdFunc>(
                                &CBlockUser::OnListCommand),
@@ -62,7 +63,7 @@ class CBlockUser : public CModule {
         return true;
     }
 
-    /*If user is the on the blocked list tries to log in, displays - MESSAGE 
+    /*If user is the on the blocked list and tries to log in, displays - MESSAGE 
     and stops their log in attempt.*/
     EModRet OnLoginAttempt(std::shared_ptr<CAuthBase> Auth) override {
         if (IsBlocked(Auth->GetUsername())) {
@@ -96,7 +97,8 @@ class CBlockUser : public CModule {
         if (PutModule(Table) == 0) PutModule("No users blocked");
     }
 
-    //Blocks a user if possible(not self, not already blocked)
+    /*Blocks a user if possible(aka not self, not already blocked).
+    Displays an error message if not possible. */
     void OnBlockCommand(const CString& sCommand) {
         CString sUser = sCommand.Token(1, true);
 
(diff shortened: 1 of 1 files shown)


modules/blockuser.cpp
```diff
@@ -178,20 +188,21 @@ class CBlockUser : public CModule {
...
-        // Disconnect all clients
+        //Disconnect all clients
         vector<CClient*> vpClients = pUser->GetAllClients();
         vector<CClient*>::iterator it;
         for (it = vpClients.begin(); it != vpClients.end(); ++it) {
             (*it)->PutStatusNotice(MESSAGE);
             (*it)->Close(Csock::CLT_AFTERWRITE);
         }
 
-        // Disconnect all networks from irc
```
DarthGandalf: Why did you remove space?

modules/blockuser.cpp
```diff
@@ -126,6 +133,7 @@ class CBlockUser : public CModule {
             PutModule("This user is not blocked");
     }
 
+    //Compressing all above functions into one for embedded web requests.
```
DarthGandalf: It doesn't compress anything...

modules/blockuser.cpp
```diff
@@ -24,6 +24,7 @@ using std::vector;
 class CBlockUser : public CModule {
   public:
     MODCONSTRUCTOR(CBlockUser) {
+        //Creating all commands
```
DarthGandalf: I don't see how this comment helps. Name of `AddCommand` function already contains what it does, no?

modules/blockuser.cpp
```diff
@@ -79,6 +82,7 @@ class CBlockUser : public CModule {
         }
     }
 
+    //Displays all blocked users as a list.
```
DarthGandalf: There is already "List blocked users" in description of this command, above

modules/blockuser.cpp
Phansa: Yes I agree/will take it out. Also would this be worthwhile to work on? I think I should have created an issue first, sorry about this.

modules/blockuser.cpp
Phansa: What i'm trying to say is this that this gives the user the same functionality as the other commands, not sure how to word it or if it is necessary.

DarthGandalf:
Re spaces, there is a clang-format configuration in this repo, so you can run clang-format to automatically fix some things.

modules/blockuser.cpp
```diff
@@ -62,6 +62,8 @@ class CBlockUser : public CModule {
         return true;
     }
 
+    /* If user is the on the blocked list and tries to log in, displays - MESSAGE
```
DarthGandalf: the on?

modules/blockuser.cpp
```diff
@@ -168,6 +175,8 @@ class CBlockUser : public CModule {
     }
 
   private:
+    /* Iterates through all blocked users and returns true if they 
+    are blocked, else false*/
```
DarthGandalf: This sounds like "if all blocked users are blocked".

modules/blockuser.cpp
Phansa: Was a typo/I fixed it.

PR #1382, opened 2017-03-09
Add NickPrefix command to view and set the nickprefix.

Cizzle:
Hides the internal keyword from ListKeys and adds the NickPrefix comand to view and optionally set another prefix.

### DIFF
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -525,23 +525,23 @@ class CCryptMod : public CModule {
     }
 
     void OnListKeysCommand(const CString& sCommand) {
-        if (BeginNV() == EndNV() || BeginNV()->first.Equals(NICK_PREFIX_KEY)) {
-            PutModule("You have no encryption keys set.");
-        } else {
-            CTable Table;
-            Table.AddColumn("Target");
-            Table.AddColumn("Key");
+        CTable Table;
+        CString::size_type rows = 0;
+        Table.AddColumn("Target");
+        Table.AddColumn("Key");
 
-            for (MCString::iterator it = BeginNV(); it != EndNV(); ++it) {
-                if (!it->first.Equals(NICK_PREFIX_KEY)) {
-                    Table.AddRow();
-                    Table.SetCell("Target", it->first);
-                    Table.SetCell("Key", it->second);
-                }
+        for (MCString::iterator it = BeginNV(); it != EndNV(); ++it) {
+            if (!it->first.Equals(NICK_PREFIX_KEY)) {
+                /* Index starts at 0, we want amount of rows */
+                rows = Table.AddRow() + 1;
+                Table.SetCell("Target", it->first);
+                Table.SetCell("Key", it->second);
             }
-
-            PutModule(Table);
         }
+        if (rows == 0)
+            PutModule("You have no encryption keys set.");
+        else
+            PutModule(Table);
     }
 
     CString MakeIvec() {
--- a/modules/crypt.cpp
+++ b/modules/crypt.cpp
@@ -510,7 +510,6 @@ class CCryptMod : public CModule {
         if (sPrefix.empty()) {
             PutModule("Nick Prefix: " + NickPrefix());
         } else {
-            CString sTail = sCommand.Token(2, true);
             CString sStatusPrefix = GetUser()->GetStatusPrefix();
             size_t sp = sStatusPrefix.size();
             size_t np = sPrefix.size();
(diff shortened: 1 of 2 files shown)


modules/crypt.cpp
```diff
@@ -501,25 +504,36 @@ class CCryptMod : public CModule {
         }
     }
 
+    void OnNickPrefixCommand(const CString& sCommand) {
+        CString sPrefix = sCommand.Token(1, true);
```
DarthGandalf: What does a prefix with spaces inside mean?

Cizzle: Hm, thought space could be used but seems they can't, messages from remote don't arrive at the client.

modules/crypt.cpp
```diff
@@ -501,25 +504,36 @@ class CCryptMod : public CModule {
...
+        if (sPrefix.empty()) {
+            PutModule("Nick Prefix: " + NickPrefix());
+        } else {
+            CString sStatusPrefix = GetUser()->GetStatusPrefix();
+            size_t sp = sStatusPrefix.size();
+            size_t np = sPrefix.size();
+            SetNV(NICK_PREFIX_KEY, sPrefix);
+            PutModule("Setting Nick Prefix to " + sPrefix);
+            if (sStatusPrefix.CaseCmp(sPrefix, std::min(sp, np)) == 0)
+                PutModule("WARNING: overlap with Status Prefix (" + sStatusPrefix + "), this Nick Prefix will not be used!");
```
Cizzle: Either "*" or ".", see NickPrefix(). Thought I'd allow setting it in case one changes it's status prefix instead afterwards.

modules/crypt.cpp
```diff
@@ -501,25 +504,36 @@ class CCryptMod : public CModule {
...
+        }
+    }
+
     void OnListKeysCommand(const CString& sCommand) {
-        if (BeginNV() == EndNV()) {
+        if (BeginNV() == EndNV() || BeginNV()->first.Equals(NICK_PREFIX_KEY)) {
```
Cizzle: NICK_PREFIX_KEY is actually also a "key", so we want to ignore it. If there are no other keys, NICK_PREFIX_KEY will be the first entry. Unless I misunderstood your comment.

modules/crypt.cpp
DarthGandalf: Are you talking to me or to the user who tried to set such prefix?
*I* know about that function. But the current behavior would be surprising, and the message which user sees is misleading.

modules/crypt.cpp
DarthGandalf: Why do you think it's the first entry? Maybe it's the last entry? Or somewhere in the middle?

modules/crypt.cpp
```diff
@@ -505,16 +505,20 @@ class CCryptMod : public CModule {
...
         if (sPrefix.empty()) {
             PutModule("Nick Prefix: " + NickPrefix());
         } else {
+            CString sTail = sCommand.Token(2, true);
             CString sStatusPrefix = GetUser()->GetStatusPrefix();
             size_t sp = sStatusPrefix.size();
             size_t np = sPrefix.size();
             SetNV(NICK_PREFIX_KEY, sPrefix);
-            PutModule("Setting Nick Prefix to " + sPrefix);
+            if (sTail.empty())
```
DarthGandalf: I'd be fine with just ignoring anything after the first parameter. It's obvious from the message anyway.

modules/crypt.cpp
```diff
@@ -501,29 +504,45 @@ class CCryptMod : public CModule {
...
-    void OnListKeysCommand(const CString& sCommand) {
-        if (BeginNV() == EndNV()) {
-            PutModule("You have no encryption keys set.");
+    void OnNickPrefixCommand(const CString& sCommand) {
+        CString sPrefix = sCommand.Token(1);
+
+        if (sPrefix.empty()) {
+            PutModule("Nick Prefix: " + NickPrefix());
+        } else if (sPrefix.size() > 1 && sPrefix.StartsWith(":")) {
+            PutModule("You cannot use : followed by other symbols as Nick Prefix.");
```
DarthGandalf: Let's just disable anything starting with `:`, even if it's `:abc`

modules/crypt.cpp
```diff
@@ -504,22 +508,28 @@ class CCryptMod : public CModule {
...
+    void OnGetNickPrefixCommand(const CString& sCommand) {
+        CString sPrefix = NickPrefix();
+        PutModule("Nick Prefix" + (sPrefix.empty() ? " disabled." : (": " + sPrefix)));
+    }
+
+    void OnSetNickPrefixCommand(const CString& sCommand) {
         CString sPrefix = sCommand.Token(1);
 
-        if (sPrefix.empty()) {
+/*        if (sPrefix.empty()) {
```
DarthGandalf: remove...

DarthGandalf:
The test fails

PR #1383, opened 2017-03-12
Adding comments to modules, updating dates to 2017

Phansa:
I have added some comments to the sample, lastseen and stickychan modules. I have updated the copyright dates on many files from 2016 to 2017 and added liscene information to some files that were missing it.

### DIFF
--- a/modules/lastseen.cpp
+++ b/modules/lastseen.cpp
@@ -96,7 +96,6 @@ class CLastSeenMod : public CModule {
     bool WebRequiresAdmin() override { return true; }
     CString GetWebMenuTitle() override { return "Last Seen"; }
 
-    // Provides GUI to configure this module by adding a widget to user page in webadmin.
     bool OnWebRequest(CWebSock& WebSock, const CString& sPageName,
                       CTemplate& Tmpl) override {
         if (sPageName == "index") {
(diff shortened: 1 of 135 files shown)


modules/lastseen.cpp
```diff
@@ -128,6 +130,7 @@ class CLastSeenMod : public CModule {
         return false;
     }
 
+    // Provides GUI to configure this module by adding a widget to user page in webadmin.
```
DarthGandalf: No, it doesn't.
What would be such configuration?

Phansa: I wasn't sure about this one so I will comment it out. Should I seperate one commit of updating all the dates and one of adding comments to modules?

DarthGandalf: Such separation is fine.

Phansa: Should I fix the commit messages and create another pull request/close this one?

DarthGandalf: Just force-push to this one.

DarthGandalf:
temp and temp2?..

Phansa:
The temp/temp2 was just when I was fixing the commits. It should all be correct now with these last 3 commits.

modules/lastseen.cpp
```diff
@@ -95,6 +96,7 @@ class CLastSeenMod : public CModule {
     bool WebRequiresAdmin() override { return true; }
     CString GetWebMenuTitle() override { return "Last Seen"; }
 
+    // Provides GUI to configure this module by adding a widget to user page in webadmin.
```
DarthGandalf: There is no configuration for this module.

PR #1386, opened 2017-03-16
Refactor duplicate code in crypt module.

Cizzle:
Moves code from OnUserMsg, OnUserNotice and OnUserAction to one function.


DarthGandalf:
Thanks.
This is not what I meant - with CMessage API you wouldn't need so many parameters to FilterOutgoing function, but it's still much better than the duplicated code before.

PR #1387, opened 2017-03-19
Off-the-Record Messaging plugin

mmilata:
Hi, some time ago I wrote [OTR encryption plugin](http://wiki.znc.in/Otr) for ZNC and am wondering if you'd like to include it upstream. If yes is there anything that needs to be changed?


DarthGandalf:
> If yes is there anything that needs to be changed?

You can start with unbreaking the build :)
Then add a test which works on travis/etc, see https://github.com/znc/znc/tree/master/test/integration
I didn't look at the code yet.

mmilata:
Hmm, both appveyor failures are python not working on i386, which might not be related to this plugin as it doesn't use python:
```
Fatal Python error: Could not allocate TLS entry
```

Looking into writing tests now.

DarthGandalf:
Yep, https://cygwin.com/ml/cygwin/2017-03/msg00281.html reports the same issue about python on 32 bit cygwin.

mmilata:
Added a basic test. The [ASAN build](https://travis-ci.org/znc/znc/jobs/222434362) fails due to a memory leak. Unfortunately the LeakSanitizer output is not very informative and I haven't been able to produce one that has function names instead of addresses.

Any idea how to debug this?

psychon:
> Any idea how to debug this?

Random guess: It tries to look up symbols, but at this time all the needed shared objects are already unloaded. So a random idea would be to remove all `dlclose()` calls and see what happens.

test/CMakeLists.txt
```diff
@@ -83,7 +83,8 @@ externalproject_add(inttest_bin
 	"-DGTEST_ROOT:path=${GTEST_ROOT}"
 	"-DGMOCK_ROOT:path=${GMOCK_ROOT}"
 	"-DZNC_BIN_DIR:path=${CMAKE_INSTALL_FULL_BINDIR}"
-	"-DQt5_HINTS:path=${brew_qt5}")
+	"-DQt5_HINTS:path=${brew_qt5}"
+	"-DWITH_OTR:bool=${OTR_FOUND}")
```
DarthGandalf: For the test, we can assume that all dependencies are available.

test/integration/main.cpp
```diff
@@ -2094,4 +2103,197 @@ TEST_F(ZNCTest, ModuleCrypt) {
...
+    Z;
+    ircd1.RelayPrivMsg(ircd2, ":user!user@user/test ", IsOtrCiphertext);
+    Z;
+    ircd2.RelayPrivMsg(ircd1, ":nick2!user2@user2/test ", IsOtrCiphertext);
+    Z;
+    ircd2.RelayPrivMsg(ircd1, ":nick2!user2@user2/test ", IsOtrCiphertext);
+    Z;
+    ircd1.RelayPrivMsg(ircd2, ":user!user@user/test ", IsOtrCiphertext);
+    Z;
+    ircd1.RelayPrivMsg(ircd2, ":user!user@user/test ", IsOtrCiphertext);
```
DarthGandalf: This sequence of relays looks magical...

test/integration/main.cpp
```diff
@@ -2094,4 +2103,197 @@ TEST_F(ZNCTest, ModuleCrypt) {
...
+    Z;
+    for (int i = 0; i < 8; i++) {
+        ircd2.RelayPrivMsg(ircd1, ":nick2!user2@user2/test ", IsOtrCiphertext);
+        Z;
+    }
+    for (int i = 0; i < 6; i++) {
+        ircd1.RelayPrivMsg(ircd2, ":user!user@user/test ", IsOtrCiphertext);
+        Z;
+    }
+    for (int i = 0; i < 3; i++) {
```
DarthGandalf: 5, 8, 6, and 3?

psychon:
I tried that "remove `dlclose()`" hack: https://travis-ci.org/psychon/znc/builds/222516902
The only symbol name that showed up was `otrl_context_priv_new`. Looking a bit at libotr's source code, I ended up with: https://github.com/psychon/znc/commit/123d5ed741ea003f326f3d277c7d890847ae8f34 https://travis-ci.org/psychon/znc/builds/222519219
However, this did not help. No idea why the result of `otrl_context_priv_new` is leaked.

This is an attempt at getting better backtraces: https://github.com/psychon/znc/commit/18ae57baac4c911e3a19b28fbd5468b7adb967f1
but that did not work either: https://travis-ci.org/psychon/znc/builds/222521761

DarthGandalf:
A random idea: maybe output can be better, if to upgrade Clang from 3.5.0 to something newer.

test/CMakeLists.txt
mmilata: That will make the test fail on cygwin/appveyor where non-ancient version of libotr is not available.

test/integration/main.cpp
mmilata: Yeah ... couldn't think of less magical way to let the two znc instances exchange messages until the handshake is done.

test/integration/main.cpp
mmilata: That's the number of lines/messages sent in each direction during the shared authentication. Same problem as above.

mmilata:
Tried with Clang 3.8 but the output seems to be the same.

mmilata:
@psychon hmm ... looking at libotr source, it seems that the memory allocated in `otrl_context_priv_new` is never freed. That function is only called in `new_context` where the result is assigned to member `context_priv` which is passed to only one function, `otrl_context_priv_force_finished`, which doesn't free the memory. So I'd say that it's libotr bug.

DarthGandalf:
Can we fix the library and use the fixed version in CI?

mmilata:
Guess I can give it a shot. Also looks like the leak is [reported already](https://bugs.otr.im/lib/libotr/issues/90).

psychon:
https://bugs.otr.im/lib/libotr/commits/master + the activity on the report linked to above

Looks like libotr is dead upstream?

mmilata:
It's fairly feature-complete but yeah, there seems to be no recent active development ...

I'm having trouble making ASan work on my devel system but I've been able to use Valgrind, I assume there will be some overlap with the leaks found by ASan: http://b42.srck.net/znc-otr-leaks.txt

psychon:
I tried to find a OTR sample and then tried to fix its memory leaks. `otrl_init` is still leaky since there is no `otrl_fini` (or something like that), but since library initialisation happens only once, i can live with that (however, dunno if ASan will detect and complain about it). The first sample that I found was inside the OTR source `tests/regression/client/client --load-key regression/client/otr.key --max-msg 3` and so I used that. The test itself is leaky (and racy), but whatever. The patch below fixes all memory leaks that valgrind finds, except for ones caused by `otrl_init` and some weird ones in `gcry_randomize` that I do not understand.

I guess upstream does not have much interest in this and so I'll just leave this patch here (and point out that this is a derived work of libotr and so the GPL and the LGPL apply to the patch, i.e. it can be picked up and sent upstream).
```diff
diff --git a/src/context.c b/src/context.c
index 44d8b86..ca42971 100644
--- a/src/context.c
+++ b/src/context.c
@@ -511,10 +511,12 @@ int otrl_context_forget(ConnContext *context)
     free(context->accountname);
     free(context->protocol);
     free(context->smstate);
+    free(context->context_priv);
     context->username = NULL;
     context->accountname = NULL;
     context->protocol = NULL;
     context->smstate = NULL;
+    context->context_priv = NULL;
 
     /* Free the application data, if it exists */
     if (context->app_data && context->app_data_free) {
diff --git a/src/instag.c b/src/instag.c
index cccd94f..609bced 100644
--- a/src/instag.c
+++ b/src/instag.c
@@ -205,10 +205,7 @@ otrl_instag_t otrl_instag_get_new()
     otrl_instag_t result = 0;
 
     while(result < OTRL_MIN_VALID_INSTAG) {
-	otrl_instag_t * instag = (otrl_instag_t *)gcry_random_bytes(
-		sizeof(otrl_instag_t), GCRY_STRONG_RANDOM);
-	result = *instag;
-	gcry_free(instag);
+	gcry_randomize((unsigned char *) &result, sizeof(result), GCRY_STRONG_RANDOM);
     }
 
     return result;
diff --git a/tests/regression/client/client.c b/tests/regression/client/client.c
index e72b661..0afa905 100644
--- a/tests/regression/client/client.c
+++ b/tests/regression/client/client.c
@@ -76,6 +76,11 @@ static OtrlFragmentPolicy fragPolicy = OTRL_FRAGMENT_SEND_SKIP;
 static const char *protocol = "otr-test";
 static const char *alice_name = "alice";
 static const char *bob_name = "bob";
+/* Some counters to fix a race condition: A message could be sent, then we
+ * decide to quit and the in-flight message is leaked.
+ */
+static int messages_to_alice_inflight = 0;
+static int messages_to_bob_inflight = 0;
 
 static const char *unix_sock_bob_path = "/tmp/otr-test-bob.sock";
 static const char *unix_sock_alice_path = "/tmp/otr-test-alice.sock";
@@ -561,6 +566,7 @@ static int recv_otr_msg(int sock, const char *to, const char *from,
 		if (new_msg) {
 			OK(strncmp(omsg->plaintext, new_msg, omsg->plaintext_len) == 0,
 					"Message exchanged is valid");
+			otrl_message_free(new_msg);
 			update_msg_counter();
 		}
 	} else {
@@ -711,7 +717,7 @@ static void *alice_thread(void *data)
 			fd = ev[i].data.fd;
 			event = ev[i].events;
 
-			if (fd == quit_pipe[0]) {
+			if (fd == quit_pipe[0] && messages_to_alice_inflight == 0) {
 				/* Time to leave. */
 				goto end;
 			} else if (fd == alice_sock) {
@@ -722,6 +728,7 @@ static void *alice_thread(void *data)
 					struct sockaddr_un sun;
 
 					/* Connection from Bob, accept it so we can handle it. */
+					len = sizeof(sun);
 					sock_from_bob = accept(fd, (struct sockaddr *) &sun,
 							&len);
 					ret = add_sock_to_pollset(epfd, sock_from_bob,
@@ -738,6 +745,7 @@ static void *alice_thread(void *data)
 				} else if (event & EPOLLIN) {
 					(void) recv_otr_msg(sock_from_bob, alice_name, bob_name,
 							&oinfo);
+					messages_to_alice_inflight--;
 				}
 				continue;
 			} else {
@@ -821,6 +829,7 @@ static void *bob_thread(void *data)
 
 		/* No event thus timeout, send message to Alice. */
 		if (nb_fd == 0) {
+			messages_to_alice_inflight++;
 			(void) send_otr_msg(sock_to_alice, alice_name, bob_name, &oinfo,
 					NULL);
 			continue;
@@ -833,7 +842,7 @@ static void *bob_thread(void *data)
 			fd = ev[i].data.fd;
 			event = ev[i].events;
 
-			if (fd == quit_pipe[0]) {
+			if (fd == quit_pipe[0] && messages_to_bob_inflight == 0) {
 				/* Time to leave. */
 				goto end;
 			} else if (fd == bob_sock) {
@@ -844,6 +853,7 @@ static void *bob_thread(void *data)
 					struct sockaddr_un sun;
 
 					/* Connection from Alice, accept it so we can handle it. */
+					len = sizeof(sun);
 					sock_from_alice = accept(fd, (struct sockaddr *) &sun,
 							&len);
 					ret = add_sock_to_pollset(epfd, sock_from_alice,
@@ -859,6 +869,7 @@ static void *bob_thread(void *data)
 				} else if (event & EPOLLIN) {
 					(void) recv_otr_msg(sock_from_alice, bob_name,
 							alice_name, &oinfo);
+					messages_to_bob_inflight--;
 				}
 				continue;
 			} else {
@@ -1015,7 +1026,8 @@ static int init_client(void)
 	int ret;
 
 	/* Init libgcrypt threading system. */
-	gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
+	// Docs say this is obsolete since version 1.6
+	//gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
 
 	/* Init OTR library. */
 	OTRL_INIT;
@@ -1151,6 +1163,11 @@ int main(int argc, char **argv)
 
 	run();
 
+	otrl_userstate_free(user_state);
+	free(opt_instag_path);
+	free(opt_key_path);
+	free(opt_key_fp_path);
+
 	return exit_status();
 
 error:
```

mmilata:
@psychon thanks for looking into this! I threw your patch in the upstream direction: https://bugs.otr.im/lib/libotr/issues/90#note_449

DarthGandalf:
If that patch works, may you install the fixed libotr to travis to fix the test?

Meanwhile, I'll start reviewing the module itself, and will try to remove the sequences of magic numbers of PRIVMSG from test.

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        assert(ctx);
+        assert(ctx->username);
+        return PutModuleBuffered(CString("[") + ctx->username + "] " + sLine);
+    }
+
+    enum Color { Blue = 2, Green = 3, Red = 4, Bold = 32 };
+
+    CString Clr(Color eClr, const CString& sWhat) {
+        if (eClr == Bold) {
+            return CString("\x02") + sWhat + CString("\x02");
```
DarthGandalf: "\x02" + sWhat + "\x02"

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+    void WarnIfLoggingEnabled() {
+        assert(GetNetwork());
+        bool bUserMod = GetUser()->GetModules().FindModule("log");
+        bool bNetworkMod = GetNetwork()->GetModules().FindModule("log");
+
+        if (bUserMod || bNetworkMod) {
+            CString sMsg =
+                Clr(Red, "WARNING:") + " The log module is loaded. Type ";
+            if (bUserMod) {
+                sMsg += Clr(Bold, "/msg *status UnloadMod log") + " ";
```
DarthGandalf: StatusPrefix is not always `*`

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+                if (bNetworkMod) {
+                    sMsg += "and ";
```
DarthGandalf: Such constructed message would be hard to translate

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        AddCommand("GenKey",
+                   static_cast<CModCommand::ModCmdFunc>(&COtrMod::CmdGenKey),
+                   "[--really|--overwrite]", "Generate new private key.");
+
+        // Load list of ignored nicks
+        GetNV("ignore").Split(" ", m_vsIgnored, false);
+
+        // Warn if we are not an administrator
+        // We should check if we are the only administrator but the user map may
+        // not be
```
DarthGandalf: New lines in comments caused by format settings?

psychon:
Is it a good idea to have otr (which is intended to provide end-to-end security) be terminated by a middleman like ZNC?

DarthGandalf:
That should be the choice of the user.
The user can even wish to terminate it in their brain, not in a middleman like a computer :)

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        delete static_cast<COtrAppData*>(appdata);
+    }
+};
+
+class COtrMod : public CModule {
+    friend class COtrTimer;
+    friend class COtrGenKeyJob;
+
+  private:
+    static OtrlMessageAppOps m_xOtrOps;
```
mmilata: Yes - it's a structure with function pointers that are the same for every user.

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+    OtrlUserState m_pUserState;
+    CString m_sPrivkeyPath;
+    CString m_sFPPath;
+    CString m_sInsTagPath;
+    list<CString> m_Buffer;
```
mmilata: I'll rewrite this to user CBuffer, I probably overlooked that it existed (or didn't exist at the time).

modules/otr.cpp
DarthGandalf: `static const` then

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        assert(ad);
+        if (ad->bSmpReply) {
+            PutModuleContext(ctx,
+                             "Authentication in progress. Use Auth to "
+                             "respond with secret, or AuthAbort to abort it");
+            return;
+        }
+
+        CString sRest = sLine.Token(2, true);
+        if (sRest.length() == 0 || sRest[0] != '[') {
```
mmilata: The AuthQ syntax is `'authq' NICK '[' QUESTION ']' ANSWER` where both QUESTION and ANSWER can be composed of multiple words. The only reason [] are used is because they are used in the irssi otr plugin.

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+            sFlat += *it;
+        }
+        SetNV("ignore", sFlat, true);
+    }
+
+    bool IsIgnored(const CString& sNick) {
+        CString sNickLower = sNick.AsLower();
+
+        for (VCString::const_iterator it = m_vsIgnored.begin();
+             it != m_vsIgnored.end(); it++) {
```
mmilata: Oh, cool, didn't know C++11 had this. I can't use this syntax when deleting elements from the vector, right?

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+            if (sNickLower.WildCmp(*it)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    void CmdIgnore(const CString& sLine) {
+        if (sLine.Token(1).Equals("")) {
+            PutModuleBuffered("OTR is disabled for following nicks:");
```
DarthGandalf: m_vsIgnored should be a set, not a vector

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+            PutModuleBuffered(sMsg);
+        }
+    }
+
+    void CmdInfo(const CString& sLine) {
+        CTable table;
+        table.AddColumn("Peer");
+        table.AddColumn("State");
+        table.AddColumn("Fingerprint");
+        table.AddColumn("Act");
```
DarthGandalf: What is act? acting?
Reading code below, it's "active", but user probably doesn't read the code

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        if (err) {
+            PutModuleBuffered(CString("Failed to write fingerprints: ") +
+                              gcry_strerror(err));
+        }
+    }
+
+    void WarnIfLoggingEnabled() {
+        assert(GetNetwork());
+        bool bUserMod = GetUser()->GetModules().FindModule("log");
+        bool bNetworkMod = GetNetwork()->GetModules().FindModule("log");
```
DarthGandalf: It can also be global

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+
+    CString Clr(Color eClr, const CString& sWhat) {
+        if (eClr == Bold) {
+            return CString("\x02") + sWhat + CString("\x02");
+        } else {
+            return CString("\x03") + CString(eClr) + sWhat + CString("\x03");
+        }
+    }
+
+    static CString HumanFingerprint(Fingerprint* fprint) {
```
DarthGandalf: HumanReadableFingerprint()
I thought it's fingerprint of a human.

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        ConnContext* ctx = GetContextFromArg(sLine);
+        if (!ctx) {
+            return;
+        }
+
+        CString sSecret = sLine.Token(2, true);
+        DoSMP(ctx, "", sSecret);
+    }
+
+    void CmdAuthQ(const CString& sLine) {
```
DarthGandalf: I see some instructions to run Auth command, but never AuthQ

modules/otr.cpp
mmilata: Sorry for stupid question but how does one initialize `static const` struct?

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+                              " to OTR ignore list.");
+        }
+    }
+
+    virtual bool OnLoad(const CString& sArgs, CString& sMessage) {
+        // Initialize libgcrypt for multithreaded usage
+        gcry_error_t err;
+        err = gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
+        if (err) {
+            sMessage = (CString("Failed to initialize gcrypt threading: ") +
```
mmilata: I think setting the callbacks multiple times to the same structure shouldn't be a problem. As you note below, problem would be if some other module or its dependency performed this operation with different argument.

modules/otr.cpp
mmilata: In the integration test?

modules/otr.cpp
```diff
@@ -0,0 +1,1291 @@
...
+        char* newmessage = NULL;
+        OtrlTLV* tlvs = NULL;
+        ConnContext* ctx = NULL;
+        CString sNick = Nick.GetNick().AsLower();
+
+        if (IsIgnored(Nick.GetNick())) {
+            return CONTINUE;
+        }
+
+        const char* accountname = GetUser()->GetUserName().c_str();
```
mmilata: Shouldn't be a problem, each network has its own nick (context) database. Also each network has its own private key:(

mmilata:
> This will cause problems.

I'm sorry but it seems I won't be able to commit significant amount of time into this PR. I think I'd be able to address the minor issues found here but can't do bigger stuff like fixing libotr memleaks. I don't even really use znc with the module anymore:/

The time you guys invested in reviewing this PR is much appreciated, however if it looks it's going to cause problems in the future maybe it would be better to drop the PR before further time is wasted on it. What do you think?

modules/otr.cpp
DarthGandalf:
```c++
struct foo {
  static const int x;
};
const int foo::x = 42;
```

modules/otr.cpp
DarthGandalf: Users never look at integration test for documentation.
Instructions to run Auth command are sent to user by this module itself, in certain cases.

DarthGandalf:
> I don't even really use znc with the module anymore:/

This is even bigger problem than OTRL_INIT. That one can be worked around by making module global (with certain other changes) and declaring that it can't work together with some other modules. But this module may require maintenance even after merging the PR. If noone uses it, it'll be very hard to do.

Well, at least I learnt a bit about libotr, and you learnt a bit about C++ :)
Thanks for the PR anyway.

PR #1395, opened 2017-04-10
Log user quit messages

elyscape:
Since `OnIRCDisconnected()` doesn't include the quit message, and `OnQuit()` isn't called when the user quits, we need to hook `OnSendToIRC()`.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -114,6 +114,8 @@ class CLogMod : public CModule {
                 const vector<CChan*>& vChans) override;
     EModRet OnTopic(CNick& Nick, CChan& Channel, CString& sTopic) override;
 
+    EModRet OnSendToIRCMessage(CMessage& Message) override;
+
     /* notices */
     EModRet OnUserNotice(CString& sTarget, CString& sMessage) override;
     EModRet OnPrivNotice(CNick& Nick, CString& sMessage) override;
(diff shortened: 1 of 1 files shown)


elyscape:
An alternative way to handle this would be to add a new module hook that fires when a user quits, something like `OnUserQuit()/OnUserQuitMessage()`. Should I do that instead?

elyscape:
Looks like `OnUserQuit()` exists and is called when a client sends a quit message, which isn't the same thing. So in that case maybe `OnUserIRCQuit()/OnUserIRCQuitMessage()`?

modules/log.cpp
```diff
@@ -114,6 +114,8 @@ class CLogMod : public CModule {
                 const vector<CChan*>& vChans) override;
     EModRet OnTopic(CNick& Nick, CChan& Channel, CString& sTopic) override;
 
+    EModRet OnSendToIRC(CString &sLine) override;
```
DarthGandalf: Move & to left

modules/log.cpp
```diff
@@ -399,6 +401,18 @@ void CLogMod::OnQuit(const CNick& Nick, const CString& sMessage,
...
 }
 
+CModule::EModRet CLogMod::OnSendToIRC(CString &sLine) {
+    if (sLine.Token(0) != "QUIT") {
+        return CONTINUE;
+    }
+    CIRCNetwork* pNetwork = GetNetwork();
+    CMessage Message(sLine);
+    OnQuit(pNetwork->GetIRCNick(),
+            static_cast<CQuitMessage>(Message).GetReason(),
```
DarthGandalf: There is Message.As()

modules/log.cpp
```diff
@@ -399,6 +401,18 @@ void CLogMod::OnQuit(const CNick& Nick, const CString& sMessage,
     }
 }
 
+CModule::EModRet CLogMod::OnSendToIRC(CString &sLine) {
+    if (sLine.Token(0) != "QUIT") {
```
DarthGandalf: This may break if any tag is sent.

elyscape: How do you mean?

elyscape: Oh, I see. That being said, do we ever send quit messages with tags?

elyscape:
If #1398 is merged before this, I can rebase this on top of that to use `OnSendToIRCMessage` instead of `OnSendToIRC`.

elyscape:
Reworked to use `OnSendToIRCMessage()`.

PR #1396, opened 2017-04-10
Fix OnUserQuitMessage documentation

elyscape:
This fixes the misleading documentation for `OnUserQuitMessage()` ~and adds a new module hook (`OnQuitIRCMessage()`) that is called when ZNC quits an IRC network~.

### DIFF
--- a/include/znc/Modules.h
+++ b/include/znc/Modules.h
@@ -892,7 +892,7 @@ class CModule {
      */
     virtual EModRet OnUserTopicRequest(CString& sChannel);
 
-    /** This module hook is called when a user requests to quit from network.
+    /** This module hook is called when a client quits ZNC.
      *  @since 1.7.0
      *  @param Message The quit message the client sent.
      *  @return See CModule::EModRet.


elyscape:
It might be a good idea to rename `OnUserQuit()/OnUserQuitMessage()` to `OnClientQuit()/OnClientQuitMessage()`, especially since they haven't been part of an actual release yet. `OnQuitIRCMessage()` could then potentially be created as `OnUserQuitMessage()` or similar. I held off because I wasn't certain that was the right path forward. If that's desired, let me know and I'll make the changes here.

psychon:
Well... In general `OnFoo` is called when we get `FOO` from the IRC network and `OnUserFoo` is called when we get `FOO` from a client, so `OnUserQuit` is consistent with the existing naming (we get a `QUIT` from a user/client).
The `Message` suffix is then just the special `CMessage`-based version of that hook.

The only existing module hooks for "we are sending something to the IRC server" are `OnJoining`, `OnIRCRegistration` and `OnSendToIRC`. No clear naming pattern there...

So much for the naming. No idea if we need a special module hook for this, nor what to call it. I guess that's for @DarthGandalf to decide.

DarthGandalf:
I don't think this hook is very useful. Especially, as ZNC is usually staying online without quitting ever.
Also then we'll need to add "sending PRIVMSG", "sending PART", "sending TOPIC", etc.

`OnSendToIRC` is enough for log module... Side comment: there should be `OnSendToIRCMessage`, but currently every thing in ZNC which sends anything, sends a raw line, not a structured message.

elyscape:
That's fair. I'll remove the hook, then, and restrict this PR to just fixing the documentation of `OnUserQuitMessage()`.

Issue #1193, opened 2015-11-24
PutIRC(CMessage)


DarthGandalf:
##

DarthGandalf:
The main problem here: need to consider how tags interact with disabled capabilities.

elyscape:
The good news is that [the message tag spec says][message-tag]:
>Client and server MAY parse the tags even without any capabilities enabled. They SHOULD ignore tags of capabilities which are not enabled.

The bad news is that that says "SHOULD", not "MUST", so in theory a system could do anything. As a result, I'm not quite sure what we should do.

[message-tag]: http://ircv3.net/specs/core/message-tags-3.2.html

DarthGandalf:
MAY doesn't mean "able to".
As I understand it, ZNC should send a tag if that tag is enabled or if [`draft/message-tags-0.2`](http://ircv3.net/specs/core/message-tags-3.3.html) is enabled. It's not ambiguous.
I was referring more to API inside ZNC - should PutIRC(msg) send the message as is, or strip unsupported tags. For consistency with CClient::PutClient(const CMessage& Message), as it's implemented now, probably it should strip unsupported tags.

elyscape:
>As I understand it, ZNC should send a tag if that tag is enabled or if [`draft/message-tags-0.2`][client-only-tags] is enabled.

It looks like client-only tags, as specified by that spec, are tags beginning with `+`. As such, ZNC should send the tag if the corresponding tag is enabled or if `draft/message-tags-0.2` is enabled and the tag begins with `+`. I can make a pull request adding that functionality.

>For consistency with CClient::PutClient(const CMessage& Message), as it's implemented now, probably it should strip unsupported tags.

Makes sense to me.

[client-only-tags]: http://ircv3.net/specs/core/message-tags-3.3.html

elyscape:
Actually, rereading the spec, you were right. The difference is that client-only tags should not be touched by the server.

elyscape:
Hmm, a pull request adding support for `draft/message-tags-0.2` would actually be more complicated than I thought, as it would need to also add support for [`TAGMSG`][TAGMSG] and [messages with up to 4607 bytes of tags][size-limit]. The former is easy, but the latter entails performing checks on the size of both server- and client-initiated tag sets, refraining from adding server-initiated tag sets that would exceed 510 bytes, and responding with 417 `ERR_INPUTTOOLONG` to messages from clients with tag data that exceeds 4094 bytes.

[...]

PR #1398, opened 2017-04-11
Add OnSendToIRCMessage and OnSendToClientMessage

elyscape:
As described in [this comment](https://github.com/znc/znc/pull/1396#issuecomment-293179090).


elyscape:
@DarthGandalf Should I implement CMessage variants of `PutIRC` and `PutUser`?

DarthGandalf:
I see several issues:
* Any changes to the message made by modules are ignored
* Rather than parsing the line for the sake of calling modules, PutClient should accept CMessage, and stringify it to get the line to send. Also see #1193

elyscape:
@DarthGandalf Any recommendations for handling changes to the message made by modules? Should we make a copy of the `CMessage` before sending it to modules?

As for making a version of `PutClient` that accepts a `CMessage`, [one already exists][putclient], but ultimately [it stringifies the `CMessage`][tostring] and [calls the `CString` variant][call], which is what [calls modules and writes to the client][write]. Changing things so that this is handled by `PutClient(CMessage)` would require some more significant changes. What do you think?

[putclient]: https://github.com/znc/znc/blob/3189ce7/src/Client.cpp#L495
[tostring]: https://github.com/znc/znc/blob/3189ce7/src/Client.cpp#L566
[call]: https://github.com/znc/znc/blob/3189ce7/src/Client.cpp#L592
[write]: https://github.com/znc/znc/blob/3189ce7/src/Client.cpp#L488-L492

elyscape:
Oh wait, I misunderstood your first issue. Give me a moment and I'll fix that.

elyscape:
Actually, on review, it wasn't that difficult to change `PutClient` such that the `PutClient(CMessage)` handled actually sending messages and calling hooks.

elyscape:
Another bonus from doing that is that it eliminated the need for a separate module hook for specifying message tags, as that can now be done inside `OnSendToClientMessage()`.

elyscape:
@DarthGandalf Please take another look when you have the opportunity. Thanks!

modules/modpython/znc.py
```diff
@@ -655,6 +655,22 @@ def OnTopicMessage(self, msg):
...
+        old = modified.s = msg.ToString()
+        ret = self.OnChanBufferPlayLine(modified, msg.GetClient())
+        if old != modified.s:
+            msg.Parse(modified.s)
+        return ret
+
+    def OnSendToIRCMessage(self, msg):
+        modified = String()
+        old = modified.s = msg.ToString()
+        ret = self.OnChanBufferPlayLine(modified)
```
elyscape: Good catch. Will fix.

elyscape:
The copying behavior was based on the way [`OnChan/PrivBufferPlayMessage` is handled][onbufferplaymessage], along with the way [most of the other `OnXMessage` hooks are handled][message-hooks]. While there is a performance hit for modules that don't override `OnSendToIRCMessage` or `OnSendToClientMessage`, if those are overridden then the non-Message variants aren't even called. Then again, `OnSendToIRCMessage` and `OnSendToClientMessage` will be called much more often, so I'll switch to a method of calling both directly.

[onbufferplaymessage]: https://github.com/znc/znc/blob/f37aa30/src/Modules.cpp#L753-L772
[message-hooks]: https://github.com/znc/znc/pull/1000/files#diff-12d169eaaada02c163f80979e8ead3a1

DarthGandalf:
For those OnXMessage hooks, I don't know how to avoid such per-module conversion back and forth. If you have any better idea, please say.

Hopefully, after several versions, after the sufficient transition period, old deprecated hooks can be removed, and such compatibility code can be dropped.

elyscape:
Okay, I've switched things such that both `OnSendToClient/IRC` and `OnSendToClient/IRCMessage` are called for performance reasons. I've also added tests for the new hooks.

src/Client.cpp
```diff
@@ -483,13 +483,8 @@ CString CClient::GetFullName() const {
...
 void CClient::PutClient(const CString& sLine) {
-    bool bReturn = false;
-    CString sCopy = sLine;
-    NETWORKMODULECALL(OnSendToClient(sCopy, *this), m_pUser, m_pNetwork, this,
-                      &bReturn);
-    if (bReturn) return;
-    DEBUG("(" << GetFullName() << ") ZNC -> CLI [" << sCopy << "]");
-    Write(sCopy + "\r\n");
+    CMessage Message(sLine);
+    PutClient(Message);
```
DarthGandalf: I like how you handled this.

src/IRCSock.cpp
```diff
@@ -1154,13 +1154,20 @@ void CIRCSock::TrySend() {
         m_iSendsAllowed--;
         bool bSkip = false;
         CString& sLine = m_vsSendQueue.front();
-        IRCSOCKMODULECALL(OnSendToIRC(sLine), &bSkip);
+
+        CMessage Message(sLine);
```
DarthGandalf: Probably send queue should store messages instead of lines... That's tied to #1193

elyscape: Yeah, I figured that I would handle in a separate pull request.

test/IRCTest.h
```diff
@@ -188,6 +198,7 @@ class TestModule : public CModule {
     std::vector<CClient*> vClients;
     std::vector<CChan*> vChannels;
     EModRet eAction = CONTINUE;
+    bool bSendHooks = false;
```
elyscape: I did this so that, when testing other hooks, the `OnSendToClient/IRCMessage` hooks don't fire. Before I put it in here, it was necessary to modify other tests to indicate that `OnSendToClient/IRCMessage` would be included in the list of hooks that fired. Additionally, it simplified handling of `HALT` for the tests.

DarthGandalf: I don't see a problem with indicating full list of callbacks called.
How is HALT simplified?

elyscape: As an example of the sort of thing that would happen, [the test for `OnJoinMessage`][onjoinmessage] would have had to be changed from this:
```cpp
TEST_F(IRCSockTest, OnJoinMessage) {
    CMessage msg(":somebody JOIN #chan");
    m_pTestSock->ReadLine(msg.ToString());

    EXPECT_THAT(m_pTestModule->vsHooks, ElementsAre("OnJoinMessage"));
    EXPECT_THAT(m_pTestModule->vsMessages, ElementsAre(msg.ToString()));
    EXPECT_THAT(m_pTestModule->vNetworks, ElementsAre(m_pTestNetwork));
    EXPECT_THAT(m_pTestModule->vChannels, ElementsAre(m_pTestChan));
}
```
to this:
```cpp
TEST_F(IRCSockTest, OnJoinMessage) {
    CMessage msg(":somebody JOIN #chan");
    m_pTestSock->ReadLine(msg.ToString());

    EXPECT_THAT(m_pTestModule->vsHooks, ElementsAre("OnJoinMessage", "OnSendToClientMessage"));
    EXPECT_THAT(m_pTestModule->vsMessages, ElementsAre(msg.ToString(), msg.ToString()));
    EXPECT_THAT(m_pTestModule->vNetworks, ElementsAre(m_pTestNetwork, m_pTestNetwork));
    EXPECT_THAT(m_pTestModule->vChannels, ElementsAre(m_pTestChan, m_pTestChan));
}
```
Basically, it polluted many of the other tests. Worse, it wasn't super obvious when writing hook tests whether or not they would be affected.

The issue with `HALT` may have been something I was misremembering, as I can't find the relevant failed tests, but the idea is that, when testing some hook that also results in `OnSendToClient/IRCMessage` being called, if the hook action is set to `HALT`, it wasn't necessarily clear if the hook actually being tested would halt execution or if the `OnSend` hook would do so, resulting in incomplete testing.

By adding a field to indicate whether `OnSendToClient/IRCMessage` should process its calls, these issues were avoided.

[onjoinmessage]: https://github.com/elyscape/znc/blob/823ac07/test/IRCSockTest.cpp#L226-L234

PR #1399, opened 2017-04-14
missingmotd: Include nick in IRC numeric command

kylef:
Since the nick argument is missing, "MOTD File is missing" is taking the nicks place and could lead to client confusion that the 422 command is sent to "MOTD File is missing".

Found a report of this against a client because the client was updating the nick from the MOTD missing message.

screen shot 2017-04-14 at 19 34 08

### DIFF
--- a/modules/missingmotd.cpp
+++ b/modules/missingmotd.cpp
@@ -15,13 +15,14 @@
  */
 
 #include <znc/Modules.h>
+#include <znc/Client.h>
 
 class CMissingMotd : public CModule {
   public:
     MODCONSTRUCTOR(CMissingMotd) {}
 
     void OnClientLogin() override {
-        PutUser(":irc.znc.in 422 :MOTD File is missing");
+        PutUser(":irc.znc.in 422 " + GetClient()->GetNick() + " :MOTD File is missing");
     }
 };
 


modules/missingmotd.cpp
```diff
@@ -15,13 +15,14 @@
...
 #include <znc/Modules.h>
+#include <znc/Client.h>
 
 class CMissingMotd : public CModule {
   public:
     MODCONSTRUCTOR(CMissingMotd) {}
 
     void OnClientLogin() override {
-        PutUser(":irc.znc.in 422 :MOTD File is missing");
+        PutUser(":irc.znc.in 422 " + GetClient()->GetNick() + " :MOTD File is missing");
```
DarthGandalf: Should it send it to all clients, or only to new client?

kylef: To be honest, I'm not sure what this module is used for so I do not know. Just had report of broken behaviour in Palaver due to the missing nickname.

elyscape: Most likely only a new client, as that's when MOTDs are sent.

aetsher:
.

modules/missingmotd.cpp
kylef: Yeah, but this will usually result in receiving a duplicate MOTD as the connected IRC server will also send an MOTD.

Is this module meant to block the servers MOTD? In that case, why not just use the block_motd module?

kylef:
Going to merge as this does fix a problem. @elyscape if you'd like to make the behavioural change you mentioned please create a pull request.

elyscape:
I'd like to but I don't know what this module is used for either. Maybe doing so would break something?

psychon:
AFAIR some weird IRC client waited until the end of the MOTD before it considered that its login succeeded. On "why does this module exist": Git history points at https://github.com/znc/znc/commit/2370567f9c031591cb67a92c982e7075d4cb497d.

kylef:
Okay, then I made https://github.com/znc/znc/pull/1403 to improve the module further.

Issue #1196, opened 2015-11-26
autoop: simplify/cleanup the "listusers" output. 


KindOne-:
The autoop module got a nice new feature that allows users to have multiple masks per user. https://github.com/znc/znc/commits/master/modules/autoop.cpp?author=orchid-hybrid 

Those commits created something like this:

```
<KindOne> listusers
<*autoop> +============+========================================+===========+=================+
<*autoop> | User       | Hostmasks                              | Key       | Channels        |
<*autoop> +============+========================================+===========+=================+
<*autoop> | foobar     | *!*@foobar.com                         | __NOKEY__ | #znc            |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | KindOne    | *!*@colchester-lug/silly-fool/donut    | __NOKEY__ | #znc            |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | |-         | *!*KindOne@gateway/shell/trekweb.org/* |           |                 |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | `-         | *!uid3688@gateway/web/irccloud.com/*   |           |                 |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | znc        | *!*@1.2.3.4                            | __NOKEY__ | #znc            |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | |-         | *!*@znc.rules.com                      |           |                 |
<*autoop> +------------+----------------------------------------+-----------+-----------------+
<*autoop> | `-         | *!*@znc/user/znc                       |           |                 |
<*autoop> +============+========================================+===========+=================+
```

Then somewhere in 1.7.x the output got altered and now it just looks horrible with all the ```User: |-```, ```User: `-```, and ```-----``` clutter.

```
<KindOne> listusers
<*autoop> ------
<*autoop> User: foobar
<*autoop> Hostmasks: *!*@foobar.com
<*autoop> Key: __NOKEY__
<*autoop> Channels: #znc
<*autoop> ------
<*autoop> User: KindOne
<*autoop> Hostmasks: *!*@colchester-lug/silly-fool/donut
<*autoop> Key: __NOKEY__
<*autoop> Channels: #znc
<*autoop> ------
<*autoop> User: |-
<*autoop> Hostmasks: *!*KindOne@gateway/shell/trekweb.org/*
<*autoop> ------
<*autoop> User: `-
[...]

PR #1415, opened 2017-05-25
autovoice: accept multiple hostmasks like autoop.

RealKindOne:
Copy and paste of the autoop module, without the ChallengeResponse,
VerifyResponse, ProcessQuery, keys, etc...

I've tested the module and it looks like it works perfectly.

https://gist.github.com/RealKindOne/5c1b0dff9f122b4b9f0782de1a16a0c5

### DIFF
--- a/modules/autovoice.cpp
+++ b/modules/autovoice.cpp
@@ -15,11 +15,13 @@
  */
 
 #include <znc/IRCNetwork.h>
-#include <znc/Modules.h>
 #include <znc/Chan.h>
 
 using std::map;
 using std::set;
+using std::vector;
+
+class CAutoVoiceMod;
 
 class CAutoVoiceUser {
   public:
@@ -27,16 +29,16 @@ class CAutoVoiceUser {
 
     CAutoVoiceUser(const CString& sLine) { FromString(sLine); }
 
-    CAutoVoiceUser(const CString& sUsername, const CString& sHostmask,
+    CAutoVoiceUser(const CString& sUsername, const CString& sHostmasks,
                    const CString& sChannels)
-        : m_sUsername(sUsername), m_sHostmask(sHostmask) {
+        : m_sUsername(sUsername) {
+        AddHostmasks(sHostmasks);
         AddChans(sChannels);
     }
 
     virtual ~CAutoVoiceUser() {}
 
     const CString& GetUsername() const { return m_sUsername; }
-    const CString& GetHostmask() const { return m_sHostmask; }
 
     bool ChannelMatches(const CString& sChan) const {
         for (const CString& s : m_ssChans) {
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Instead of copypasting it should reuse the code, e.g. by merging two modules together, or perhaps using some common header. Otherwise fixes like #1196 will inevitably affect only one of the two modules.

Manual test is good... but not enough. Later someone can accidentally break something and not notice, because there is no integration test for this module.

modules/autovoice.cpp
```diff
@@ -335,16 +395,16 @@ class CAutoVoiceMod : public CModule {
...
         if (m_msUsers.find(sUser) != m_msUsers.end()) {
             PutModule(t_s("That user already exists"));
             return nullptr;
         }
 
-        CAutoVoiceUser* pUser = new CAutoVoiceUser(sUser, sHost, sChans);
+        CAutoVoiceUser* pUser = new CAutoVoiceUser(sUser, sHosts, sChans);
         m_msUsers[sUser.AsLower()] = pUser;
-        PutModule(t_f("User {1} added with hostmask {2}")(sUser, sHost));
+        PutModule(t_f("User {1} added with hostmask(s) {2}")(sUser, sHosts));
```
DarthGandalf: `t_p()` is for selecting between singular and plural.

RealKindOne: https://github.com/znc/znc/blob/master/modules/autoop.cpp#L490 Is that also wrong?

DarthGandalf: Yes, thanks, I didn't notice it when making strings in that module to be translatable.

RealKindOne:
```
$ make
Building module autovoice...
autovoice.cpp: In member function ‘CAutoVoiceUser* CAutoVoiceMod::AddUser(const CString&, const CString&, const CString&)’:
autovoice.cpp:407:60: error: no matching function for call to ‘CAutoVoiceMod::t_p(const char [36])’
         PutModule(t_p("User {1} added with hostmask(s) {2}")(sUser, sHosts));
                                                            ^
In file included from ./../include/znc/znc.h:22:0,
                 from ./../include/znc/IRCNetwork.h:24,
                 from autovoice.cpp:17:
./../include/znc/Modules.h:1346:26: note: candidate: CInlineFormatMessage CModule::t_p(const CString&, const CString&, int, const CString&) const
     CInlineFormatMessage t_p(const CString& sEnglish, const CString& sEnglishes,
                          ^~~
./../include/znc/Modules.h:1346:26: note:   candidate expects 4 arguments, 1 provided
Makefile:131: recipe for target 'autovoice.o' failed
make[1]: *** [autovoice.o] Error 1
Makefile:121: recipe for target 'modules' failed
make: *** [modules] Error 2
```

DarthGandalf:
Well, of course, because `t_p` accepts a different number of parameters.

PR #1416, opened 2017-05-29
Change default flood rates to match standard IRC networks settings, pre…

rubinlinux:
…vent excess flood problems

I noticed that the default network flood rates are wrong (for most networks). It has been very standard since the beginning of IRC to use a value of 2 seconds per line, with a 10 line head start. Almost all networks work this way, and if you don't follow, having a lot of channels or a lot of clients connected to ZNC results in excess flood disconnections and much frustration for the user. 

I think you should make the default value for these 2 and 9 instead of 1 and 4 so that this nasty bug doesn't hit people who will not know how to debug it. You find these same types of settings built into every IRC client. 2 and 10 is the standard. 2 and 9 provides a slight buffer so there is no mistake.

Thanks
-Rubin


psychon:
Sorry, but

> It has been very standard since the beginning of IRC

[citation needed]

> Almost all networks work this way,

[citation needed]

> I think you should make the default value for these 2 and 9 instead of 1 and 4 so that this nasty bug doesn't hit people

I'm not sure I understand why this causes people to flood off. If I sent 9 lines "immediately" and then wait for two seconds, then I have not yet exceeded the new limit, but I already exceeded the old one. Thus, doesn't this commit actually kind-of-raise the limit instead of lower it?
(kind-of, because after the limit is hit, we only send 0.5 lines per second while previously this was 1 line per second, more or less as you say)

DarthGandalf:
> [citation needed]

+1

> Should that test really implicitly test for the default values...?

It tests only UI of controlpanel without testing that values modified there affect anything in ZNC behavior. But that test triggers some functions not used in other tests...
While the test should be ~fixed~rewritten, I wouldn't oppose someone to just remove it for now. But that's out of scope of this change.

psychon:
I looked at inspircd's default config. If I understand [this](https://github.com/inspircd/inspircd/blob/5583d8dc5ff202d411d7985b6bbfb240beeacddd/docs/conf/inspircd.conf.example#L361-L374) correctly, then inspircd defaults to a "head start" of 10 lines and then allows one line per second.

I did not find anything in unrealircd, but [queue size limits](https://github.com/unrealircd/unrealircd/blob/4d0dbd40c4eee24ddb35c9f9fb64ba2d63774034/doc/conf/examples/example.conf#L87-L88).

rubinlinux:
Alright I'll try my best at some citations. 

# Citation 1:  For the first citation I would like to argue my own experience.

I have been writing bots for IRC since the 90s and the 2/10 rule was kinda common knowledge and used widely in bots (and shared with newbies learning).

# Citation 2: But on to some real world code. The Undernet ircu:

ircu/undernet:
  ircd/s_bsd.c line 650: 
```c
      while (DBufLength(&(cli_recvQ(cptr))) && !NoNewLine(cptr) &&
             (IsTrusted(cptr) || cli_since(cptr) - CurrentTime < 10))
```
The '10' there is the 10 I spoke of. Each line adds to the timestamp in the future where we start reading from that client again. If that value gets to be 10 or bigger, the ircd stops reading lines from it until time advances. The excess flood comes from the sendq exceeding (1024 usually) because the ircd isnt clearing it out.

The 2 part comes from ircd/parse.c line 887:
```c
  if ((mptr->flags & MFLG_SLOW) || !IsAnOper(cptr))
    cli_since(cptr) += (2 + i / 120);
```

This is way complicated, but essentially the cli_since value is increased by 2 + (small factor depending on line length)


# Citation 3: Hexchat client source:

Hexchat nearly duplicates this to avoid triggering it, as seen in src/common/server.c line 197 and 207:
```c
                if (serv->next_send - now >= 10)
```
...
```c
                serv->next_send += (2 + i / 120);
```

#Citation 4: RFC 1459

The final citation I would offer you is the IRC RFC itself, which I think we can all agree ZNC should try to follow:
https://www.ietf.org/rfc/rfc1459.txt
In section 8.10 it says:
```
8.10 Flood control of clients

   With a large network of interconnected IRC servers, it is quite easy
   for any single client attached to the network to supply a continuous
   stream of messages that result in not only flooding the network, but
   also degrading the level of service provided to others.  Rather than
   require every 'victim' to be provide their own protection, flood
   protection was written into the server and is applied to all clients
   except services.  The current algorithm is as follows:

        * check to see if client's `message timer' is less than
          current time (set to be equal if it is);

        * read any data present from the client;

        * while the timer is less than ten seconds ahead of the current
          time, parse any present messages and penalize the client by
          2 seconds for each message;
```
...

So, yes Modern IRC networks change a lot of things, and relaxing flood settings may be done, and some networks are more relaxed than others. Newer irc servers such as inspircd.

I argue that ZNC default values should be configured such that users adding any new network is unlikely to have a problem. The values are changeable to them in the control panel, so if they know their network has less limits, they can adjust it for better performance. 

But I don't think it should be on the user to 'figure out' that the reason their ZNC gets disconnected for excess flood a lot when they add more channels is because the defaults don't work with their (standards compliant) network of choice.

I have personally seen a lot of ZNC users struggle with this on my network, which is why I went investigating why it was happening and discovered the errant defaults.

Basically the default should be the one that works most widely, not the one that offers the best performance in a limited case while breaking others.

Thanks again,
-Rubin

DarthGandalf:
This sounds good.
But I still can't merge this if this breaks tests. Please fix the test.

rubinlinux:
Can you point me in the right direction for how to find them? I'm not much
of a cpp developer.

Thanks, -Rubin

jobe1986:
Here rubin:
https://github.com/znc/znc/blob/ab34760613214a898ef8614abd10eb3cdb6c50fd/test/integration/main.cpp#L801

The results from the failed tests:
[2017-05-30 21:52:54.107264] (user/test) CLI -> ZNC [PRIVMSG *controlpanel :GetNetwork FloodRate user freenode]
[2017-05-30 21:52:54.107307] (user/test) ZNC -> CLI [:*controlpanel!znc@znc.in PRIVMSG nick :FloodRate = 2.00]

You'll probably have to fix the test for FloodBurst too as that expects 4, and you've changed it to 9

rubinlinux:
Looks like I've made a mess of the commit logs on this, apologies (still learning how pull requests work). Let me know if you'd prefer I close this and open a clean one. (Assuming the tests pass now, we'll see in a bit)
-Rubin

psychon:
@rubinlinux I think [this](https://github.com/rubinlinux/znc/blob/bfafaa4f7ef61bfdfdec1c74aef35e8c4255f51a/test/integration/main.cpp#L1389) is where the test fail right now. Needs to be changed to 2.00 as well.

> Looks like I've made a mess of the commit logs on this, apologies (still learning how pull requests work). 

No problem, I think. @DarthGandalf could rebase this onto master before merging to get rid of the merges (or just squash the PR into a single commit).

DarthGandalf:
No problem, it's a good opportunity for to learn how to do it :)
No need to close pull request, you can fix it locally, then force-push to your `patch-1` branch.

rubinlinux:
I think I fixed the tests,  Can I trigger the checks to run again here?

DarthGandalf:
Either push, so that they trigger automatically on CI, or run `make inttest` locally, and then push.

rubinlinux:
Do I need to worry about this codecov/patch failed check? What is that?

psychon:
Thanks for going through the trouble to make this work!

@DarthGandalf What's the further plan here?

> Do I need to worry about this codecov/patch failed check?

No

> What is that?

It's supposed to help with having 100% code coverage. Basically, this says that no unit tests execute the code that you changed. I think there must be some error here, because the changed lines are definitely executed. I'd just ignore that.

DarthGandalf:
It says 2 files decrease their coverage:
* Template rendering in webadmin is not covered, and you change these lines.
* IRCSock.cpp is a more interesting case. https://codecov.io/gh/znc/znc/pull/1416/src/src/IRCSock.cpp says that with old flood settings a PONG was queued at least once, with new settings it was always sent immediately during the test (PING response is the only thing which uses `PutIRCQuick`

So, there is no error, but I wouldn't worry about the check.

PR #1428, opened 2017-07-11
Clearer ControlPanel/{Add,Del}Server syntax usage help

Zarthus:
please squash before merge


modules/controlpanel.cpp
```diff
@@ -1130,7 +1132,7 @@ class CAdminMod : public CModule {
         CString sPass = sLine.Token(5);
 
         if (sServer.empty()) {
-            PutModule("Usage: DelServer <username> <network> <server>");
+            PutModule("Usage: DelServer <username> <network> <server> [port] [password]");
```
psychon: Wouldn't it be enough to just change the `Usage` messages and to leave the rest of the code unchanged?
The rest of this commit actually loses functionality: You can no longer use `+6697` as a port number to specify SSL. Please look at what `CIRCNetwork::AddServer(const CString&)` actually does: https://github.com/znc/znc/blob/master/src/IRCNetwork.cpp#L1129-L1150

Zarthus: That sounds like a flaw in the AddServer API for taking a uInt rather than a CString, I looked at `DelServer` (Which is newer code) and seemed to be more concrete, therefore it made more sense for me to refactor the old to use the new. You raise a valid concern though.

Zarthus: For what it's worth, I use docs.znc.in for my development, which does not point out this quirk.

PR #1434, opened 2017-08-17
controlpanel: CloneNetwork functionality

kashike:


modules/controlpanel.cpp
```diff
@@ -619,6 +619,35 @@ class CAdminMod : public CModule {
...
+        }
+
+        CIRCNetwork* pOldNetwork = FindNetwork(pUser, sOldNetwork);
+        if (!pOldNetwork) {
+            PutModule("Error: Network [" + sOldNetwork + "] not found.");
+            return;
+        }
+
+        CIRCNetwork* pNewNetwork = new CIRCNetwork(pUser, *pOldNetwork);
+        pNewNetwork->SetName(sNewNetwork);
```
DarthGandalf: This won't work if new name contains invalid characters.

DarthGandalf:
Need to check `pUser->HasSpaceForNewNetwork()` as AddNetwork does, otherwise this will be used to workaround limits set by administrator.

kashike:
Is this what you meant?

modules/controlpanel.cpp
psychon: > Is this what you meant?

`SetName` can fail. Please look at the following and handle this function's return value:
https://github.com/znc/znc/blob/f885699d1a3cb6376dc6162f9e8440791034a329/src/IRCNetwork.cpp#L771-L778

kashike: https://github.com/znc/znc/pull/1434/files#diff-29017c32fc7e54eb3ad373b49ef002f1R633

DarthGandalf:
What happens if network with the new name exists already?
Any test for this?

PR #1435, opened 2017-08-17
controlpanel: allow get/set of active/connect enabled

kashike:


DarthGandalf:
There is `/znc disconnect` already...

kashike:
It isn't the same. `disconnect `will only disconnect, it doesn't allow setting to `true`. In my use case, I want to use a non-active network as a template, along with #1434, and enable after configuration is done.

DarthGandalf:
To connect, there is /znc connect

2017-08-20 1:50 GMT+01:00 kashike <notifications@github.com>:

PR #1443, opened 2017-10-01
route_replies: rizon: [MODE] Add 403, 442 expectations

### LINKED ISSUE #1421: route_replies module doesn't recognise ERR_NOTONCHANNEL reply to MODE
It seems that some servers (Rizon) may send `ERR_NOTONCHANNEL` (`442`) when a mode `e` or `I` list is requested for a channel that the user is not on.
This confused the `route_replies` module.

Running ZNC on Ubuntu 16.04.2:
```
ZNC 1.7.x-git-764-2e90c9a - http://znc.in
IPv6: yes, SSL: yes, DNS: threads, charset: yes, i18n: no, build: cmake
```

Zarthus:
Rizon replies with a 403 if the channel does not exist,
and with a 442 if you're not on that channel.

Resolves #1421

### DIFF
--- a/modules/route_replies.cpp
+++ b/modules/route_replies.cpp
@@ -171,6 +171,8 @@ static const struct {
         // MODE e
         {"348", false},
         {"349", true},
+        {"403", true}, /* rfc1459 ERR_NOSUCHCHANNEL */
+        {"442", true}, /* rfc1459 ERR_NOTONCHANNEL */
         {"467", true}, /* rfc1459 ERR_KEYSET */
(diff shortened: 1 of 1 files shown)


modules/route_replies.cpp
```diff
@@ -171,6 +171,8 @@ static const struct {
         // MODE e
         {"348", false},
         {"349", true},
+        {"403", true},
+        {"442", true},
```
DarthGandalf: Add a comment similar to below?

Zarthus: Done.

PR #1444, opened 2017-10-01
Tidy up README

Zarthus:
Adds ToC, minor grammar improvements, web link improvements
as GitHub does not support ircs:// in markdown

Rendered view available here:

https://github.com/Zarthus/znc/blob/tidy_up_readme/README.md

Further suggestions:

- We should remove this block because it mentions a bad practice that voids warranty, it feels better belonging on a wiki:
https://github.com/znc/znc/blob/a75ad25aaeb4b5cca71057de5a5cdf2a399235a0/README.md#L168-L170

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,16 @@
 
 ## Table of contents
 
-- Minimal Requirements
-- Optional Requirements
-- Installing ZNC
-- Setting up znc.conf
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Saving config is fine, reloading it not so.

PR #1445, opened 2017-10-02
debugging: Add Filter method to hide filter sensitive data

Zarthus:
Remove sensitive data from DEBUG mode.

So far only includes the /PASS command, but can be further extended in the future if we'd like to.


DarthGandalf:
On one hand, this is good.
On the other hand, user can't rely on this filtering anyway, because PASS is far the only sensitive command. But seeing PASS replaced can trick them into believing that other stuff is filtered too.

Other stuff includes: nickserv passwords, channel keys. Channel names? Different people consider different things as sensitive. But the less info is in the log, the harder it is to troubleshoot.

Zarthus:
The framework now exists to add more things to it, but I kept it extremely minimal. You'll most commonly be seeing the ZNC password, as NickServ auth is only sent once, but ZNC password every time you connect. NickServ password also is already in file.

I agree we shouldn't go overboard, which is why I also wrote #1446: If you don't want someone to see it by accident, don't send a NickServ password while you're connected and debug mode is on.

DarthGandalf:
NickServ password is sent even without client connected, by sasl module, and by nickserv module.

Zarthus:
if it's automated you should consider your password compromised and change it ;)

PR #1446, opened 2017-10-02
transperancy: Make the user aware that DEBUG mode is enabled.

Zarthus:
Shows the user when DEBUG mode is enabled.

```
(14:45:47) * Connect retry #1 server.com (+32498)
(14:45:47) Caps supported: batch cap-notify echo-message multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso
(14:45:48) Caps set: multi-prefix server-time userhost-in-names
(14:45:48) Password required
(14:45:47) -irc.znc.in- *** ZNC is presently running in DEBUG mode. Sensitive data during your current session may be exposed to the host.
(14:45:47) -irc.znc.in- *** You need to send your password. Configure your client to send a server password.
(14:45:47) -irc.znc.in- *** To connect now, you can use /quote PASS <username>:<password>, or /quote PASS <username>/<network>:<password> to connect to a specific network.
```

```
(14:45:49) <*status> ZNC is presently running in DEBUG mode. Sensitive data during your current session may be exposed to the host.
```

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -20,6 +20,7 @@
 #include <znc/User.h>
 #include <znc/IRCNetwork.h>
 #include <znc/Query.h>
+#include <znc/ZNCDebug.h>
 
 using std::set;
 using std::map;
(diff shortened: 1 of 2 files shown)


DarthGandalf:
Sensitive data may be exposed regardless of debug mode. The question is whether user trusts provider or not. Provider can see passwords even without debug mode.

Zarthus:
Sure, but when znc is running in the foreground it's a logical assumption someone is actively trying to watch the output.

Nothing stops anyone from logging all raw input without user consent.

PR #1449, opened 2017-10-09
debug: transperancy: Stop nagging the user with --enable-debug.

Zarthus:
Only nag if a TTY is connected to the ZNC session.

KindOne pointed me towards the fact that Debug::CDebug() is always true with --enable-debug.

### DIFF
--- a/include/znc/ZNCDebug.h
+++ b/include/znc/ZNCDebug.h
@@ -46,6 +46,7 @@ class CDebug {
     static bool StdoutIsTTY() { return stdoutIsTTY; }
     static void SetDebug(bool b) { debug = b; }
     static bool Debug() { return debug; }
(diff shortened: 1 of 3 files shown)


DarthGandalf:
Next step is to add `VeryActiveDebug()`?
Why you don't change `Debug` to ignore `--enable-debug`?

Zarthus:
>Why you don't change Debug to ignore --enable-debug?

(1) Because it looks intentional and I don't understand ZNC internals enough to know the implications.
https://github.com/znc/znc/blob/2ebd358521bde0b08b364d41aaa806df57e36c1b/include/znc/ZNCDebug.h#L24-L34
(2) I considered making Debug() take a boolean "only if tty" but I didn't think that solution made much more sense. It also gets called often so I don't think mucking with the logic to make it more complex is desirable.

PR #1455, opened 2017-10-24
Add sub-second precision for timestamp formatting

CyberShadow:
This pull request adds the prerequisites and implementation for allowing ZNC to display timestamps with sub-second precision.

Notable user-visible changes is that the log module's `-timestamp` parameter, and ZNC's `TimestampFormat` option, now understand the additional formatting sequences `%L` and `%N`. These are non-standard extensions to `strftime`. The characters chosen here use the same syntax as Ruby's `strftime` implementation (though other languages extend `strftime` in a similar way).

### DIFF
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -79,13 +79,11 @@ class CUtils {
     static CString CTime(time_t t, const CString& sTZ);
     static CString FormatTime(time_t t, const CString& sFormat,
                               const CString& sTZ);
-    /** Supports additional format specifiers for formatting sub-second values:
+    /** Supports an additional format specifier for formatting sub-second values:
      *
-     * - %L - millisecond, 3 digits (ruby extension)
-     * - %N - sub-second fraction (ruby extension)
-     *   - %3N - millisecond
-     *   - %6N - microsecond
-     *   - %9N - nanosecond (default if no digit specified)
+     * - %f - sub-second fraction
+     *   - %3f - millisecond (default, if no width is specified)
+     *   - %6f - microsecond
      *
      * However, note that timeval only supports microsecond precision
      * (thus, formatting with higher-than-microsecond precision will
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -79,6 +79,23 @@ class CUtils {
     static CString CTime(time_t t, const CString& sTZ);
     static CString FormatTime(time_t t, const CString& sFormat,
                               const CString& sTZ);
+    /** Supports additional format specifiers for formatting sub-second values:
+     *
+     * - %L - millisecond, 3 digits (ruby extension)
+     * - %N - sub-second fraction (ruby extension)
+     *   - %3N - millisecond
+     *   - %6N - microsecond
+     *   - %9N - nanosecond (default if no digit specified)
+     *
+     * However, note that timeval only supports microsecond precision
+     * (thus, formatting with higher-than-microsecond precision will
+     * always result in trailing zeroes), and IRC server-time is specified
+     * in millisecond precision (thus formatting received timestamps with
+     * higher-than-millisecond precision will always result in trailing
+     * zeroes).
+     */
+    static CString FormatTime(const timeval& tv, const CString& sFormat,
+                              const CString& sTZ);
     static CString FormatServerTime(const timeval& tv);
     static timeval ParseServerTime(const CString& sTime);
     static SCString GetTimezones();
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -499,6 +499,67 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
     return s;
 }
 
+CString CUtils::FormatTime(const timeval& tv, const CString& sFormat,
+                           const CString& sTimezone) {
+    // Parse additional format specifiers before passing them to
+    // strftime, since the way strftime treats unknown format
+    // specifiers is undefined.
+    CString sFormat2;
+
+    // Make sure %% is parsed correctly, i.e. %%L is passed through to
+    // strftime to become %L, and not 123.
+    bool bInFormat = false;
+    int iDigits;
+    CString::size_type uLastCopied = 0, uFormatStart;
+
+    for (CString::size_type i = 0; i < sFormat.length(); i++) {
+        if (!bInFormat) {
+            if (sFormat[i] == '%') {
+                uFormatStart = i;
+                bInFormat = true;
+                iDigits = 9;
+            }
+        } else {
+            switch (sFormat[i]) {
+                case '0': case '1': case '2': case '3': case '4':
+                case '5': case '6': case '7': case '8': case '9':
+                    iDigits = sFormat[i] - '0';
+                    break;
+                case 'L':
+                    iDigits = 3;
+                    // fall-through
+                case 'N': {
+                    int iVal = tv.tv_usec;
+                    int iDigitDelta = iDigits - 6; // tv_user is in 10^-6 seconds
+                    for (; iDigitDelta > 0; iDigitDelta--)
+                        iVal *= 10;
+                    for (; iDigitDelta < 0; iDigitDelta++)
+                        iVal /= 10;
+                    sFormat2 += sFormat.substr(uLastCopied,
+                        uFormatStart - uLastCopied);
+                    CString sVal = CString(iVal);
+                    sFormat2 += CString(iDigits - sVal.length(), '0');
+                    sFormat2 += sVal;
+                    uLastCopied = i + 1;
+                    bInFormat = false;
+                    break;
+                }
+                default:
+                    bInFormat = false;
+            }
+        }
+    }
+
+    if (uLastCopied) {
+        sFormat2 += sFormat.substr(uLastCopied);
+        return FormatTime(tv.tv_sec, sFormat2, sTimezone);
+    } else {
+        // If there are no extended format specifiers, avoid doing any
+        // memory allocations entirely.
+        return FormatTime(tv.tv_sec, sFormat, sTimezone);
+    }
+}
+
 CString CUtils::FormatServerTime(const timeval& tv) {
     CString s_msec(tv.tv_usec / 1000);
     while (s_msec.length() < 3) {
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -506,8 +506,8 @@ CString CUtils::FormatTime(const timeval& tv, const CString& sFormat,
     // specifiers is undefined.
     CString sFormat2;
 
-    // Make sure %% is parsed correctly, i.e. %%L is passed through to
-    // strftime to become %L, and not 123.
+    // Make sure %% is parsed correctly, i.e. %%f is passed through to
+    // strftime to become %f, and not 123.
     bool bInFormat = false;
     int iDigits;
     CString::size_type uLastCopied = 0, uFormatStart;
@@ -517,7 +517,7 @@ CString CUtils::FormatTime(const timeval& tv, const CString& sFormat,
             if (sFormat[i] == '%') {
                 uFormatStart = i;
                 bInFormat = true;
-                iDigits = 9;
+                iDigits = 3;
             }
         } else {
             switch (sFormat[i]) {
@@ -525,10 +525,7 @@ CString CUtils::FormatTime(const timeval& tv, const CString& sFormat,
                 case '5': case '6': case '7': case '8': case '9':
                     iDigits = sFormat[i] - '0';
                     break;
-                case 'L':
-                    iDigits = 3;
-                    // fall-through
-                case 'N': {
+                case 'f': {
                     int iVal = tv.tv_usec;
                     int iDigitDelta = iDigits - 6; // tv_user is in 10^-6 seconds
                     for (; iDigitDelta > 0; iDigitDelta--)
--- a/test/UtilsTest.cpp
+++ b/test/UtilsTest.cpp
@@ -117,3 +117,46 @@ TEST(UtilsTest, ServerTime) {
     }
     tzset();
 }
+
+TEST(UtilsTest, FormatTime) {
+    struct timeTest {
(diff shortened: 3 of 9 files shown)


src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
...
+ *   - %3N - millisecond
+ *   - %6N - microsecond
+ *   - %9N - nanosecond (default if no digit specified)
+ *
+ * However, note that timeval only supports microsecond precision
+ * (thus, formatting with higher-than-microsecond precision will
+ * always result in trailing zeroes), and IRC server-time is specified
+ * in millisecond precision (thus formatting received timestamps with
+ * higher-than-millisecond precision will always result in trailing
+ * zeroes).
```
psychon: So, is there any reason to make nanosecond precision the default? Or even a supported option?

src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
...
+    // specifiers is undefined.
+    CString sFormat2;
+
+    // Make sure %% is parsed correctly, i.e. %%L is passed through to
+    // strftime to become %L, and not 123.
+    bool bInFormat = false;
+    int iDigits;
+    CString::size_type pLastCopied = 0, pFormatStart;
+
+    for (CString::size_type i = 0; i < sFormat.length(); i++) {
```
psychon: Basically fine by me, but why not do a loop based on `.find('%')`? I'm not totally sure, but that might be easier/simpler.

src/Utils.cpp
CyberShadow: The reason is to stick to an existing formatting syntax instead of proliferating them further by diverging or inventing our own.

src/Utils.cpp
CyberShadow: I don't see how that would be easier or simpler considering the constraints (see comments).

src/Utils.cpp
DarthGandalf: Why do you want at least millisecond precision for format string?

src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
...
+                    iDigits = sFormat[i] - '0';
+                    break;
+                case 'L':
+                    iDigits = 3;
+                    // fall-through
+                case 'N': {
+                    int iVal = tv.tv_usec;
+                    int iDigitDelta = iDigits - 6; // tv_user is in 10^-6 seconds
+                    for (; iDigitDelta > 0; iDigitDelta--)
+                        iVal *= 10;
```
DarthGandalf: This is likely to trigger signed overflow, which is undefined behavior in C++

test/UtilsTest.cpp
```diff
@@ -117,3 +117,46 @@ TEST(UtilsTest, ServerTime) {
...
+TEST(UtilsTest, FormatTime) {
+    struct timeTest {
+        int sec, usec;
+        CString sResultL, sResultN, sResult3N;
+    };
+    timeTest aTimeTests[] = {
+        {42,  12345, "42.012", "42.012345000", "42.012"}, // leading zeroes
+        {42, 999999, "42.999", "42.999999000", "42.999"}, // (no) rounding
+        {42,      0, "42.000", "42.000000000", "42.000"}, // no tv_usec part
+    };
```
DarthGandalf: https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#value-parameterized-tests

src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
...
 
+/* Supports additional format specifiers for formatting sub-second values:
+ *
+ * - %L - millisecond, 3 digits (ruby extension)
+ * - %N - sub-second fraction (ruby extension)
+ *   - %3N - millisecond
+ *   - %6N - microsecond
+ *   - %9N - nanosecond (default if no digit specified)
+ *
+ * However, note that timeval only supports microsecond precision
```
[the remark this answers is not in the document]
CyberShadow: ZNC already uses `timeval` in a number of places. I'm not sure adding more types would be beneficial here.

src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
...
+ * in millisecond precision (thus formatting received timestamps with
+ * higher-than-millisecond precision will always result in trailing
+ * zeroes).
+ */
+CString CUtils::FormatTime(timeval tv, const CString& sFormat,
+                           const CString& sTimezone) {
+    // Parse additional format specifiers before passing them to
+    // strftime, since the way strftime treats unknown format
+    // specifiers is undefined.
+    CString sFormat2;
```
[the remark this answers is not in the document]
CyberShadow: I had that in my initial implementation, then removed it. Reserving memory here would mean we're always doing at least one allocation (assuming small-string optimization doesn't kick in of course), which we're trying to avoid if it's not necessary.

src/Utils.cpp
CyberShadow: The possible range of valid values cannot result in an overflow. It's only possible if the timeval was invalid.

test/UtilsTest.cpp
CyberShadow: This test case is based on the one in `UserTest.cpp`.

I do not see any existing instances of parametrized tests in ZNC's codebase.

include/znc/Utils.h
```diff
@@ -79,6 +79,8 @@ class CUtils {
     static CString CTime(time_t t, const CString& sTZ);
     static CString FormatTime(time_t t, const CString& sFormat,
                               const CString& sTZ);
+    static CString FormatTime(timeval tv, const CString& sFormat,
```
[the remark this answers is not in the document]
CyberShadow: Done

src/Utils.cpp
```diff
@@ -499,6 +499,82 @@ CString CUtils::FormatTime(time_t t, const CString& sFormat,
     return s;
 }
 
+/* Supports additional format specifiers for formatting sub-second values:
```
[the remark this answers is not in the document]
CyberShadow: Done

test/UtilsTest.cpp
DarthGandalf: That's not an excuse :P
UserTest.cpp can be fixed as well... though not in this pull request, of course.

src/Utils.cpp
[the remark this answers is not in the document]
DarthGandalf: It also uses timespec, though not in API.
It didn't use std::chrono, because it didn't exist. Now C++11 is required to compile ZNC anyway.

CyberShadow:
@DarthGandalf I looked at value-parametrized tests and I don't see a way to apply them constructively here. Their intended use seems to be with classes which implement a common interface (i.e. descending from an abstract superclass). So, either I've misunderstood the documentation and their provided examples, or your suggestion is inapplicable. Could you please elaborate?

Other than that, is there anything left for me to do here?

DarthGandalf:
@CyberShadow https://github.com/DarthGandalf/znc/commit/fd8a36d3ea202a7e00786b674d59343be59079df

> Other than that, is there anything left for me to do here?

I don't insist, but consider using timespec or std::chrono. I agree that timeval is already presented in other ZNC APIs, but I don't like that the format accepts %N which only can result in zeroes, without ability to pass value with enough precision for the result to not have zeros at the end.
As another solution, we can leave only %L and remove support for %N (which is not fully supported in this PR anyway)

CyberShadow:
> @CyberShadow DarthGandalf/znc@fd8a36d

Aah, thanks! Cherry-picked that here.

It didn't seem to result in an improvement in code clarity, but I trust your judgement and experience with gtest that there is some benefit to this approach that I'm not seeing.

> I don't insist, but consider using timespec or std::chrono.

Sure, I don't mind adding an overload. Shall I add an overload for one or both?

I see that the only use of `timespec` in the code base is currently guarded by a detected host feature (`HAVE_CLOCK_GETTIME`). Would the same apply to the overload?

DarthGandalf:
> It didn't seem to result in an improvement in code clarity

It's a more common pattern in gtest usage. What you wrote is more common for Go though. While they are approximately the same in isolation, I think following common patterns is generally good.

That reminds me that I need to cleanup integration test from some very uncommon patterns (`Z;`) and upgrade to a newer gtest version which allows `EXPECT_EQ(actual, expected)`.

> Shall I add an overload for one or both?

I just want `%N` to work as expected.

CyberShadow:
> I just want %N to work as expected.

OK. Here's my take on the situation:

Ideally there would be one type which indicates a point in time used throughout the ZNC code base. Currently this role is being fulfilled by a mix of `time_t` and `timeval`, however the issue presented here that the formatting syntax introduced by this pull request (same as Ruby's strftime function) allows specifying a precision higher than the one representable by either of those types.

The two alternatives presented were `timespec` and `std::chrono::time_point`. It's not clear whether `timespec` can be used today in the ZNC API: though it is specified as part of C11, it's not obvious whether C11 support is an agreed-upon requirement for building ZNC. For instance, ZNC's documentation only mentions the minimum compiler version, however `time.h` is part of the system's standard library, and in the case of glibc, was introduced in version 2.16. Considering also that the currently only occurrence of `timespec` is guarded by a define that's enabled by detecting its presence on the host system, it doesn't strike be as a good candidate for using it in the API.

As for `std::chrono::time_point` - from what I saw, it appears to be a class template tied to a specific clock, meaning that its precision and epoch would depend on said clock. Aside using one of the standard clock types in `std::chrono`, the resolution and epoch of which are (AIUI) implementation-defined, it may be possible to define a custom clock for use in ZNC, however that seems like an over-complication at this point.

Yet another option would be to define a custom type for representing points in time for use in ZNC's codebase. This wouldn't be out of place as ZNC already defines a few custom types such as `CString`. Such a type could implicitly cast to `timeval`, `timespec` etc. if desired. Though IMO a sensible option, it's probably too out of scope for this PR.

In the end I decided to add an overload that takes a `time_t` representing the integer second part and an `unsigned int` for representing the sub-second part as nanoseconds. In the future, should `timespec` be deemed sufficiently universally available to be used without prior feature detection, it could be added as an overload as well.

Let me know what you think.

DarthGandalf:
Another option: `std::timespec`, which appears in C++17, so it would also need to be ifdef-protected for now.
Some parts of API are already in ifdef: threads, SSL.

Why do you need nanoseconds at all, why microseconds are not enough, and why need to follow Ruby?
Python supports microseconds with `%f`, `absl::FormatTime()` supports `%E9f` for nanoseconds, http://search.cpan.org/dist/Time-Moment/lib/Time/Moment.pod also uses `%f` or `%9f` for nanoseconds.

CyberShadow:
> Why do you need nanoseconds at all

They're not necessary. I just wanted to stick to an existing formatting scheme, which happens to support them.

> why microseconds are not enough

I think microseconds should be more than enough for any foreseeable uses of time formatting. One may argue that milliseconds are sufficient, but it's possible to come up with a few situations where they may be useful, such as measuring how long it takes to handle one event from a batch. E.g. `dmesg` prints timestamps with microsecond precision by default.

> and why need to follow Ruby?

I looked at how some languages or libraries extend strftime, and of all the options I saw, Ruby's seemed more sensible to follow in ZNC.

> Python supports microseconds with `%f`

I considered that, however Python does not allow formatting the time to millisecond precision (or any other precision than microseconds), and microsecond precision is excessive for most uses in ZNC. Giving the user more control over timestamp formatting also means that users could configure the timestamp format to match that of their existing logs (from an IRC client or otherwise).

> absl::FormatTime() supports `%E9f` for nanoseconds

I'm not familiar with that library. I assume you mean [this one](https://github.com/abseil/abseil-cpp)? Looking at the implementation, its time formatting code looks like a thin wrapper around Google's [cctz](https://github.com/google/cctz). Its `strftime` extensions are documented [here](https://github.com/google/cctz/blob/4f9776a310f4952454636363def82c2bf6641d5f/include/cctz/time_zone.h#L219-L224).

In any case, we could certainly use cctz's syntax. I don't see how that changes the situation much. The two pertinent differences I see are:
1) the library seems to support precisions up to femtoseconds, which is much lower than nanoseconds and unquestionably of no use to ZNC;
2) the "full fractional precision" format specifiers are variable width, which may not be a good choice for the purpose of formatting timestamps (in logs or as sent to clients), as they are going to cause 10% of lines' contents to be misaligned after the timestamp.

Let me know what you want to do.

CyberShadow:
> http://search.cpan.org/dist/Time-Moment/lib/Time/Moment.pod also uses `%f` or `%9f` for nanoseconds.

Sorry, forgot to get to that. It looks like a good option. None of the caveats above apply to it, so it seems like the best option so far. I'll change the formatting syntax to `%f` to match `Time::Moment` then?

DarthGandalf:
Ok. Then it doesn't need the nanosecond-taking overload at all, which simplifies API

CyberShadow:
Sorry for the delay. I have time to work on this again.

On closer look, the Perl formatting API is variable-width, which is not acceptable for ZNC's use. In fact, the Perl library does not have a way to force fixed-width formatting - if the sub-second fraction just so happens to be zero, nothing is printed. It also unconditionally includes the leading period in the output string, however some locales use other characters for decimal fractions, like the comma.

That, and considering that it's not even an official Perl API, but a poorly designed third-party library, really leads me to doubt whether it's a precedent worth following.

I don't suppose I can convince you to just accept one of the implementations from one of the current or previous iterations of this pull request?

CyberShadow:
OK, it's become clear that we likely won't find a sufficiently suitable / satisfying precedent to re-use.

I changed the implementation to use a simple `%f` / `%#f` scheme, similar to the Perl one. I think that should satisfy outstanding issues in previous implementations.

CyberShadow:
Thanks!

PR #1461, opened 2017-11-21
Avoid calling OnWho on every channel in the network WHO was called in.

weabot:
Unless there's a very good, specific reason for it, I think this behaviour should be avoided. I thought of keeping the old behaviour if the channel wasn't valid, but I don't see the point unless there's something I'm not seeing, which might be likely.

Many IRC clients call WHO on every channel to update and format nick lists and such, and this dramatically reduces performance during such requests, to the point where I had over 10s latency to normal channel messages. Some tests I did pointed me to the WHO query where I found this.

From the tests I've done with this patch, WHO works as expected, including calling it on a channel or on a potential nick including wildcards. No signs of the issue I was having before either.

Even if there's a reason for this behaviour, doing it after more sanity checks, queueing, scheduled WHO calls, etc. would be a much better solution for what this could possibly accomplish.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -856,15 +856,13 @@ bool CIRCSock::OnNumericMessage(CNumericMessage& Message) {
             m_pNetwork->SetIRCNick(m_Nick);
             m_pNetwork->SetIRCServer(sServer);
 
-            const vector<CChan*>& vChans = m_pNetwork->GetChans();
-
-            for (CChan* pChan : vChans) {
-                pChan->OnWho(sNick, sIdent, sHost);
-            }
-
             CChan* pChan = m_pNetwork->FindChan(sChan);
-            if (pChan && pChan->IsDetached()) {
-                return true;
+
+            if (pChan) {
+                pChan->OnWho(sNick, sIdent, sHost);
+                if (pChan->IsDetached()) {
+                    return true;
+                }
             }
 
             break;


psychon:
Uhm. From your description (and the title of this pull request) it sounds like you think that previously ZNC sent a `WHO` request for every channel (when ??? happened) and with your patch it only sends a `WHO` for.... some channel? Also, I only looked at the patch after reading your description and the patch did not feel to "match" to the description. Basically, you got me confused.

For example:

> I don't see the point of this when it actually brings exponential overhead to the channel count.

Where exactly is there anything exponential in here? I am 99% sure that this piece of code has linear complexity in the number `m_pNetwork->GetChans().size()`.

Since you seem to be implying that ZNC was slow in handling 352 numerics with many channels: Could you run `znc --debug` (this prints timestamps) and show the difference with and without your patch? There should be a (large) gap to the next timestamp if handling 352's is slow. This does not have to be your main znc, I guess, you could set up a different instance just for testing that joins many channels (and I guess many large channels).

Oh and: What exactly is a large number of channels?

psychon:
>  By default, OnWho only updates some info on the nick inside of ZNC I believe, so it's not that bad. A module adding onto it would make it much worse, and consider the case of channel hoarders. 

There is no `OnWho` module call and `CChan::OnWho` is not something that modules can hook into. So modules do not make a difference here.

Anyway, I'm not convinced that this patch makes much of a difference (your ZNC must freeze for seconds whenever any client logs in to it and gets a buffer replay!), but this is also not my decision to make.

psychon:
@DarthGandalf What's your understanding of what this patch does and why it is a good idea?

DarthGandalf:
When response to WHO arrives, it contains channel name. Previous code updated all channels, now it updates only this channel. It shouldn't affect performance, but why do you think this patch makes behavior worse?

psychon:
A nick can only have one ident and hostname. Yes, you can query this information per-channel, but it is still global. For example, if the client supports UHNAMES, but the IRC server does not, then AFAIR "passive snooping of `WHO` replies" is the only way that ZNC can figure out the ident and host for the UHNAMES replies.
So, doesn't this basically make it more likely that ZNC has outdated ident / host information of a nick and sends it to clients in `CChan::AttachUser`?

DarthGandalf:
Ok :) But let's protect this by a comment. The code without the comment looked like someone forgot that there is a channel, and wrote the loop anyway (and I checked this loop was there since the initial SVN commit, so commit message didn't tell much either).

PR #1474, opened 2017-12-19
Add "AuthOnlyViaModule" global/user setting

fwilson42:
Setting BuiltinAuthDisabled on a user causes CheckPass to never return true, causing all authentication attempts using the configured password to fail, both on IRC connections and for webadmin. This is useful in situations where an external module (cyrusauth, certauth, imapauth) handles authentication.

Setting the global BuiltinAuthDisabled option causes similar behavior across every user. If BuiltinAuthDisabled is set to true globally, it cannot be overridden per-user.

### DIFF
--- a/include/znc/User.h
+++ b/include/znc/User.h
@@ -150,6 +150,7 @@ class CUser {
     void SetTimestampFormat(const CString& s) { m_sTimestampFormat = s; }
     void SetTimestampAppend(bool b) { m_bAppendTimestamp = b; }
     void SetTimestampPrepend(bool b) { m_bPrependTimestamp = b; }
+    void SetBuiltinAuthDisabled(bool b) { m_bBuiltinAuthDisabled = b; }
     void SetTimezone(const CString& s) { m_sTimezone = s; }
     void SetJoinTries(unsigned int i) { m_uMaxJoinTries = i; }
     void SetMaxJoins(unsigned int i) { m_uMaxJoins = i; }
@@ -185,6 +186,7 @@ class CUser {
     bool IsAdmin() const;
     bool DenySetBindHost() const;
     bool MultiClients() const;
+    bool BuiltinAuthDisabled() const;
     const CString& GetStatusPrefix() const;
     const CString& GetDefaultChanModes() const;
     /** How long must an IRC connection be idle before ZNC sends a ping */
@@ -250,6 +252,7 @@ class CUser {
     bool m_bBeingDeleted;
     bool m_bAppendTimestamp;
     bool m_bPrependTimestamp;
+    bool m_bBuiltinAuthDisabled;
 
     CUserTimer* m_pUserTimer;
 
--- a/include/znc/User.h
+++ b/include/znc/User.h
@@ -150,7 +150,7 @@ class CUser {
     void SetTimestampFormat(const CString& s) { m_sTimestampFormat = s; }
     void SetTimestampAppend(bool b) { m_bAppendTimestamp = b; }
     void SetTimestampPrepend(bool b) { m_bPrependTimestamp = b; }
-    void SetBuiltinAuthDisabled(bool b) { m_bBuiltinAuthDisabled = b; }
+    void SetOnlyModulesMayAuth(bool b) { m_bOnlyModulesMayAuth = b; }
     void SetTimezone(const CString& s) { m_sTimezone = s; }
     void SetJoinTries(unsigned int i) { m_uMaxJoinTries = i; }
     void SetMaxJoins(unsigned int i) { m_uMaxJoins = i; }
@@ -186,7 +186,7 @@ class CUser {
     bool IsAdmin() const;
     bool DenySetBindHost() const;
     bool MultiClients() const;
-    bool BuiltinAuthDisabled() const;
+    bool OnlyModulesMayAuth() const;
     const CString& GetStatusPrefix() const;
     const CString& GetDefaultChanModes() const;
     /** How long must an IRC connection be idle before ZNC sends a ping */
@@ -252,7 +252,7 @@ class CUser {
     bool m_bBeingDeleted;
     bool m_bAppendTimestamp;
     bool m_bPrependTimestamp;
-    bool m_bBuiltinAuthDisabled;
+    bool m_bOnlyModulesMayAuth;
 
     CUserTimer* m_pUserTimer;
 
--- a/include/znc/User.h
+++ b/include/znc/User.h
@@ -150,7 +150,7 @@ class CUser {
     void SetTimestampFormat(const CString& s) { m_sTimestampFormat = s; }
     void SetTimestampAppend(bool b) { m_bAppendTimestamp = b; }
     void SetTimestampPrepend(bool b) { m_bPrependTimestamp = b; }
-    void SetOnlyModulesMayAuth(bool b) { m_bOnlyModulesMayAuth = b; }
+    void SetAuthOnlyViaModule(bool b) { m_bAuthOnlyViaModule = b; }
     void SetTimezone(const CString& s) { m_sTimezone = s; }
     void SetJoinTries(unsigned int i) { m_uMaxJoinTries = i; }
     void SetMaxJoins(unsigned int i) { m_uMaxJoins = i; }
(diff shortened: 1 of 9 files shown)


modules/controlpanel.cpp
```diff
@@ -442,6 +446,10 @@ class CAdminMod : public CModule {
             bool b = sValue.ToBool();
             pUser->SetTimestampAppend(b);
             PutModule("AppendTimestamp = " + CString(b));
+        } else if (sVar == "builtinauthdisabled") {
+            bool b = sValue.ToBool();
+            pUser->SetBuiltinAuthDisabled(b);
+            PutModule("BuiltinAuthDisabled = " + CString(b));
```
psychon: Is it a good idea for users to be able to edit this setting for themselves? I'd think that only admin should be able to change this setting.

modules/webadmin.cpp
```diff
@@ -299,6 +299,8 @@ class CWebAdminMod : public CModule {
             WebSock.GetParam("appendtimestamp").ToBool());
         pNewUser->SetTimestampPrepend(
             WebSock.GetParam("prependtimestamp").ToBool());
+        pNewUser->SetBuiltinAuthDisabled(
+            WebSock.GetParam("builtinauthdisabled").ToBool());
```
psychon: Same as above: Should users be able to edit this for themselves or should this be treated similar to `IsAdmin`?

include/znc/User.h
```diff
@@ -150,6 +150,7 @@ class CUser {
     void SetTimestampFormat(const CString& s) { m_sTimestampFormat = s; }
     void SetTimestampAppend(bool b) { m_bAppendTimestamp = b; }
     void SetTimestampPrepend(bool b) { m_bPrependTimestamp = b; }
+    void SetBuiltinAuthDisabled(bool b) { m_bBuiltinAuthDisabled = b; }
```
DarthGandalf: Let's don't have negation in setting names. Not double negation is not more confusing for human mind.

modules/webadmin.cpp
```diff
@@ -1567,6 +1569,13 @@ class CWebAdminMod : public CModule {
                 o12["Checked"] = "true";
             }
 
+            CTemplate& o13 = Tmpl.AddRow("OptionLoop");
+            o13["Name"] = "builtinauthdisabled";
+            o13["DisplayName"] = t_s("Built-in Authentication Disabled");
```
DarthGandalf: What does Built-in mean? I know what you mean by this, but someone who's looking at the UI can e.g. mistake this by "enable anonymous login without any password"... and what @psychon said makes this mistake even more dangerous.

psychon:
@DarthGandalf I don't have much of an opinion on that question either. This PR adds a global setting while a new hashing scheme would be purely per user...

modules/webadmin.cpp
```diff
@@ -1567,6 +1569,13 @@ class CWebAdminMod : public CModule {
                 o12["Checked"] = "true";
             }
 
+            CTemplate& o13 = Tmpl.AddRow("OptionLoop");
```
DarthGandalf: I think this option should be near 'change password' block, not hidden in random list of checkboxes below.

DarthGandalf:
Another random idea: probably ZNC should prevent users shooting themselves, and allow setting this only if some module which provides another authentication is loaded already. How to implement this cleanly is another question...

include/znc/User.h
fwilson42: Is "BuiltinAuthEnabled" (defaulting to true) okay instead?

modules/webadmin.cpp
fwilson42: That's a good point, I don't have many better ideas on this. Would making this admin-only and documenting it on the wiki be an acceptable fix?

fwilson42:
@DarthGandalf on preventing users from shooting themselves, would a new `CModInfo` field `m_bProvidesAuth` be sufficient? Detecting whether or not another authentication method is available would then be pretty trivial.

fwilson42:
Also, re:alternative hash type, one thing to consider is that would mean overwriting the password with no way to get it back in case someone wanted to re-enable builtin authentication. That plus the ability to disable it globally makes this option a little better (but I'm perhaps a bit biased 😛).

include/znc/User.h
fwilson42: Thinking about it a little more, maybe "OnlyModulesAuth" (still defaults to false) is more descriptive... thoughts?

modules/webadmin.cpp
fwilson42: Renaming to "OnlyModulesAuth" and having help text like "Only allow authentication with a module" would fix this problem as well.

fwilson42:
Alright, now:
- changing it per-user is admin-only
- it's under the Authentication section on the add/edit user page instead of Flags
- it's called `OnlyModulesMayAuth` and has clearer/better labeling in the UI
- there's a test.

The only thing that's missing is the foot-shooting prevention part, which might not be needed since the setting is now admin-only and has better labeling.

modules/webadmin.cpp
```diff
@@ -210,6 +210,13 @@ class CWebAdminMod : public CModule {
             pNewUser->SetPass(sHash, CUser::HASH_DEFAULT, sSalt);
         }
 
+        sArg = WebSock.GetParam("builtinauthdisabled");
+        if(spSession->IsAdmin()) {
+            if(!sArg.empty()) {
+                pNewUser->SetBuiltinAuthDisabled(sArg.ToBool());
+            }
+        }
```
psychon: Doesn't this need `else if (pUser) { pNewUser->SetBuildinAuthDisabled(pUser->BuiltinAuthDisabled()); }`? When a user is edited, instead of resetting this setting to false, it should be preserved. Right?
(I'm not sure if the above is necessary, but e.g. BindHost is handled in a similar way below)

psychon:
Left a minor comment, but I think this is mainly for @DarthGandalf to decide.

modules/webadmin.cpp
fwilson42: Yeah, I think you're right. I didn't run into this while testing but will push something just in case. If it ends up being unnecessary I'll go ahead and revert.

test/UserTest.cpp
```diff
@@ -119,3 +119,29 @@ TEST_F(UserTest, IsHostAllowed) {
...
+    // user-level only
+    user.SetOnlyModulesMayAuth(true);
+    EXPECT_FALSE(user.CheckPass("password"));
+
+    // re-enabling built-in authentication
+    user.SetOnlyModulesMayAuth(false);
+    EXPECT_TRUE(user.CheckPass("password"));
+
+    // on at global level, off at user level
+    CZNC::Get().SetOnlyModulesMayAuth(true);
```
DarthGandalf: Please set this global variable to its default value in the end of test, to not accidentally affect other tests which are in the same process later.
Something like
```c++
bool bDefault = CZNC::Get().GetOnlyModules();
...
CZNC::Get().Set(bDefault);
```
Bonus points if done in an exception-safe manner.

DarthGandalf:
> would a new CModInfo field m_bProvidesAuth be sufficient?

It might be.
There are so many ways to configure authentication. Some of them can coexist with built-in passwords, if only some users are coming from external (LDAP?) auth. Another interesting example: certauth. Mere presence of that module doesn't mean that a given user configured their certificate. In general, `CModInfo::m_bProvidesAuth` isn't enough to tell for sure whether the module is able to authenticate a given user.
Alternatively, add a new callback "hey module, do you have an ability to authenticate this user?", which might e.g. go to an SQL database and check whether there is a row in the Users table. But how would imapauth implement that?

fwilson42:
That's a good point, even `cyrusauth` couldn't implement that in all cases, I think (in saslauthd mode it definitely couldn't, in auxprop mode I'm not sure), and that's 2/3 of the currently available auth modules. Given that this is admin-only functionality, detecting if a user can actually authenticate against a loded module is pretty difficult, and recovery from this is fairly simple (shutdown ZNC, edit config, start ZNC again), it might not even be worth adding anti-foot-shooting functionality.

Anyway, all other changes you requested have been made.

Issue #1430, opened 2017-07-25
certauth: Should accept authentication to webadmin through a trusted proxy


00dani:
The certauth module allows users to log into ZNC's webadmin using their client certificate, if the webadmin is accessed over TLS. However, one common configuration of ZNC involves running the webadmin behind a [reverse proxy](http://wiki.znc.in/Reverse_Proxy) such as nginx, and it is impossible to "forward" a client's certificate through the proxy to ZNC.

A workaround, which many backend servers support, is to optionally accept a client certificate in the form of an HTTP header, rather than requiring the certificate to have actually been used for the TLS negotiation. Unfortunately, certauth does not currently support this feature.

I propose that, if certauth is validating a request from a declared `TrustedProxy`, it check for an HTTP header such as `X-Client-Fingerprint`, and use that fingerprint if present rather than require a client certificate to be presented in the request. This change would make it trivial to pass through a previously verified client certificate from nginx:
```nginx
ssl_verify_client optional_no_ca;
location / {
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Client-Fingerprint $ssl_client_fingerprint;
  proxy_pass https://[::1]:7878; # this is where my webadmin's listening ;)
}
```
Of course, if the request to ZNC did *not* come from a `TrustedProxy`, it's not safe to trust a fingerprint presented as a header, and in that case the header should be ignored.

PR #1478, opened 2017-12-23
[WIP] Allow trusted proxies to forward client certificate fingerprints (#1430)

fwilson42:
These changes allow trusted proxies to set a header overriding the detected client certificate fingerprint. This allows webadmin behind a HTTPS-supporting proxy to correctly work with the certauth module.

WIP because I'm *really* uncomfortable with the `dynamic_cast` that's used here. The alternatives are to change types in a bunch of other places, or just make `Csock::GetPeerFingerprint` virtual. I think the last one is the best solution, and if you agree I can get a PR off to CSocket and fix this one up.

Aside from the nasty `dynamic_cast` I'd be interested in any feedback.


include/znc/HTTPSock.h
```diff
@@ -86,6 +86,7 @@ class CHTTPSock : public CSocket {
     const CString& GetURI() const;
     const CString& GetURIPrefix() const;
     bool IsPost() const;
+    long GetPeerFingerprint(CS_STRING& sResult) const override;
```
DarthGandalf: It's `CString` here.

src/HTTPSock.cpp
```diff
@@ -181,6 +188,12 @@ void CHTTPSock::ReadLine(const CString& sData) {
             // X-Forwarded-For list in both cases use it as the endpoind
             m_sForwardedIP = sIP;
         }
+    } else if (sName.Equals("X-Forwarded-CertFP:")) {
+        bool bTrusted = IsTrustedProxy(GetRemoteIP());
+        if(bTrusted) {
```
DarthGandalf: space missing
you can use clang-format to fix style with a single keypress, and not care about indentation and other things manually

src/HTTPSock.cpp
```diff
@@ -181,6 +188,12 @@ void CHTTPSock::ReadLine(const CString& sData) {
             // X-Forwarded-For list in both cases use it as the endpoind
             m_sForwardedIP = sIP;
         }
+    } else if (sName.Equals("X-Forwarded-CertFP:")) {
+        bool bTrusted = IsTrustedProxy(GetRemoteIP());
```
DarthGandalf: `if (IsTrusted...)` is fine. The new variable doesn't add anything here.

src/HTTPSock.cpp
```diff
@@ -181,6 +188,12 @@ void CHTTPSock::ReadLine(const CString& sData) {
...
+        if(bTrusted) {
+            m_sForwardedCertFP = sLine.Token(1, true);
+            DEBUG("Got a forwarded CertFP '" << m_sForwardedCertFP << "'");
```
DarthGandalf: You might want to output a socket name here.

src/HTTPSock.cpp
```diff
@@ -181,6 +188,12 @@ void CHTTPSock::ReadLine(const CString& sData) {
             // X-Forwarded-For list in both cases use it as the endpoind
             m_sForwardedIP = sIP;
         }
+    } else if (sName.Equals("X-Forwarded-CertFP:")) {
```
DarthGandalf: I can't find anything about this name.

an-empty-string: Yeah, it's not standardized anywhere. X-Forwarded-For is standardized by convention but as far as I know there isn't anything similar to TLS fingerprints, as this sort of thing isn't as widely done. This would have to go in documentation somewhere.

an-empty-string:
I'll work on a test later; I'm having some trouble figuring out a good strategy for testing this since things like `IsTrustedProxy` are network-dependent.

DarthGandalf:
Check out integration test, it uses network.
With virtual function it'd be cleaner, yes. But what hash algorithm is used
by this new http header? I wonder whether it's possible to change the
algorithn somehow. I am a bit wary that result of this function changes
over time but probably that should be fine.
Also it somewhat turns hash into password. Previously, if trusted proxy is
hacked, attacker could only circumvent whitelist of ips, and fail2ban. But
now would gain complete access to a user account if they learnt
fingerprint, which is not supposed to be secret.

23 дек. 2017 г. 8:57 PM пользователь "Empty String" <
notifications@github.com> написал:

I'll work on a test later; I'm having some trouble figuring out a good
strategy for testing this since things like IsTrustedProxy are
network-dependent.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/znc/znc/pull/1478#issuecomment-353747793>, or mute the
thread

an-empty-string:
The hash algorithm is sha1 and it can't be changed, but this is the case with normal certauth currently also.

That's a good point about a compromised TrustedProxy. How about a global config flag that enables/disables this behavior?

I'll take a look at the integration test stuff, that sounds like what I want.

PR #1492, opened 2018-02-09
Stop creating files with odd permissions.

citrus-it:
bits are being set in file mode rather than open flags. As a result file permissions get strange bits set.

```
build# ls -l /etc/opt/znc/znc.conf
-rwS------   1 znc      znc        1.62K Feb  9 14:47 /etc/opt/znc/znc.conf
build# stat -c %a /etc/opt/znc/znc.conf
4600
```

### DIFF
--- a/src/FileUtils.cpp
+++ b/src/FileUtils.cpp
@@ -352,11 +352,11 @@ bool CFile::Open(int iFlags, mode_t iMode) {
     }
 
     // We never want to get a controlling TTY through this -> O_NOCTTY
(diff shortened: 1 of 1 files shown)


DarthGandalf:
This code is broken since ZNC 0.064, but I never saw it causing these weird permissions. What OS is this, and what are values of O_NOCTTY and O_BINARY?

citrus-it:
It's OmniOS (illumos distribution)
O_BINARY is 0, O_NOCTTY is 0x800

Issue #1230, opened 2016-01-09
NickServ: detect when network supports SASL and suggest loading the SASL module instead


Mikaela:
It would make the module more useful if it could discover what networks supoort sasl during capability negotiation or performing cap ls (maybe even 3.2 to see what the mechanisms are) and suggest loading the sals lodule instead saying that it identifies you before being visible to the network (so a lot faster than nickserv does). 

I think it should be automatic without switch to disable the behaviour except that it could be quiet if sasl was also loaded.

GLolol:
> I think it should be automatic without switch to disable the behaviour except that it could be quiet if sasl was also loaded.

If I'm interpreting this correctly, automatic `*sasl` setup when `*nickserv` password is set would be nice. But how would ZNC guess the SASL username?

Personally I prefer having both modules loaded in a setup so initial connections use SASL, but `*nickserv` will still identify me in cases where services go down and lose my login status.

Mikaela:
> If I'm interpreting this correctly, automatic *sasl setup when *nickserv password is set would be nice. But how would ZNC guess the SASL username?

No, only telling *This network supports SASL so you might want to load the SASL module in order to identify before you are visible on the network (as opposed to this module which often identifies you after you are already joined to channels).*

> Personally I prefer having both modules loaded in a setup so initial connections use SASL, but *nickserv will still identify me in cases where services go down and lose my login status.

This is what I mean with *I think it should be automatic without switch to disable the behaviour except that it could be quiet if sasl was also loaded.* or in other words, always say that unless SASL is also loaded.

GLolol:
> or in other words, always say that unless SASL is also loaded.

Couldn't it just be announced the first time the module's loaded?

Mikaela:
> Couldn't it just be announced the first time the module (*nickserv) is loaded?

No as the user might not be connected to the network at that time and network that doesn't have SASL today could get SASL tomorrow.

GLolol:
Or the first time the SASL cap is received but `*sasl` isn't loaded... I disagree with having the notice always on, it seems too naggy to me.

Mikaela:
Naggy is the point and there is no reason to nag when sasl is loaded so it can well do it until thatm

kylef:
> Personally I prefer having both modules loaded in a setup so initial connections use SASL, but *nickserv will still identify me in cases where services go down and lose my login status.

[...]

PR #1496, opened 2018-02-18
Add authtoservices module, replacing sasl/nickserv (#1494)

### LINKED ISSUE #1494: Combine nickserv and sasl modules
They do the same thing - authenticate user on network. Why people have to choose which of them to use? Especially, when they don't know what SASL is, they are more likely to enable nickserv instead of sasl, even if networks supports SASL.

Anything I'm missing?

an-empty-string:
The authtoservices module is an amalgamation of the nickserv and sasl modules. In short, it:
- Tries SASL authentication, if available on the network (this can be disabled with /msg \*authtoservices TrySasl no)
- If that fails, it'll respond to NickServ prompts to identify, like the nickserv module would (this can be disabled with /msg \*authtoservices RequireSasl yes)

The first time it is loaded, it will automatically migrate configuration from the sasl module or the nickserv module (precedence is given to sasl), if available. This ensures there's an upgrade path, if we decide the best way forward is to remove the sasl/nickserv modules from the tree and have them load authtoservices instead.

Aside from a few name changes and what's described above, it functions identically to the nickserv and sasl modules.

This PR fixes #1494.

### DIFF
--- a/modules/authtoservices.cpp
+++ b/modules/authtoservices.cpp
@@ -1,0 +1,491 @@
+/*
+ * Copyright (C) 2004-2018 ZNC, see the NOTICE file for details.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <znc/Modules.h>
+#include <znc/User.h>
+#include <znc/IRCNetwork.h>
+#include <znc/IRCSock.h>
+
+class Mechanisms : public VCString {
+  public:
+    void SetIndex(unsigned int uiIndex) { m_uiIndex = uiIndex; }
+
+    unsigned int GetIndex() const { return m_uiIndex; }
+
+    bool HasNext() const { return size() > (m_uiIndex + 1); }
+
+    void IncrementIndex() { m_uiIndex++; }
+
+    CString GetCurrent() const { return at(m_uiIndex); }
+
+    CString GetNext() const {
+        if (HasNext()) {
+            return at(m_uiIndex + 1);
+        }
+
+        return "";
+    }
+
+  private:
+    unsigned int m_uiIndex = 0;
+};
+
+class CAuthToServices : public CModule {
+    const struct {
+        const char* szName;
+        CDelayedTranslation sDescription;
+        bool bDefault;
+    } SupportedMechanisms[2] = {
+        {"EXTERNAL", t_d("TLS certificate, for use with the *cert module"),
+         true},
+        {"PLAIN", t_d("Plain text negotiation, this should work always if the "
+                      "network supports SASL"),
+         true}};
+
+  public:
+    MODCONSTRUCTOR(CAuthToServices) {
+        AddCommand("Help", t_d("search"), t_d("Generate this output"),
+                   [=](const CString& sLine) { PrintHelp(sLine); });
+        AddCommand("SetAccount", t_d("account"),
+                   t_d("Set your services account name"),
+                   [=](const CString& sLine) { SetAccountCommand(sLine); });
+        AddCommand("ClearAccount", "", t_d("Clear your services account name"),
+                   [=](const CString& sLine) { ClearAccountCommand(sLine); });
+        AddCommand("SetPassword", t_d("password"),
+                   t_d("Set your services password"),
+                   [=](const CString& sLine) { SetPasswordCommand(sLine); });
+        AddCommand("ClearPassword", "", t_d("Clear your services password"),
+                   [=](const CString& sLine) { ClearPasswordCommand(sLine); });
+        AddCommand("SetNSCommand", t_d("cmd new-pattern"),
+                   t_d("Set pattern for NickServ commands"),
+                   [=](const CString& sLine) { SetNSCommandCommand(sLine); });
+        AddCommand("ViewNSCommands", "",
+                   t_d("Show patterns for NickServ commands"),
+                   [=](const CString& sLine) { ViewNSCommandsCommand(sLine); });
+        AddCommand("SetNSName", t_d("nickname"),
+                   t_d("Set NickServ name (Useful on networks like EpiKnet, "
+                       "where NickServ is named Themis"),
+                   [=](const CString& sLine) { SetNSNameCommand(sLine); });
+        AddCommand("ClearNSName", "",
+                   t_d("Reset NickServ name to default (NickServ)"),
+                   [=](const CString& sLine) { ClearNSNameCommand(sLine); });
+        AddCommand("SaslMechanism", t_d("[mechanism[ ...]]"),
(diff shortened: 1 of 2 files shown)


modules/authtoservices.cpp
```diff
@@ -0,0 +1,491 @@
...
+        return result.ReadFromDisk(sPath) == MCString::MCS_SUCCESS;
+    }
+
+    bool OnLoad(const CString& sArgs, CString& sMessage) override {
+        if (!sArgs.empty() && sArgs != "<hidden>") {
+            SetNV("Password", sArgs);
+            SetArgs("<hidden>");
+        }
+
+        if(!GetNV("NotFirstLoad").ToBool()) {
```
DarthGandalf: This module should be automatically loaded if sasl and/or nickserv was loaded before. That can be done in IRCNetwork.cpp (search for "legacy crap"). Also the migrated settings can mimic the previous settings more closely if you take into account which of those modules were loaded.

Also that could help getting rid of "NotFirstLoad" hack.

an-empty-string:
The most recent commits likely still have a few bugs... expect updates, but I'm about to start another busy week so it might be a while before I can get them in.

an-empty-string:
502d4dd fixes an issue where the account name wouldn't properly default to the nickname during SASL authentication. That's the only problem I observed when testing this.

src/IRCNetwork.cpp
```diff
@@ -369,6 +369,21 @@ struct TOption {
     void (CIRCNetwork::*pSetter)(T);
 };
 
+bool CIRCNetwork::ModuleInConfig(CConfig* pConfig, const CString& sModule) const {
+    VCString vsList;
+    bool bFound;
```
DarthGandalf: This is undefined behavior without `= false`
Compiler is free to optimize this whole function to `return true;`

src/IRCNetwork.cpp
```diff
@@ -369,6 +369,21 @@ struct TOption {
...
+
+    pConfig->FindStringVector("loadmodule", vsList);
```
DarthGandalf: It removes all "loadmodule"s from config.

src/IRCNetwork.cpp
```diff
@@ -450,8 +468,43 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
...
+            }
+
+            if (sModName == "nickserv") {
+                if (ModuleInConfig(pConfig, "sasl")) {
+                    CUtils::PrintAction(
+                        "NOTICE: [nickserv] was merged into [authtoservices], "
+                        "which will load using your [sasl] configuration. Not "
+                        "loading [nickserv]");
+                    continue;
+                } else if (ModuleInConfig(pConfig, "authtoservices")) {
```
DarthGandalf: How did you test this?

src/IRCNetwork.cpp
```diff
@@ -450,8 +468,43 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
...
+                        "loading that with [sasl]'s configuration instead";
+                    sModName = "authtoservices";
+                    sArgs = "as_sasl";
+                }
+            }
+
+            if (sModName == "nickserv") {
+                if (ModuleInConfig(pConfig, "sasl")) {
+                    CUtils::PrintAction(
+                        "NOTICE: [nickserv] was merged into [authtoservices], "
```
DarthGandalf: What if sasl module was loaded, but not configured - e.g. user never told sasl the password.

modules/authtoservices.cpp
```diff
@@ -0,0 +1,499 @@
...
+    // NickServ authentication
+
+    void NickServAuthenticate() {
+        MCString msValues;
+        msValues["account"] = GetAccount();
+        msValues["password"] = GetNV("Password");
+        PutIRC(CString::NamedFormat(GetNV("IdentifyCmd"), msValues));
+    }
+
+    void HandleMessage(CNick& Nick, const CString& sMessage) {
```
DarthGandalf: Need a way to disable this, e.g. via additional "mechanism"

Zarthus:
it's bikesheddy, but I don't know if `authtoservices` is a good module name

(1) sasl doesn't depend on services, it's an ircd feature
(2) ircds are working to get rid of services support and integrate authentication directly

I have no really good name proposals, but i think just `authentication` or `auth` (already taken i think?) might suit better.

modules/authtoservices.cpp
```diff
@@ -0,0 +1,499 @@
...
+    } SupportedMechanisms[2] = {
+        {"EXTERNAL", t_d("TLS certificate, for use with the *cert module"),
+         true},
+        {"PLAIN", t_d("Plain text negotiation, this should work always if the "
+                      "network supports SASL"),
+         true}};
+
+  public:
+    MODCONSTRUCTOR(CAuthToServices) {
+        AddCommand("Help", t_d("search"), t_d("Generate this output"),
```
Zarthus: Isn't there a `AddHelpCommand();` function?

Zarthus:
I realize the PR feedback here is perhaps better suited for improvement tickets in the future; I just gave thoughts where I thought they fit, in the end it's remants from the old code.

modules/authtoservices.cpp
```diff
@@ -0,0 +1,499 @@
...
+        bool bDefault;
+    } SupportedMechanisms[2] = {
+        {"EXTERNAL", t_d("TLS certificate, for use with the *cert module"),
+         true},
+        {"PLAIN", t_d("Plain text negotiation, this should work always if the "
+                      "network supports SASL"),
+         true}};
+
+  public:
+    MODCONSTRUCTOR(CAuthToServices) {
```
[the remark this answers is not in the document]
Zarthus: The settings can use the same variables, but if I want to configure SASL I just want to know what I need to configure SASL, introducing more ("redundant") options for a nicer to use "API" is, to me, more beneficial than keeping it a confusing mess with 50 commands that don't follow any kind of unified pattern.

modules/authtoservices.cpp
kylef: :+1:, this can also be a security flaw on sasl only networks where there isn't a NickServ nick. It might be possible for a user to get sasl passwords for ZNC users (and other clients that have similar logic) because they can change their nick to `NickServ` and send one of messages below to get clients to send passwords to them. Some servers can offer blocking nick names out from use like NickServ and common typos, but that is relying on the server being configured correctly.

I think the spirit of what we're trying to do with #1494 / #1230 is to encourage users to use sasl over nickserv (instead of, not with) because it has many benefits, including the decrease in risks where passwords can be erroneously taken. If its not by automated tooling, even there are known impersonation attacks in the wild with bots using nicknames similar to nickserv such as nicksarv and others to trick users into sending them their passwords.

If a network offers SASL I don't think users should consider using nickserv to idenify. Even if the sasl capability is currently missing or unavailable during connection, there can be room for more sophisticated attacks where you can abuse this feature of clients while service may be offline or down for maintenance. I'd avoid the room for downgrade attacks from sasl to nickserv auth. The original sasl module also has `requireauth` option to disconnect if we can't authenticate before registration (providing upstream server supports capabilities).

---

The problem right now is that some users do not know what sasl is, and they are already familiar with nickserv and thus would setup the nickserv module instead. I would suggest that we go with an approach that allows the module to be configured to only allow sasl authentication, and if at any time the module determines it can "upgrade" from from nickserv to sasl auth, it should then set this flag so it would never downgrade without user consent. Or even as alternative to merging the modules together, make the current NickServ module fail if it detects the server supports sasl and facilitate the user on switching to the sasl module instead (and some way to continue with nickserv if the user really knows what they are doing and understands the risks). However this approach may be a bit more user hostile than people would like.

PR #1505, opened 2018-04-02
simple_away: Add TimeZone to the away string subst

Zarthus:
If TimeZone is not set GetUser()->GetTimezone() returns an empty string ("") so nothing gets appended.

Currently the away time (on the same day) is completely useless if you don't know what TimeZone the user is from.

Mild backwards incompatibility if you have a custom string that adds the timezone, but not much we can do about that.

### DIFF
--- a/modules/simple_away.cpp
+++ b/modules/simple_away.cpp
@@ -211,7 +211,8 @@ class CSimpleAway : public CModule {
         if (sReason.empty()) sReason = SIMPLE_AWAY_DEFAULT_REASON;
 
         time_t iTime = time(nullptr);
-        CString sTime = CUtils::CTime(iTime, GetUser()->GetTimezone());
+        CString sTime = CUtils::CTime(iTime, GetUser()->GetTimezone()) + " " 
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Alternatively, use UTC by default which is probably better for privacy: now you're telling everyone which city user is in.

Zarthus:
Added new PR while we decide which solution is better (conflicts with #1506)

teward:
The proper approach would seem to me to serve this timestamp as UTC always, or get the timezone offset for the time zone that is selected and serve that rather than serve the city-style timestamp that is selectable in the ZNC interface, in my opinion.

Consider that if I choose any of the America timestamps in the USA which has a city, the timestamp could change based on daylight savings time, so UTC-5 is US Eastern Standard Time vs. UTC-4 US Eastern Daylight Time.  If we are unable to get this offset easily (though it should be trivial to get this), then we should serve UTC as a standardized response.

Zarthus:
https://github.com/znc/znc/pull/1506#issuecomment-377927449 was merged, this conflicts with it.

PR #1506, opened 2018-04-02
simple_away: Convert to UTC time

Zarthus:
- convert to utc time
- add tz to the away string subst

Currently the away time (on the same day) is completely useless if you don't know what TimeZone the user is from.

Mild backwards incompatibility if you have a custom string that adds the timezone, but not much we can do about that.

Conflict with #1505

### DIFF
--- a/modules/simple_away.cpp
+++ b/modules/simple_away.cpp
@@ -211,7 +211,7 @@ class CSimpleAway : public CModule {
         if (sReason.empty()) sReason = SIMPLE_AWAY_DEFAULT_REASON;
 
         time_t iTime = time(nullptr);
-        CString sTime = CUtils::CTime(iTime, GetUser()->GetTimezone());
+        CString sTime = CUtils::CTime(iTime, "Etc/UTC") + " UTC";
         sReason.Replace("%awaytime%", sTime);
         sReason = ExpandString(sReason);
         sReason.Replace("%s", sTime);  // Backwards compatibility with previous


modules/simple_away.cpp
```diff
@@ -211,7 +211,7 @@ class CSimpleAway : public CModule {
         if (sReason.empty()) sReason = SIMPLE_AWAY_DEFAULT_REASON;
 
         time_t iTime = time(nullptr);
-        CString sTime = CUtils::CTime(iTime, GetUser()->GetTimezone());
+        CString sTime = CUtils::CTime(iTime, "Etc/UTC") + " UTC";
```
DarthGandalf: Is "Etc/UTC" different from "GMT"? Also `gmtime_r` would be cleaner than this.

teward: @DarthGandalf https://www.timeanddate.com/time/gmt-utc-time.html

> Although GMT and UTC share the same current time in practice, there is a basic difference between the two:
> - GMT is a time zone officially used in some European and African countries. The time can be displayed using both the 24-hour format (0 - 24) or the 12-hour format (1 - 12 am/pm).
> - UTC is not a time zone, but a time standard that is the basis for civil time and time zones worldwide. This means that no country or territory officially uses UTC as a local time.

Zarthus: Etc/UTC is what the user would have specified in their GetTimezone() configuration.

PR #1511, opened 2018-04-05
Remove 'BoyOh' and ' - ZNC' in messages

precla:
BoyOh sound very childish and is not needed.
Same goes for the appending of ' - ZNC' to the nick, no need for that one.
Maybe make those optional instead of forced?

### DIFF
--- a/modules/sample.cpp
+++ b/modules/sample.cpp
@@ -91,11 +91,11 @@ class CSampleMod : public CModule {
     }
 
     void OnIRCConnected() override {
-        PutModule(t_s("You got connected BoyOh."));
+        PutModule(t_s("You got connected."));
     }
 
     void OnIRCDisconnected() override {
-        PutModule(t_s("You got disconnected BoyOh."));
+        PutModule(t_s("You got disconnected."));
     }
 
     EModRet OnIRCRegistration(CString& sPass, CString& sNick, CString& sIdent,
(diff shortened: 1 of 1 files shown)


modules/sample.cpp
```diff
@@ -91,16 +91,15 @@ class CSampleMod : public CModule {
...
     }
 
     void OnIRCDisconnected() override {
-        PutModule(t_s("You got disconnected BoyOh."));
+        PutModule(t_s("You got disconnected."));
     }
 
     EModRet OnIRCRegistration(CString& sPass, CString& sNick, CString& sIdent,
                               CString& sRealName) override {
-        sRealName += " - ZNC";
```
DarthGandalf: This is a sample module - to show what can be done in a module, and how approximately to do it. With `return CONTINUE` only, this callback does nothing, and should be removed. Maybe replace this line with something more useful if you don't like appending "-ZNC" to realname?

magnum7385:
I thought the entire reason it didn't sound professional was to encourage changing it.  It's sample.cpp after all.

teward:
@magnum7385 The idea behind the "sample" module was to show you various things on how to *create* a very basic module.  The specific wording of things inside it is 'irrelevant' in that the module is not supposed to be used, it's supposed to be used as a 'basic' guideline for how a module could be written.  It's therefore not really 'relevant' whether the wording in it is 'professional' or not.  (As the module is not supposed to really be used for anything, really, other than an example of code)

PR #1513, opened 2018-04-12
Fix echo-message regression from 1.6.x

throwaway-user:
This patch prevents PRIVMSGs from clients from being echoed to other clients if their target has the status prefix. This matches the behaviour in 1.6.x.


throwaway-user:
Without this patch, with two clients connected:

clientA sends PRIVMSG *status :help
znc sends help text back to clientA
clientB echoes the PRIVMSG to *status, but gets no response.

With this patch:

clientA sends PRIVMSG *status :help
znc sends help text back to clientA
clientB sees nothing.

This behaviour is particularly obvious when using the third party znc-playback module, as reconnect clients will cause connected ones to get spammed with messages to *playback.

PR #1524, opened 2018-05-06
Make some changes to sample module (redo of #1511)

teward:
***This module was written as an example, and should be treated as such***.

It details the various types of things that a module can do; however it is not really a "true" end-user-usable module in that it's not really designed to be "professionally written" like the rest of the modules.

To that end, #1511 tried to make it more professional, but removed a bit of code that was supposed to be kept.  This was done intentionally so that you could show examples of how modules could be used.

This commit fixes the issues in 1511 but doesn't remove the "RealName" suffix appending example.  It instead puts " - Running on ZNC" as the example string.

It also adds a comment to the beginning stating that the module as-is is more an example of things you can do with the module, rather than a truly useful end-user module in its current state.

### DIFF
--- a/modules/sample.cpp
+++ b/modules/sample.cpp
@@ -14,6 +14,13 @@
  * limitations under the License.
  */
 
+/*
+ * NOTE: This module's primary purpose is to provide a sample of the various
+ * things that modules can do. It is not written to be an actual module that
+ * you would enable on a sample installation, and serves more as a developer 
+ * "example" rather than an actual module that would be used.
+ */
+
 #include <znc/Client.h>
 #include <znc/Chan.h>
 #include <znc/Modules.h>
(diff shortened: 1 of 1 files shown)


modules/sample.cpp
```diff
@@ -24,7 +31,7 @@ using std::vector;
 class CSampleJob : public CModuleJob {
   public:
     CSampleJob(CModule* pModule)
-        : CModuleJob(pModule, "sample", "Message the user after a delay") {}
+        : CModuleJob(pModule, "sample", "An example module that messages the user after a delay.") {}
```
DarthGandalf: This is not description of the module

teward:
Looks like my forked repo went away, not sure how or why.  Will resubmit this PR later once I figure out what happened...

PR #1550, opened 2018-06-04
added Bahasa Indonesia

eggoez:
Dropping by crowdin project

### DIFF
--- a/translations/id-ID
+++ b/translations/id-ID
@@ -1,0 +1,1 @@
+SelfName Indonesian


translations/id-ID
```diff
@@ -0,0 +1 @@
+SelfName Indonesian
```
DarthGandalf: Not "Bahasa Indonesia"?

eggoez: we people indonesia say Bahasa Indonesia, but word say Indonesian in list

PR #1554, opened 2018-06-10
Fix build without SSL support

baruchsiach:
The <memory> headers is needed for unique_ptr even when SSL is not
enabled. This fixes the following build failure:

src/Utils.cpp: In static member function 'static bool CUtils::CheckCIDR(const CString&, const CString&)':
src/Utils.cpp:674:5: error: 'unique_ptr' is not a member of 'std'
     std::unique_ptr<addrinfo, decltype(deleter)> aiHost(aiHostC, deleter);
     ^


DarthGandalf:
Thanks, applied to 1.7.x branch

Issue #1370, opened 2017-01-20
Allow setting DEBUG mode during runtime.


Zarthus:
Sometimes ZNC fails during runtime and you're not sure why. ZNC doesn't log, and DEBUG() is your only hope of resolution.

Allowing DEBUG mode to be turned on during runtime would allow you to troubleshoot this a bit better, especially as "failed to write to config" errors aren't uncommon and znc is great at telling you absolutely nothing useful at all without DEBUG. [example here](https://github.com/znc/znc/blob/2fcde9f2e111e23c15c4941a1ed0248944e6d7a5/src/znc.cpp#L457-L615)

Piping DEBUG() through *status while it's enabled for bonus points.

PR #1556, opened 2018-06-13
Add module that dynamically enables/disables debug mode

Zarthus:
Add very basic module that enables / disables Debug mode. The current idea is a MVP wherein it directly mirrors the PR changes in #1547, where improvements can be added later.

Supersedes: https://github.com/znc/znc/pull/1547

Known "issues":
* Only supports logging to stdout right now. 
  * Could add writing to logfile later, syslog, as a adminonly broadcast, or to the person who enabled it. At the moment I just directly mirrored #1547 
* ~~Translations don't support printf format, I have no idea how to get reliable translations done.~~
* ~~Anyone can load the module, presumably. (see https://github.com/znc/znc/pull/1556#pullrequestreview-128499826)~~
* No wiki page yet. Will be added whenever it's considered ready for merging.
* Would be nice to set the NV variable to the ZNC account name rather than nick. Need to look into correct method name.

Closes #1547
References #1370

(Doesn't close 1370 because I think in order for this issue to be resolved to full satisfaction at least one other method of logging should exist.)

### DIFF
--- a/modules/admindebug.cpp
+++ b/modules/admindebug.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2004-2018 ZNC, see the NOTICE file for details.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <znc/FileUtils.h>
+#include <znc/Server.h>
+#include <znc/IRCNetwork.h>
+#include <znc/User.h>
+#include <znc/ZNCDebug.h>
+
+class CAdminDebugMod : public CModule {
+  public:
+    MODCONSTRUCTOR(CAdminDebugMod) {
+        AddHelpCommand();
+        AddCommand("Enable", "", t_d("Enable Debug Mode"),
+                   [=](const CString& sLine) { CommandEnable(sLine); });
+        AddCommand("Disable", "", t_d("Disable Debug Mode"),
+                   [=](const CString& sLine) { CommandDisable(sLine); });
+        AddCommand("Status", "", t_d("Show the Debug Mode status"),
+                   [=](const CString& sLine) { CommandStatus(sLine); });
+    }
+
+    void CommandEnable(const CString& sCommand) {
+        if (GetUser()->IsAdmin() == false) {
+            PutModule(t_s("Access denied!"));
+            return;
+        }
+
+        ToggleDebug(true, GetUser()->GetNick());
+    }
+
+    void CommandDisable(const CString& sCommand) {
+        if (GetUser()->IsAdmin() == false) {
+            PutModule(t_s("Access denied!"));
+            return;
+        }
+
+        ToggleDebug(false, GetNV("enabled_by"));
+    }
+
+    bool ToggleDebug(bool bEnable, CString sEnabledBy) {
+        bool bValue = CDebug::Debug();
+
+        if (bEnable == bValue) {
+            if (bEnable) {
+                PutModule(t_s("Already enabled."));
+            } else {
+                PutModule(t_s("Already disabled."));
+            }
+            return false;
+        }
+
+        CDebug::SetDebug(bEnable);
+        CString sEnabled = CString(bEnable ? "on" : "off");
+        CZNC::Get().Broadcast(CString(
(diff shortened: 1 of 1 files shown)


modules/admindebug.cpp
```diff
@@ -0,0 +1,98 @@
...
+        AddCommand("Enable", "", t_d("Enable Debug Mode"),
+                   [=](const CString& sLine) { CommandEnable(sLine); });
+        AddCommand("Disable", "", t_d("Disable Debug Mode"),
+                   [=](const CString& sLine) { CommandDisable(sLine); });
+        AddCommand("Status", "", t_d("Show the Debug Mode status"),
+                   [=](const CString& sLine) { CommandStatus(sLine); });
+    }
+
+    bool OnLoad(const CString& sArgs, CString& sMessage) override {
+        if (GetUser() != nullptr) {
```
Zarthus: Probably is always NULL. Could be removed, but I added it for futureproofing.

modules/admindebug.cpp
```diff
@@ -0,0 +1,98 @@
...
+            return GetUser()->IsAdmin();
```
psychon: How about an error message explaining why loading was denied? Also, have you tested this? I think global modules are loaded with `GetUser() == nullptr`.

Zarthus: >Anyone can load the module, presumably. (see #1556 (review))
The idea was to deny loading of the module unless it's on a global level by an administrator, GetUser() always seemed to be null. It doesn't accomplish anything, I was just trying to satisfy the "OnLoad returns a bool" condition, because it seems pointless to always return true for a module mainly designed for admins.

DarthGandalf: You don't do it to "satisfy some weird requirement"; that requirement is there for a reason - to block loading the module if needed. You can simply drop the whole `OnLoad`, it already returns `true` by default.
This module is global; it can't be loaded by anyone other than admins anyway.

> Anyone can load the module, presumably. (see #1556 (review))

What exactly are you linking at?

modules/admindebug.cpp
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        ToggleDebug(false, GetNV("enabled_by"));
+    }
+
+    bool ToggleDebug(bool bEnable, CString sEnabledBy) {
+        bool bValue = CDebug::Debug();
+
+        if (bEnable == bValue) {
+            PutModule(t_s("Already " + CString(bEnable ? "enabled" : "disabled") + "."));
```
DarthGandalf: This isn't possible to translate, even though it does say "t_s".
```c++
if (...) {
  PutModule(t_s("enabled..."));
} else {
  PutModule(t_s("disabled..."));
}
```

modules/admindebug.cpp
```diff
@@ -0,0 +1,98 @@
...
+        CString sEnabled = CString(bEnable ? "on" : "off");
+        CZNC::Get().Broadcast(CString(
+            "An administrator has just turned Debug Mode \02" + sEnabled + "\02. It was enabled by \02" + sEnabledBy + "\02."
+        ));
+        if (bEnable) {
+            CZNC::Get().Broadcast(
+                CString("Messages, credentials, and other sensitive data may"
+                " become exposed to the host during this period.")
+            );
+            SetNV("enabled_by", sEnabledBy);
```
DarthGandalf: So it reads `enabled_by` from NV, then sets it back to the same value?
What's the point of `enabled_by` in NV?

Zarthus: (1) To show who enabled it when you 'Disable'
(2) Forward compatibility, for when you want to log to the existing ZNC user in the future. This is why my PR summary suggests that "Would be nice to set the NV variable to the ZNC account name rather than nick. Need to look into correct method name."

modules/admindebug.cpp
Zarthus: The previous code before I test it was `t_s("Already %s.", CString(...))`, but t_s doesn't support printf syntax. I will change this.

modules/admindebug.cpp
Zarthus: @psychon 
Remnants from https://github.com/znc/znc/blob/master/modules/adminlog.cpp#L42-L58 (because that's the module I copied from.) I will change.

@DarthGandalf I was linking to https://github.com/znc/znc/pull/1556#discussion_r195183328

modules/admindebug.cpp
DarthGandalf: Why does it need to be persistent? A member field is not enough?

modules/admindebug.cpp
DarthGandalf: With printf-like syntax, it would be `t_f("Already {1}.")(CDebug::Debug())`, but that's harder to translate - if you provide a full sentence, the translation can be more natural, e.g. not in every language "does" and "does not" can be represented using the same structure.

modules/admindebug.cpp
Zarthus: Should be enough, I just didn't think of using it.

DarthGandalf:
@Zarthus please https://wiki.znc.in/Admindebug

PR #1571, opened 2018-06-17
Implement Happy Eyeballs. Fix IPv4/IPv6 in a better way.

### LINKED ISSUE #241: RFC 6555 - Happy Eyeballs
It would be nice to have it implemented, e.g. for case when bindhost is not set.

http://tools.ietf.org/html/rfc6555
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1308-rfc-6555-happy-eyeballs?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DarthGandalf:
Close #241


DarthGandalf:
The problem with this approach: if server sends something immediately upon connect, that data is delivered to the temporary socket class, and it's lost. Looking how to solve this...

Gunni:
This ticket is not a blocker for IPv6 support, all it means is that if a client gets a dual stack DNS result, it will prefer IPv6 by default on most systems.

If IPv6 is broken on the path, then, yes, the connection will time out and ipv4 will be tried eventually.

But that's not your fault, that's the fault of, the server, the ISPs between, or the user.

Maybe just make it clear in the documentation that we don't have Happy Eyeballs yet and to either yolo with dual stack dns records, or, make a primary/dual, v4, v6 record seperately and give that to users.

DarthGandalf:
@Gunni I've replied on znc/znc-docker#19

That said, when I finish #1639, it'll be easy to fix ZNC->server side. Failure to implement this feature with current sockets we have caused me to start refactoring it. If only I had more time to finish #1639 faster...

psychon:
> The problem with this approach: if server sends something immediately upon connect, that data is delivered to the temporary socket class, and it's lost. Looking how to solve this...

Without having looked at any code: Does `PauseRead()` and `UnpauseRead()` when connection is established help? (I remember CSocket having such methods...)

DarthGandalf:
AFAIR, no, it didn't help.

PR #1573, opened 2018-06-27
SASL Authentication for Clients

MrLenin:
Here's a peek at where this code is at, still needs to handle too many authentication attempts, but is otherwise fairly complete. PLAIN mechanism has been moved to a rudimentary module since the last time you've seen it. I wasn't sure if there was a more proper way to add the SASL mechanisms to the CAP LS output, so I just hacked it in for now. Any suggestions are welcome. I will be further testing module behavior for more complex mechanism types shortly.

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -115,6 +115,10 @@ class CClient : public CIRCSocket {
           m_bBatch(false),
           m_bEchoMessage(false),
           m_bSelfMessage(false),
+          m_bSasl(false),
+          m_bSaslAuthenticating(false),
+          m_bSaslAuthenticated(false),
+          m_bSaslMultipart(false),
           m_bPlaybackActive(false),
           m_pUser(nullptr),
           m_pNetwork(nullptr),
@@ -123,6 +127,8 @@ class CClient : public CIRCSocket {
           m_sUser(""),
           m_sNetwork(""),
           m_sIdentifier(""),
+          m_sSaslBuffer(""),
+          m_sSaslMechanism(""),
           m_spAuth(),
           m_ssAcceptedCaps(),
           m_ssSupportedTags(),
@@ -150,6 +156,7 @@ class CClient : public CIRCSocket {
                {true, [this](bool bVal) { m_bAccountNotify = bVal; }}},
               {"extended-join",
                {true, [this](bool bVal) { m_bExtendedJoin = bVal; }}},
+              {"sasl", {false, [this](bool bVal) { m_bSasl = bVal; m_bSaslAuthenticating = bVal; }}},
           }) {
         EnableReadLine();
         // RFC says a line can have 512 chars max, but we are
@@ -326,6 +333,10 @@ class CClient : public CIRCSocket {
     unsigned int DetachChans(const std::set<CChan*>& sChans);
 
     bool OnActionMessage(CActionMessage& Message);
+    void OnAuthenticateMessage(CAuthenticateMessage& Message);
+
+    CString EnumerateSaslMechanisms(SCString& ssMechanisms);
+
     bool OnCTCPMessage(CCTCPMessage& Message);
     bool OnJoinMessage(CJoinMessage& Message);
     bool OnModeMessage(CModeMessage& Message);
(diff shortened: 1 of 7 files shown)


src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+
+		return;
+	}
+
+	if ((m_bSaslMultipart && !sMessage.Equals("+"))) {
+		m_sSaslBuffer.append(sMessage);
+		m_bSaslMultipart = false;
+	} else if (!m_bSaslMultipart && !sMessage.Equals("+")) {
+		m_sSaslBuffer.assign(sMessage);
+	}
```
DarthGandalf:
```c++
if (sMessage != "+") {
    m_sSaslBuffer += sMessage;
}
```
Is `m_bSaslMultipart` needed at all?

src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+        m_sSaslMechanism = "";
+        m_sSaslBuffer = "";
+        m_bSaslMultipart = false;
+    };
+
+    auto SaslChallenge = [this](CString sChallenge) {
+        sChallenge.Base64Encode();
+        auto sChallengeSize = sChallenge.length();
+
+        if (sChallengeSize > uiMaxSaslMsgLength) {
```
DarthGandalf: Is this condition necessary?

src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+            for (auto i = 0u; i < sChallengeSize; i += uiMaxSaslMsgLength) {
+                CString sMsgPart = sChallenge.substr(i, uiMaxSaslMsgLength);
+                PutClient("AUTHENTICATE " + sMsgPart);
+            }
+        } else {
+            PutClient("AUTHENTICATE " + sChallenge);
+        }
+    };
```
DarthGandalf: If last message is 400, need to send another one with `+`

src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+            SaslChallenge(sChallenge);
+        } else {
+            PutClient("AUTHENTICATE +");
+        }
+		return;
+	}
+
+	if (sBufferSize == uiMaxSaslMsgLength) {
+		m_bSaslMultipart = true;
+		m_sSaslBuffer.append(sMessage);
```
DarthGandalf: Need to prevent the buffer growing too much by a sequence of multiple 400-bytes messages.

src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+	if ((m_bSaslMultipart && !sMessage.Equals("+"))) {
+		m_sSaslBuffer.append(sMessage);
+		m_bSaslMultipart = false;
+	} else if (!m_bSaslMultipart && !sMessage.Equals("+")) {
+		m_sSaslBuffer.assign(sMessage);
+	}
+
+	m_sSaslBuffer.Base64Decode();
+
+	auto sAuthcId = m_sUser;
```
DarthGandalf: `AUTHENTICATE` can happen before `USER`

src/Client.cpp
```diff
@@ -958,6 +989,149 @@ bool CClient::OnActionMessage(CActionMessage& Message) {
...
+					m_sSaslMechanism, m_sSaslBuffer, sAuthcId,
+                    sResponse, bAuthenticationSuccess),
+					&bResult);
+
+	if (bResult && !sResponse.empty()) {
+		SaslChallenge(sResponse);
+        return;
+	}
+
+	m_sSaslBuffer.clear();
```
DarthGandalf: move above; it should be cleared even if `bResult` was true

jonny3010:
Any new on this?

include/znc/Client.h
```diff
@@ -326,6 +333,10 @@ class CClient : public CIRCSocket {
     unsigned int DetachChans(const std::set<CChan*>& sChans);
 
     bool OnActionMessage(CActionMessage& Message);
+    void OnAuthenticateMessage(CAuthenticateMessage& Message);
+
+    CString EnumerateSaslMechanisms(SCString& ssMechanisms);
```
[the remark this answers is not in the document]
kylef:
```suggestion
    CString EnumerateSASLMechanisms(SCString& ssMechanisms);
```

Perhaps more of a question, given the code style below with `OnCTCPMessage`, the acronym is capitalised. Should we do the same here?

delthas:
Rebased and addressed comments in https://github.com/znc/znc/pull/1873

Issue #1580, opened 2018-07-21
admindebug causes crash on disable


A-UNDERSCORE-D:
Using ZNC 1.7.1, with znc launched in the default way (so forking).
Enabling the debug mode works, though the stdout file discriptor [cannot be found](https://github.com/znc/znc/blob/master/src/main.cpp#L502-L508)
On disable ZNC crashes with no logs in syslog
steps to reproduce (assuming default prefix for modules):

```
/znc loadmod adminlog
/msg *admindebug enable
/msg *admindebug disable
```

relevant logs from my crash:
```
[12:46:11] -*status- *** An administrator has just turned Debug Mode on. It was enabled by A_D. 
[12:46:11] -*status- *** Messages, credentials, and other sensitive data may become exposed to the host during this period. 
[12:49:04] -*status- *** An administrator has just turned Debug Mode off. It was enabled by A_D. 
[12:49:04] * Disconnected (Remote host closed socket).
```
Relevant logs from a discussion in #znc with @Zarthus:

```
<A_D> Zarthus: methinks I found a bug with your admindebug plugin
<Zarthus> if it has to do with the fact you just disconnected i'm not sure if that's a good thing :P
<A_D> yeahhhhhhhhhhh
<A_D> so I turned debug on went looking for the logs, which I couldnt find because it had forked and bad things happen to stdout when that happens
<A_D> and then I turned it off
<A_D> and then znc turned off
<Zarthus> so you ran it with debug on at first, or debug off at first?
<A_D> debug off at first, standard launch
<A_D> sec
<-- spcBNC has quit (Ping timeout: 244 seconds)
<Zarthus> the way i tested it without forking, so that kind of makes sense (I ran debug on and then played around with it)
<A_D> relavent logs: https://paste.snoonet.org/paste/3dfmp
<A_D> and the service file I use
<Zarthus> interesting
<A_D> I should add a pidfile to that >.>
<A_D> no loud noises in syslog
<A_D> what was my disconnect message here?
<A_D> connection closed by remote host?
<Zarthus> (12:49:04) * A_D (A_D@unaffiliated/ad/x-3954327) Quit (Remote host closed the connection)
<A_D> okay then it hard fell over
<A_D> well, not hard. it was nice enough to close sockets
<Zarthus> lol
<A_D> but not nice enough to run quit messages
<Zarthus> I'm not sure what the issue would be beyond "stdout is not there when you fork"
[...]

PR #1581, opened 2018-07-21
admindebug.cpp: #1580: Enforce need of TTY to turn on debug mode

Zarthus:
Prevents a crash from happening, but does not apply a workaround (e.g. opening a new file for stdout)


Zarthus:
Fixed now.

PR #1582, opened 2018-07-22
Add swap command to autoattach module

zchrykng:
This allows you do swap the position of two rules in the order

To make this work well, we needed to change how settings for the mofule
are stored. This includes a function and check to migrate existing rules over.


modules/autoattach.cpp
```diff
@@ -116,22 +118,36 @@ class CChanAttach : public CModule {
         }
     }
 
+    void HandleSwap(const CString& sLine) {
+        u_int iA = sLine.Token(1).ToUInt();
```
DarthGandalf: Where does this typedef come from?

modules/autoattach.cpp
```diff
@@ -116,22 +118,36 @@ class CChanAttach : public CModule {
...
+        u_int iB = sLine.Token(2).ToUInt();
+
+        Swap(iA, iB);
+    }
+
     void HandleList(const CString& sLine) {
+        int Index;
```
DarthGandalf: `iIndex`? please follow the style around...

modules/autoattach.cpp
```diff
@@ -116,22 +118,36 @@ class CChanAttach : public CModule {
...
             Table.AddRow();
+            Table.SetCell(t_s("Index"), to_string(Index));
             Table.SetCell(t_s("Neg"), it->IsNegated() ? "!" : "");
             Table.SetCell(t_s("Chan"), it->GetChans());
             Table.SetCell(t_s("Search"), it->GetSearch());
             Table.SetCell(t_s("Host"), it->GetHostMask());
+            Index += 1;
         }
 
+        delete Index;
```
DarthGandalf: WAT?

modules/autoattach.cpp
```diff
@@ -151,38 +167,39 @@ class CChanAttach : public CModule {
...
-        for (it = BeginNV(); it != EndNV(); ++it) {
-            CString sAdd = it->first;
-            bool bNegated = sAdd.TrimPrefix("!");
-            CString sChan = sAdd.Token(0);
-            CString sSearch = sAdd.Token(1);
-            CString sHost = sAdd.Token(2, true);
-
-            Add(bNegated, sChan, sSearch, sHost);
-        }
+        Save();
```
DarthGandalf: Why?

modules/autoattach.cpp
```diff
@@ -264,13 +289,61 @@ class CChanAttach : public CModule {
...
+        // Load our saved settings, ignore errors
+        MCString::iterator it;
+        for (it = BeginNV(); it != EndNV(); ++it) {
+            CString sAdd = it->first;
+
+            rules.push_back(sAdd);
+        }
+
+
+        for (VCString::const_iterator rit = rules.begin(); rit != rules.end();
```
DarthGandalf: Why not foreach loop?

modules/autoattach.cpp
zchrykng: At least on my machine it is part of the system headers. I was looking at how one of the other modules handled numeric input and copied it from there.

modules/autoattach.cpp
zchrykng: Will push a fix for this and other some of the other issues soon.

modules/autoattach.cpp
zchrykng: Sorry, my C++ is really rusty. Will push with fix soon.

modules/autoattach.cpp
zchrykng: Because I apparently can't read and misunderstood was some of what the original code was doing. Push coming soon.

test/integration/tests/modules.cpp
```diff
@@ -166,6 +166,10 @@ TEST_F(ZNCTest, AutoAttachModule) {
     ircd.Write(":foo PRIVMSG #znc :hello");
     ircd.ReadUntil("TEST");
     client.ReadUntil("hello");
+    client.Write("PRIVMSG *autoattach :Add * test *");
+    client.ReadUntil("Added to list");
+    client.Write("PRIVMSG *autoattach :Swap 1 2")
+    client.ReadUntil("Rules Swapped");
```
DarthGandalf: I can show you much easier implementation which will still pass this test:

```c++
AddCommand("Swap", "", "",
                   [=](const CString& sLine) { PutModule("Rules Swapped"); });
```

Can you code here the exact issue which you were facing without this new functionality in place? That's the point of having a test.

modules/autoattach.cpp
```diff
@@ -264,13 +276,59 @@ class CChanAttach : public CModule {
...
 
+    bool Swap(int Start, int End) {
+        // CAttachMatch
+        if (Start > m_vMatches.size() || Start <= 0 ||
+            End > m_vMatches.size() || End <= 0) {
+            PutModule(t_s("Illegal # Requested"));
+            return false;
+        } else {
+            std::iter_swap(m_vMatches.begin() + (Start - 1),
+                           m_vMatches.begin() + (End - 1));
```
DarthGandalf:
```c++
std::swap(matches[start], maches[end]);
```
?

modules/autoattach.cpp
```diff
@@ -264,13 +276,59 @@ class CChanAttach : public CModule {
...
 
-        DelNV(it->ToString());
         m_vMatches.erase(it);
 
+        Save();
+
         return true;
     }
 
+    bool Swap(int Start, int End) {
```
DarthGandalf: Something is wrong with naming of the variables. Where is the prefix?

test/integration/tests/modules.cpp
zchrykng: Okay, I am working on expanding this, but the thing I'm not sure how to test is the client correctly not attaching when the ignore rule is in place.

Issue #1253, opened 2016-03-16
Please allow the configuration to be spread over multiple files


mapreri:
my use case is very similar of the one in #975 where I'd like to have the main config file under git (possibly public), and the user stuff elsewhere, backed up separately.

atm I can't think of a sane way to have znc save to disk in a sensible way, unless another config option is added like `SeparatedUserConf = true` which removes the user and networks definition and move them to files named `<datadir>/config/users/*.conf`, but I'm not sure I'd call this sane.

Wolf480pl:
I think it'd make sense to move user configs to `users/<username>/user.conf`, and network configs to `users/<username>/networks/<networkname>/network.conf`. It'd make it easier to have the main config under confuguration management or version control, and it'd also make it easier to move users between ZNC instances, which IMO would be very nice to have.

PR #1583, opened 2018-07-22
Allow having a read-only config file

### LINKED ISSUE #975: Read-only config file(s)
I'd like to make my config files read-only. I use a provisioning tool to manage my servers, including ZNC configuration, and would ideally like to prevent a running ZNC process from making any changes to the files.

I thought I could achieve this by making the files read-only (`chmod 400`), but apparently ZNC [opens znc.conf in read+write mode even when only reading+parsing the file](https://github.com/znc/znc/blob/a314d3057efdb19d8259c7bf4160f31392418f7b/src/znc.cpp#L934), so that's not a possibility.

Wolf480pl:
* use a separate lockfile instead of locking the config file
* open the config read-only if we don't have write permissions
* don't try to overwrite the config if it was opened read-only

This should technically fix #975 for the narrow usecase described by that issue's OP.

Later I'm going to submit another pull request, taking care of #1253, which should take care of most usecases where part of the config is supposed to be provided externally, while the rest of it should be configurable through znc (eg. the users should still be allowed to edit their own settings, add networks, etc.)


Wolf480pl:
As suggested by MetaNova on #znc-dev, made it take a `--readonly` commandline argument, instead of trying to detect if config is read-only.

teward:
I'm not a dev, but **I'm going to give this a -1 based on my testing** and some undesired behavior/results coming from it.

The most prevalent issue is that when you try and issue `/znc shutdown` to gracefully stop the ZNC, when running in Read Only mode you get a hard error that prohibits shutdown process:

    [2018-07-23 16:14:35] <*status> ERROR: Writing config file to disk failed! Aborting. Use SHUTDOWN FORCE to ignore.

Also, you have a case where if you do anything that'd result in a config file write, it has the tendency to be able to spam the user with error messages (if they're an admin, I assume, but that's not tested).  You can end up with four messages rapidfired within a minute worth of "spam" to your notices tab or to your active screen if your client doesn't split them out into other tabs:

    [2018-07-23 16:14:04] -*status- *** Writing the config file failed
    [2018-07-23 16:14:10] -*status- *** Writing the config file failed
    [2018-07-23 16:14:14] -*status-*** Writing the config file failed
    [2018-07-23 16:14:20] -*status- *** Writing the config file failed

If we're running in read only mode, we *expect* that writing the config file will fail, or that it won't even attempt to write the config file, and therefore we shouldn't be getting these errors.

Further, if we're running in Read Only mode, we shouldn't have a hard failure when we attempt to use `/znc shutdown` to terminate the ZNC process.

My suggestions are:

(1) If we're running in Read Only mode, recognize that and don't crit-fail when you call `/znc shutdown`
(2) If running in Read Only mode, don't spam the admin/user about not being able to write the config file, as we don't expect that to succeed when in RO mode.

Wolf480pl:
@teward thanks for spotting this, I really should have tested my change better.
With the latest commit, the issues you've mentioned should be fixed.

I'd love if some dev looked that latest commit and told me if there's a better way to do it.
I'd rather have `WriteConfig()` return some tri-state (or more) enum, but haven't seen anything like that in ZNC codebase, and I'm reluctant to introduce such new pattern when the convention is to return `bool`.

src/znc.cpp
```diff
@@ -229,6 +229,15 @@ void CZNC::Loop() {
                 // stop pending configuration timer
                 DisableConfigTimer();
 
+                if (GetReadonlyConfig()) {
+                    if (eState == ECONFIG_NEED_VERBOSE_WRITE) {
+                        Broadcast("Writing the config skipped because "
```
DarthGandalf: This will regularly spam admins whenever e.g. "in-config" channel's key is changing...

src/znc.cpp
```diff
@@ -994,8 +994,43 @@ void CZNC::BackupConfigOnce(const CString& sSuffix) {
...
+    m_pLockFile = pLock;
+    return true;
+}
+
+bool CZNC::ParseConfig(const CString& sConfig, bool bReadonlyConfig, CString& sError) {
     m_sConfigFile = ExpandConfigPath(sConfig, false);
+    m_bReadonlyConfig = bReadonlyConfig;
+
+    if (m_bReadonlyConfig) {
+        CUtils::PrintError("Running in read-only mode. Any changes to the settings, users,");
```
DarthGandalf: Why is it an error?

DarthGandalf:
Does this mode also prevent writing of .registry files?

src/znc.cpp
Wolf480pl: I thought these changes trigger non-verbose config saves... I grepped for `ECONFIG_NEED_VERBOSE_WRITE` and the only place that sets it that I found was `SIGUSR1` handler.

src/znc.cpp
Wolf480pl: I wanted it to be a warning, but couldn't find anything like `PrintWarning`. I looked at what `--allow-root` does, and it prints its warning as an error.

Wolf480pl:
>Does this mode also prevent writing of .registry files?

No. I thought it'd be better if it doesn't affect the .registry files.

>Many module callbacks return EModRet.

So in this case it'd be `CONTINUE` for success, `HALTCORE` for read-only mode, and `HALT` for failure? If it's ok to use EModRet here, in something that isn't a module callback, stretching the semantics like that, then I guess I'll use it.

DarthGandalf:
> No. I thought it'd be better if it doesn't affect the .registry files.

Then some random changes will be saved, and some other changes will be not.

> If it's ok to use EModRet here, in something that isn't a module callback

That's not what I meant... You wanted an example of enum return value instead of bool?

Wolf480pl:
>Then some random changes will be saved, and some other changes will be not.

Now that I think of it... yeah, that'll need to be read-only too.

>That's not what I meant... You wanted an example of enum return value instead of bool?

I didn't want to introduce a new enum that'll be used in only one place. But maybe I should...

PR #1590, opened 2018-08-03
Unblock signals when spawning child processes.

TheWug:
The signal mask is inherited by children, so if we don't remove it
the shell module spawns processes which are accidentally resistant
to most signals.

see: man sigprocmask (znc uses pthread_sigmask, but they act the same)


TheWug:
Shouldn't be. Unlike the signal mask, pending signals are not inherited by children, so any such signal would need to have been delivered to the process after calling `fork()`.  It's pretty unlikely that such a signal would be generated during the short interval between fork and exec, and even if it did, my opinion is that the default signal handler is a perfectly reasonable way to respond to it.

PR #1591, opened 2018-08-03
Fix double-prefixing in log module when using explicit relative path with -d

TheWug:
CDir::CheckPathPrefix is called twice, both in OnLoad, and again in PutLog. Normally
the second is a no-op, but if the module path is explicitly relative (e.g. to "."),
It will incorrectly add the prefix twice.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -291,7 +291,6 @@ void CLogMod::PutLog(const CString& sLine,
                                          .Replace_n("\\", "-")).AsLower());
 
     // Check if it's allowed to write in this specific path
-    sPath = CDir::CheckPathPrefix(GetSavePath(), sPath);
     if (sPath.empty()) {
         DEBUG("Invalid log path [" << m_sLogPath << "].");
         return;


DarthGandalf:
What if I'm talking to `..` ?

TheWug:
ZNC handles `..` poorly with or without the patch.

Without the patch, sPath on :295 has the value:
`./users/WugTest/moddata/log/users/WugTest/moddata/log/2018-08-15.log`

With the patch, sPath has the values:
`./users/WugTest/moddata/log/TestNetwork/../2018-08-15.log`

These are both using the default logpath for user modules, `$NETWORK/$WINDOW/%Y-%m-%d.log`. 
The only difference between ordinary behavior and this special case is that CheckPathPrefix elides the .. path element by removing it and the prior one, it doesn't stop the undesirable filesystem traversal.

I can patch this too while I'm here. I propose "." and ".." be substituted with "log.dot" and "log.dotdot" if they appear by themselves in a path element.
- if they aren't by themselves, they'll be left alone
- $WINDOW's replacement contains a literal dot to prevent collision with nicknames, which never do

DarthGandalf:
If the problem is with relative path in -d, other parts of ZNC can behave
poorly with such path. In that case not log module should be fixed, but -d
flag: it should convert relative path to absolute path upon startup.

2018-08-15 14:55 GMT+01:00 Ryan <notifications@github.com>:

TheWug:
I disagree.  If something behaves badly when you give it a relative path, then I think ensuring all paths are absolute just masks possible bugs.

I think CDir::CheckPathPrefix is poorly named, poorly documented, and its implementation very inelegant. If a function like that were included now in a pull request, you would surely reject it. I think a much better solution would be to replace it with a new function whose behavior is more strongly defined.

DarthGandalf:
Well, both. -d should be fixed, and CDir::CheckPathPrefix should be fixed.


2018-08-15 15:37 GMT+01:00 Ryan <notifications@github.com>:

> I disagree. If something behaves badly when you give it a relative path,
> then I think ensuring all paths are absolute just masks possible bugs.
>

Are you willing to provide test cases for every possible ZNC codepath which
happens to use file paths? Ensuring all paths are absolute is more
self-contained change which fixes not only issue you discovered, but also
future issues with relative paths.

TheWug:
I have a build now that replaces CDir::CheckPathPrefix with 2 new functions, CFile::GetRelativeFile and CFile::Contains, that do the same thing, but generalized to relative paths.  CFile::GetRelativeFile in conjunction with another new function CFile::Canonize can also probably replace CDir::ChangeDir.  The only thing I'm not sure about is whether to kill them completely, or leave them around but deprecated, since the python and perl bindings use them, and removing them will break any python/perl scripts that depend on them.  I'll clean it up and push it to a new branch tonight.

DarthGandalf:
> I'll clean it up and push it to a new branch tonight.

Any progress?

PR #1608, opened 2018-10-13
Update submodule 'docker' to latest master.

Zarthus:
Uses znc 1.7.1 for local docker, avoiding the exploits 1.7.0 has


Zarthus:
Huh, that's interesting.

Is there a reason that file is completely separated? Is znc shipping 3 Dockerfiles?

Why not just let the `/Dockerfile` utilize `/docker/slim/.../Dockerfile`?

DarthGandalf:
If you have an idea how to reduce duplication, please say.
The Dockerfile(s) in znc-docker repo downloads the tarball to build the release version (https://hub.docker.com/_/znc/), while Dockerfile in znc repo builds from local source, resulting in e.g. https://hub.docker.com/r/zncbouncer/znc-git/tags/

PR #1615, opened 2018-11-08
Fix compilation without deprecated APIs in OpenSSL

neheb:
Added a few implicit headers that don't get included anymore and switched
to OpenSSL 1.0.0's THREAD API.

Note that this must be revisited for OpenSSL 1.1.0 and above as TLSv1_* functions are deprecated by TLS_*

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -368,6 +368,7 @@ if test "x$SSL" != "xno"; then
 		AC_LINK_IFELSE([
 			AC_LANG_PROGRAM([[
 				#include <openssl/ssl.h>
+				#include <openssl/dh.h>
 			]], [[
 				SSL_CTX* ctx = SSL_CTX_new(TLSv1_method());
 				SSL* ssl = SSL_new(ctx);
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -27,6 +27,8 @@
 #include <znc/Message.h>
 #ifdef HAVE_LIBSSL
 #include <openssl/ssl.h>
+#include <openssl/bn.h>
+#include <openssl/rsa.h>
 #endif /* HAVE_LIBSSL */
 #include <memory>
 #include <unistd.h>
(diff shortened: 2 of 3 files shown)


teward:
Correct me if I'm wrong, but isn't there a way to do this programmatically, which is, depending on which version of OpenSSL library is lying around, use different API calls?  (I'm not an OpenSSL expert, but it sounds like something like this should be doable to make it work for multiple OpenSSL library version targets...)

psychon:
ZNC already contains lots of stuff like `#if !defined(OPENSSL_VERSION_NUMBER) || OPENSSL_VERSION_NUMBER < 0x10100004`. This needs something similar, too.

Also, what OpenSSL version are you targeting @neheb? I thought that OpenSSL switched to "auto-magick" pthread support and does no longer need this thread id integration...? (And I also thought ZNC was already made to work with OpenSSL 1.0.0)

neheb:
This commit fixes 1.0.2 and parts of 1.1.0.

With deprecated APIs, these headers are included inside unrelated ones like engine.h Without, they are not.

@teward The thread stuff is already defined out if OpenSSL 1.1.0 or above is used as it's no longer necessary. It could be further defined to provide support for older than OpenSSL 1.0.0 releases, but those are long EOL.

DarthGandalf:
Build fixes should go to 1.7 branch. We can definitely remove support for ancient OpenSSL, but that needs to be done in master branch. So between these 2 commits the changes should be protected with ifdef.

neheb:
Ok. Added compatibility for OpenSSL 1.0.0. Should I make another pull request for the 1.7 branch?

dgw:
Or [change the base branch](https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/) of this one. The commit log shows that `1.7.x` is regularly merged back into the `master` branch.

DarthGandalf:
I cherry-picked it.

PR #1616, opened 2018-11-11
Make CUser::PutAllUser actually send to all networks

iczero:
Previously it would only send to the first network a user had instead of
them, this breaks partyline

Full description of original issue:
When a user has multiple networks, the partyline module would only work for the client connected to the first network of a user, as `CUser::PutAllUser` would only send to clients on that network. As a result, if a user had clients on two networks, network a and network b, only the client on network a would receive messages and other events for the channel. The client on network b would be able to send, but not receive anything.


iczero:
This patch does not work, as stated *kindly* by @moonheart08 and @handicraftsman. Closing for now and adding related configuration options for the partyline module instead.

handicraftsman:
> _kindly_
That italicized text.
lol
Ok shutting up - won't interject anymore.

PR #1619, opened 2018-11-24
Fix compilation without deprecated APIs in OpenSSL 1.1

neheb:
Prior commit was tested with 1.0.2. This one with 1.1.


DarthGandalf:
It's easier to just merge 1.7 to master after merging #1620 to 1.7. 1.7 is not that different yet from master, so the regular merge is simple.

neheb:
Ok

PR #1623, opened 2018-12-09
Fix compilation with LibreSSL < 2.7.0

neheb:
LibreSSL was stubborn and did not implement all of OpenSSL 1.1 despite
advertising support for it. Fixed in 2.7.0.


DarthGandalf:
They advertised the support from which version?
Cherry-picking to 1.7...

neheb:
Seems I was mistaken. They define it to 0x20000000L with all versions. What a joke.

PR #1625, opened 2018-12-23
ctcp: Distinguish Channel CTCP Requests and Replies

### LINKED ISSUE #1624: ZNC erroneously replies to CTCP NOTICE messages to channels.
It seems that a CTCP `NOTICE` message to a channel, such as a KVIrc user changing their avatar, is treated the same as if it were a `PRIVMSG`, so ZNC is sending a configured response if there is one and not passing it on to the client. This seems to be erroneous: the `NOTICE` message should not be replied to automatically like this.

https://github.com/znc/znc/blob/b52fecc42d5c1d354c55d0c90eecd49051d4ab34/src/IRCSock.cpp#L428-L434

(For normal private CTCP messages, the following lines deal with `NOTICE` replies:)

https://github.com/znc/znc/blob/b52fecc42d5c1d354c55d0c90eecd49051d4ab34/src/IRCSock.cpp#L420-L422

Zarthus:
Fixes #1624

Tested and confirmed non-regressing, though maybe @AndrioCelos's usecase is different from what I tried.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -429,7 +429,12 @@ bool CIRCSock::OnCTCPMessage(CCTCPMessage& Message) {
         if (pChan) {
             Message.SetChan(pChan);
             FixupChanNick(Message.GetNick(), pChan);
-            IRCSOCKMODULECALL(OnChanCTCPMessage(Message), &bResult);
+            if (Message.IsReply()) {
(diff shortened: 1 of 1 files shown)


DarthGandalf:
As this is regression fix, it should be in 1.7 branch

Zarthus:
Done.

PR #1634, opened 2019-01-21
Automatically discover DESC value in `make-tarball.sh`

DjLegolas:


make-tarball.sh
```diff
@@ -32,8 +32,7 @@ else
 	ZNCDIR=znc-$VERSION
 	TARGZ=$ZNCDIR.tar.gz
 	SIGN=1
-	DESC=""
-	# DESC="-rc1"
+	DESC="$(sed -En 's/set\(alpha_version (".*")\).*/\1/p' CMakeLists.txt)"
```
DarthGandalf: But that includes `"` in the version number

DjLegolas: Fixed, as far as i was able to test it.

PR #1639, opened 2019-02-08
WIP: refactor sockets

DarthGandalf:
Some things don't work yet, but this is more or less ready to review.

@psychon

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "docker"]
 	path = docker
 	url = https://github.com/znc/znc-docker
+[submodule "third_party/asio"]
+	path = third_party/asio
+	url = https://github.com/chriskohlhoff/asio
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,14 @@ if(WANT_ICU)
 endif()
 set(HAVE_ICU "${ICU_FOUND}")
 
+set(WANT_ASIO false CACHE BOOL "Use experimental asio backend")
+set(ZNC_USE_ASIO "${WANT_ASIO}")
+if(WANT_ASIO)
+	set(maybe_copy_csocket_h)
+else()
+	set(maybe_copy_csocket_h copy_csocket_h)
+endif()
+
 set(WANT_PERL false CACHE BOOL "Support Perl modules")
 set(WANT_PYTHON false CACHE BOOL "Support Python modules")
 set(WANT_PYTHON_VERSION "python3" CACHE STRING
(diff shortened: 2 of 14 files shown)


psychon:
I don't have much time to look at this in depth, but do I understand correctly that you took Csocket, ripped out some of the internals and replaced them with calls to https://github.com/chriskohlhoff/asio?
What's the goal here? Is this supposed to be faster, more secure, ...?

DarthGandalf:
> do I understand correctly that you took Csocket, ripped out some of the internals

That's correct. I copied some functions over as is, but replaced most of it with asio. I chose asio because it's supposed to be included in C++20, at which point we can hopefully drop that library from dependencies and simply bump the compiler requirements.

> What's the goal here? Is this supposed to be faster, more secure, ...?

It's less monolithic, so easier to modify. The main feature here are classes `base_layer` and `optional_layer`, which are combined in `Pipeline`. Currently there are only 2 layers (TCP and SSL), but more layers I can foresee:
* encoding (for now I copied it from Csocket verbatim, but it would be cleaner as a separate layer)
* socks proxy

I got an idea to do this long time ago, when was implementing encoding support in Csocket, but now the primary motivator to finally do it was Happy Eyeballs (#241), because #1571 has certain problems (races) which are not solvable outside of Csocket.

DarthGandalf:
@jimloco could you also take a look? Note that I didn't implement full Csocket API, but parts which map to asio relatively easy and which are used in ZNC.

DarthGandalf:
Status update: I don't like this callback hell, so trying to rewrite this using C++20 coroutines.

girst:
> C++20 coroutines.

as a user, I'd rather not have a dependency on such a new language standard. for example, centos7 ships with gcc 4.8--i wouldn't be able to compile znc any longer. other distributions aren't much more up to date either: ubuntu 18.04 ships 7.4 and even rhel8 and debian buster "only" ship g++ 8.3. ubuntu 20.04 will likely ship with gcc 9.2. 

according to [cppreference](https://en.cppreference.com/w/cpp/compiler_support), coroutines are supported only partially even by gcc 10; [gcc itelf](https://gcc.gnu.org/projects/cxx-status.html#cxx2a) currently says 'no support'.

DarthGandalf:
@girst of course.

I'm not making this a hard dependency yet. The socket implementation would be selectable via cmake, and if C++20 is not available, then you'd just get the old implementation, which means that bugs which this refactor is supposed to fix won't be fixed for your setup either. I think that's a fine compromise.

girst:
ah, good to know. that is a perfectly fine compromise!

DarthGandalf:
Update: this code I've written is terrible, and maintainable even less than existing Csocket.

Let's wait until a networking lands in C++ standard, be it asio or something else, then will look how to do such refactor in a better way.

PR #1651, opened 2019-04-12
modules: Remove partyline

### LINKED ISSUE #1632: Remove partyline
I think most users who enable partyline are misguidingly thinking it's 100% functional, and supporting them is a bit annoying. Clearly nobody has cared to fix it since 1.0 so it's on nobodies priority list. I'd say we should remove it entirely and refer to it as a third party module.

Zarthus:
Fixes #1632

There may be some additional considerations left, I spotted a
```
third_party/Csocket/examples/ChatServer.cc:                     WriteAll( "[" + m_sName + "] Has joined the partyline.\n" );
```

but the core idea gets across. I may look into additional cleanup (if necessary) around next Hacktoberfest.


src/IRCNetwork.cpp
```diff
@@ -425,6 +425,12 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
             CString sModName = sValue.Token(0);
             CString sNotice = "Loading network module [" + sModName + "]";
 
+            // We removed partyline on 2019-04-12, znc 1.8
+            if (sModName == "partyline") {
```
Zarthus: This prevents even a 3rd party module from being loaded if they re-add it, they can work around this by naming it "partyline2" or whatever.

I considered not adding this block but I think it makes more sense to warn people than to give a generic `not found`. WDYT?

src/IRCNetwork.cpp
```diff
@@ -424,6 +424,18 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
         for (const CString& sValue : vsList) {
             CString sModName = sValue.Token(0);
             CString sNotice = "Loading network module [" + sModName + "]";
+            CString sModRet;
+            CString sArgs = sValue.Token(1, true);
+
+            // We removed partyline on 2019-04-12, znc 1.8
```
DarthGandalf: I don't think this should be here. Removing partyline is a breaking change. If user had it loaded, probably they need it, and silently making it not work anymore in that case would be bad.

Zarthus: I've changed the code to explain why it failed instead, which can then be removed (if we want) in the next-next release of ZNC.  I think this resolves your concern.
```

[zarthus@nomenclature dev/znc/build.compile] master  ^?$ ./bin/znc -d /tmp/znc                                                                       [d94605c3]
[ .. ] Opening config [/tmp/znc/configs/znc.conf]...
[ .. ] Loading global module [webadmin]...
[ .. ] Binding to port [+1111]...
[ ** ] Loading user [za]
[ ** ] Loading network [freenode]
[ .. ] Loading network module [partyline]...
[ !! ] Unable to find module partyline
[ ** ] NOTICE: [partyline] is unavailable, cannot load.
[ ** ] NOTICE: [partyline] was removed in this release of ZNC. Please either remove it from your config or install it as a third party module with the same name.
[ ** ] NOTICE: More info can be found on https://wiki.znc.in/Partyline
[ ** ]
[ ** ] Unrecoverable config error.
```

src/IRCNetwork.cpp
```diff
@@ -450,9 +468,6 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
                 sModName = "modules_online";
             }
 
-            CString sModRet;
-            CString sArgs = sValue.Token(1, true);
-
             bool bModRet = LoadModule(sModName, sArgs, sNotice, sModRet);
 
             if (!bModRet) {
```
DarthGandalf: Why you didn't add another "if" below this line?

src/IRCNetwork.cpp
```diff
@@ -456,6 +456,21 @@ bool CIRCNetwork::ParseConfig(CConfig* pConfig, CString& sError,
...
+                // Partyline is removed in znc 1.8
+                if (sModName == "partyline") {
+                    CUtils::PrintError(
+                        "NOTICE: [partyline] is unavailable, cannot load.");
+                    CUtils::PrintError(
+                        "NOTICE: [partyline] is removed in this release"
+                        " of ZNC. Please either remove it from your config or "
+                        "install it"
```
DarthGandalf: Why "install it" is on its own line?

Zarthus: clang-format 🤷‍♂️

PR #1652, opened 2019-04-18
Add multiple hostmasks to autovoice

mjeveritt:
Porting the functionality of multiple hostmasks per user from the autoop module to autovoice.


psychon:
How did you test this? Did you test this at all?

mjeveritt:
> How did you test this? Did you test this at all?

It looks like my system isn't presently correctly set up for compilation, but I will investigate further...

mjeveritt:
> It looks like my system isn't presently correctly set up for compilation, but I will investigate further...

Someone forgot they needed to autogen.sh before configure/make .. all working fine now!

DarthGandalf:
> Someone forgot they needed to autogen.sh before configure

Or you can use cmake

DarthGandalf:
It would be more maintainable if autoop and autovoice share the code, instead of "porting" features from one to another...

mjeveritt:
> It would be more maintainable if autoop and autovoice share the code, instead of "porting" features from one to another...

Yes, I did think this .. what kind of module/program structure would you propose to do this best? Are there any more similar modules that would benefit from the same functions being made available "centrally"?

PR #1653, opened 2019-04-20
modules: Remove Q

Zarthus:
Probably merge-conflicts with #1651


DarthGandalf:
Add the reason for the removal to the commit message

dchmelik:
but I use Q

DarthGandalf:
@dchmelik ok, are you going to maintain the module?

Zarthus:
@dchmelik And you may continue to use it, you will just need to jump through one additional hoop with the pending changes.

This PR just marks the notion that ZNC will stop supporting it on an official level, because it already provides alternatives and QuakeNet are unwilling to provide generalized authentication methods like SASL.

PR #1654, opened 2019-04-21
ChanServ style replacement for two-column tables

### LINKED ISSUE #914: Revise tabular output
As discussed in #775, it would be nice if the massive wrapping tables were replaced with simple ChanServ-style output that looks good regardless of screen/window size, also with non-fixed-width fonts.

For example:

<pre>
    /msg *status help
    ...
    <b>RESTART</b> Restart ZNC
    <b>SHUTDOWN</b> Shutdown ZNC
    ...

    /msg *status help shutdown
    ...
    Syntax: SHUTDOWN [message]
    ...
</pre>

- introduce new commands to show more details instead of having lots of table columns (eg. listchans vs. showchan)
- reduce the amount of columns in all tables to two
- adjust CTable formatting or introduce a new type that doesn't even let add more than two "columns"
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9163398-revise-tabular-output?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

girst:
(fixes #914)

I'm sure there's stuff to improve on this PR (coding style, tests), but I'd like to get some feedback at this point from the @znc team please (see "requested feedback" below).

There has been an earlier attempt at cleaning up the tables (#922), but that was reverted later on due to its massive increase in vertical size. My patch is **opt-in**: meaning, only tables that are deemed safe are(quickly and easily) switched to the ChanServ style. I've ported the `/znc help` output as an example.

### Screenshots
**old format**: as you can see, the table is completely garbled on narrow screens (let alone mobile)
[znc-status-oldformat](https://user-images.githubusercontent.com/11820748/56472185-e131b700-644a-11e9-8f18-858750f3827e.png)

**new format**: this new format is much more compact than the old format in both axes. Bolding is used to differentiate between command and description (i.e. first and second column)
znc-status-newirssi
znc-status-newformat

### requested feedback

 - I'm unfamiliar with your test suite, so a pointer on which kind of tests you'd require would be appreciated. 
 - I'd like to apply this to more than just `/znc help`. Where else does it make sense?
 - In its current iteration this mode only works for two-column tables, but it can easily extended to concatenate columns 2..n with e.g. commas. Is that something you want?
 - if you don't like the optional parameter style, I've implemented a second version that adds a `CTable::SetStyle()` method to change the style in advance. This does not require changing anything other than adding that line (assuming it is a two-column table already) -- [see branch](/girst/znc/tree/set-table-style)
 - anything else is appreciated too! ;-)

### commit message
by passing an optional parameter (eType) to CTable::GetLine() one can
switch from the bulky and unreadable-on-narrow-devices-or-nonmonospaced-
fonts tables to a more compact "definition list" style of output. The
first "column" will be bolded for better visibility.

This is currently designed to replace two column tables only.

### DIFF
--- a/include/znc/Client.h
+++ b/include/znc/Client.h
@@ -266,7 +266,7 @@ class CClient : public CIRCSocket {
      *  \endcode
      */
     bool PutClient(const CMessage& Message);
-    unsigned int PutStatus(const CTable& table);
+    unsigned int PutStatus(const CTable& table, CTable::EStyle eStyle = CTable::GridStyle);
     void PutStatus(const CString& sLine);
     void PutStatusNotice(const CString& sLine);
     void PutModule(const CString& sModule, const CString& sLine);
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -128,7 +128,7 @@ class CException {
 };
 
 
-/** Generate a grid-like output from a given input.
+/** Generate a grid-like or list-like output from a given input.
  *
  *  @code
  *  CTable table;
@@ -152,9 +152,21 @@ class CException {
 +-------+-------+
 | hello | world |
 +-------+-------+@endverbatim
+ *
+ *  List-style output can be generated like so (only supports two columns!):
+ *  @code
+ *  while (table.GetLine(idx++, tmp, CTable::ListStyle)) {
+ *      // Output tmp somehow
+ *  }
+ *  @endcode
+ *  Output (asterisks mark bold text; Note that the header will be omitted):
+ *  @verbatim
+*hello*: world
+@endverbatim
  */
 class CTable : protected std::vector<std::vector<CString>> {
   public:
+    typedef enum { GridStyle, ListStyle } EStyle;
     CTable() {}
     virtual ~CTable() {}
 
@@ -185,9 +197,10 @@ class CTable : protected std::vector<std::vector<CString>> {
     /** Get a line of the table's output
      *  @param uIdx The index of the line you want.
      *  @param sLine This string will receive the output.
+     *  @param eStyle (optional) Display style.
      *  @return True unless uIdx is past the end of the table.
      */
-    bool GetLine(unsigned int uIdx, CString& sLine) const;
+    bool GetLine(unsigned int uIdx, CString& sLine, EStyle eStyle = GridStyle) const;
 
     /** Return the width of the given column.
      *  Please note that adding and filling new rows might change the
--- a/src/ClientCommand.cpp
+++ b/src/ClientCommand.cpp
@@ -1670,7 +1670,8 @@ void CClient::HelpUser(const CString& sFilter) {
         if (sFilter.empty() || sCmd.StartsWith(sFilter) ||
             sCmd.AsLower().WildCmp(sFilter.AsLower())) {
             Table.AddRow();
-            Table.SetCell(t_s("Command", "helpcmd"), sCmd + " " + sArgs);
+            Table.SetCell(t_s("Command", "helpcmd"),
+                          sCmd + (sArgs == "" ? "" : " ") + sArgs);
             Table.SetCell(t_s("Description", "helpcmd"), sDesc);
         }
     };
@@ -1893,6 +1894,6 @@ void CClient::HelpUser(const CString& sFilter) {
     if (Table.empty()) {
         PutStatus(t_f("No matches for '{1}'")(sFilter));
     } else {
-        PutStatus(Table);
+        PutStatus(Table, CTable::ListStyle);
     }
 }
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -805,11 +805,25 @@ bool CTable::SetCell(const CString& sColumn, const CString& sValue,
     return true;
 }
 
-bool CTable::GetLine(unsigned int uIdx, CString& sLine) const {
+bool CTable::GetLine(unsigned int uIdx, CString& sLine, EStyle eStyle) const {
     std::stringstream ssRet;
 
     if (empty()) {
         return false;
+    }
+
+    if (eStyle == ListStyle) {
+        if (m_vsHeaders.size() > 2) return false; // ListStyle mode can only do up to two columns
+        if (uIdx >= size()) return false;
+
+        const std::vector<CString>& mRow = (*this)[uIdx];
+        ssRet << "\x02" << mRow[0] << "\x0f"; //bold first column
+        if (m_vsHeaders.size() >= 2 && mRow[1] != "") {
+            ssRet << ": " << mRow[1];
+        }
+
+        sLine = ssRet.str();
+        return true;
     }
 
     if (uIdx == 1) {
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -593,10 +593,10 @@ void CClient::PutStatusNotice(const CString& sLine) {
     PutModNotice("status", sLine);
 }
 
-unsigned int CClient::PutStatus(const CTable& table) {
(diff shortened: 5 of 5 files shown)


src/Utils.cpp
```diff
@@ -805,13 +805,28 @@ bool CTable::SetCell(const CString& sColumn, const CString& sValue,
...
     }
 
+    if (as_deflist) {
+        if (m_vsHeaders.size() > 2) return false; // definition list mode can only do up to two columns
+        if (uIdx >= size()) return false;
+
+        const std::vector<CString>& mRow = (*this)[uIdx];
+        ssRet << "\x02" << mRow[0] << "\x0f"; //bold first column
+        if (m_vsHeaders.size() >= 2) {
+            ssRet << ": " << mRow[1];
```
DarthGandalf: And here you get a trailing space anyway if `mRow[1]` is empty.

src/Utils.cpp
```diff
@@ -805,13 +805,28 @@ bool CTable::SetCell(const CString& sColumn, const CString& sValue,
...
+        }
+
+        sLine = ssRet.str();
+        return true;
+
```
DarthGandalf: Remove this empty line

src/Utils.cpp
```diff
@@ -805,13 +805,28 @@ bool CTable::SetCell(const CString& sColumn, const CString& sValue,
...
 
     if (empty()) {
         return false;
     }
 
+    if (as_deflist) {
+        if (m_vsHeaders.size() > 2) return false; // definition list mode can only do up to two columns
+        if (uIdx >= size()) return false;
+
+        const std::vector<CString>& mRow = (*this)[uIdx];
```
DarthGandalf: nit: `VString` is an alias for that.

What does `m` stand for?

src/Utils.cpp
```diff
@@ -805,13 +805,28 @@ bool CTable::SetCell(const CString& sColumn, const CString& sValue,
     return true;
 }
 
-bool CTable::GetLine(unsigned int uIdx, CString& sLine) const {
```
DarthGandalf: Please add some test.

src/ClientCommand.cpp
```diff
@@ -1670,7 +1670,7 @@ void CClient::HelpUser(const CString& sFilter) {
         if (sFilter.empty() || sCmd.StartsWith(sFilter) ||
             sCmd.AsLower().WildCmp(sFilter.AsLower())) {
             Table.AddRow();
-            Table.SetCell(t_s("Command", "helpcmd"), sCmd + " " + sArgs);
+            Table.SetCell(t_s("Command", "helpcmd"), sCmd + (sArgs==""?"":" ") + sArgs);
```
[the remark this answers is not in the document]
girst: You can see an example on why this is necessary on the (outdated) hexchat screenshot at "Topics" (or any other command without arguments). A space is inserted between the command and the colon.

girst:
Thanks for the review! However, I am an idiot and pushed the wrong thing! I've replaced the boolean with an enum (so more than one alternative style is possible and it's more obvious than `true`).

include/znc/Client.h
```diff
@@ -266,7 +266,7 @@ class CClient : public CIRCSocket {
      *  \endcode
      */
     bool PutClient(const CMessage& Message);
-    unsigned int PutStatus(const CTable& table);
+    unsigned int PutStatus(const CTable& table, bool as_deflist = false);
```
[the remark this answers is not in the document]
girst: "definition list". sorry, should've explained that term. It is obsoleted by the enum now anyways.

include/znc/Utils.h
```diff
@@ -185,9 +185,10 @@ class CTable : protected std::vector<std::vector<CString>> {
     /** Get a line of the table's output
      *  @param uIdx The index of the line you want.
      *  @param sLine This string will receive the output.
+     *  @param as_deflist (optional) Opt into ChanServ style output (two-columns only)
      *  @return True unless uIdx is past the end of the table.
      */
-    bool GetLine(unsigned int uIdx, CString& sLine) const;
+    bool GetLine(unsigned int uIdx, CString& sLine, bool as_deflist = false) const;
```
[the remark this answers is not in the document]
girst: GetLine is always called though a wrapper that prints it out in one go, and since the change is required in there, that's where I put it. 

I think you might like my [second solution for this](/girst/znc/tree/set-table-style) better, then?

src/Utils.cpp
girst: I can add a non-empty check, no problem.

src/Utils.cpp
girst: all those variables are pre-existing ones. Those lines are adapted from the originals a few more lines down

src/Utils.cpp
girst: I'm not really sure what kind of test would be appropriate (there are no test for anything CTable related currently). Anything you have in mind?

include/znc/Utils.h
```diff
@@ -152,9 +152,21 @@ class CException {
 +-------+-------+
 | hello | world |
 +-------+-------+@endverbatim
+ *
+ *  ChanServ-style output can be generated like so (only supports two columns!):
```
Zarthus: I question the usefulness of this comment. ChanServ-style depends on services package and is subject to change in the future.

Zarthus: Same a couple of lines up.

girst: I was copying the wording from the original issue (#914). To be frank, I was at loss for a good description, so I switched between "definition list" and "ChanServ style".

Zarthus: Plain-style? Mobile-friendly?

Zarthus: You could do a show-dont-tell approach and just paste a small sample of the output it would generate.

girst: `GridStyle` v `ListStyle`, maybe? (Note that this is not a full replacement: you can only use it on two-column-tables. Things like `/znc ListNetworks` will for now stay at the old style)

girst: > You could do a show-dont-tell approach 

the GridStyle table also has such a sentence in front of it ("The above code would generate the following output:") and it helps to clarify the difference between the two output formats. 

I don't really see the point in arguing about such details right now if I'm honest. (If "ListStyle" is OK, I'll update all documentation to use the new term, of course)

Zarthus: Oh yeah, it's just a very minor point. It's part of the reason I deleted the comment :) - More of a "if we can quickly think of something better, go wild, if not - eh, alright"

girst:
just updated the nomenclature to GridStyle and ListStyle in the enum and documentation for both the [optional parameter](https://github.com/girst/znc/commit/bf0e83ae45ec7a6437ad0a5792925fdf0984a214) as well as the [SetStyle method](https://github.com/girst/znc/commit/9654031462b4249073e0f6ad1a325e9a4ca68d49) variants.

src/Utils.cpp
Zarthus: I'm not sure how valueable tests are considering the CI is almost always red and takes forever to run 🤷‍♂️

girst:
I'm closing this in favour of #1657, which is the `SetStyle()` variant of this same improvement. SetStyle() is a cleaner interface than this one. All improvements from the code review have been simultaneously ported to the new PR as well.

PR #1657, opened 2019-05-12
ListStyle (compact) replacement for two-column tables

### LINKED ISSUE #914: Revise tabular output

girst:
(fixes #914)
(obsoletes #1654)

**Note**: this obsoletes my earlier PR. Adding `SetStyle()` provides a cleaner API than patching any function that calls `GetLine()`.

Please see the `requested feedback` section below!

There has been an earlier attempt at cleaning up the tables (#922), but that was reverted later on due to its massive increase in vertical size. My patch is **opt-in**: meaning, only tables that are deemed safe are(quickly and easily) switched to the ChanServ style. I've ported the `/znc help` output as an example.

### Screenshots
**old format**: as you can see, the table is completely garbled on narrow screens (let alone mobile)
[znc-status-oldformat](https://user-images.githubusercontent.com/11820748/56472185-e131b700-644a-11e9-8f18-858750f3827e.png)

**new format**: this new format is much more compact than the old format in both axes. Bolding is used to differentiate between command and description (i.e. first and second column)
znc-status-newirssi
Screenshot_2019-05-19_14-05-29

### requested feedback

 - I'd like to apply this to more than just `/znc help`. Where else does it make sense?
 - In its current iteration this mode only works for two-column tables, but it can easily extended to concatenate columns 2..n with e.g. commas. Is that something you want?
 - anything else is appreciated too! ;-)

### commit message
by calling CTable::SetStyle(CTable::ListStyle) one can switch from the
bulky and unreadable-on-narrow-devices-or-nonmonospaced-fonts tables to
a more compact list style of output. The first "column" will be bolded
for better visibility and seperated with a colon from the second.

This is currently designed to replace two column tables only.

### DIFF
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -128,7 +128,7 @@ class CException {
 };
 
 
-/** Generate a grid-like output from a given input.
+/** Generate a grid-like or list-like output from a given input.
  *
  *  @code
  *  CTable table;
@@ -152,9 +152,25 @@ class CException {
 +-------+-------+
 | hello | world |
 +-------+-------+@endverbatim
+ *
+ *  If the table has at most two columns, one can switch to ListStyle output
+ *  like so:
+ *  @code
+ *  CTable table;
+ *  table.AddColumn("a");
+ *  table.AddColumn("b");
+ *  table.SetStyle(CTable::ListStyle);
+ *  // ...
+ *  @endcode
+ *
+ *  This will yield the following (Note that the header is omitted; asterisks
+ *  denote bold text):
+ *  @verbatim
+*hello*: world@endverbatim
  */
 class CTable : protected std::vector<std::vector<CString>> {
   public:
+    enum EStyle { GridStyle, ListStyle };
     CTable() {}
     virtual ~CTable() {}
 
@@ -162,9 +178,17 @@ class CTable : protected std::vector<std::vector<CString>> {
      *  Please note that you should add all columns before starting to fill
      *  the table!
      *  @param sName The name of the column.
-     *  @return false if a column by that name already existed.
+     *  @return false if a column by that name already existed or the current 
+     *  style does not allow this many columns.
      */
     bool AddColumn(const CString& sName);
+
+    /** Selects the output style of the table.
+     *  Select between different styles for printing. Default is GridStyle.
+     *  @param eNewStyle Table style type.
+     *  @return false if the style cannot be applied (usually too many columns).
+     */
+    bool SetStyle(EStyle eNewStyle);
 
     /** Adds a new row to the table.
      *  After calling this you can fill the row with content.
@@ -213,6 +237,7 @@ class CTable : protected std::vector<std::vector<CString>> {
     std::vector<CString> m_vsHeaders;
     // Used to cache the width of a column
     std::map<CString, CString::size_type> m_msuWidths;
+    EStyle eStyle = GridStyle;
 };
 
 #ifdef HAVE_LIBSSL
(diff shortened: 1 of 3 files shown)


include/znc/Utils.h
```diff
@@ -152,20 +152,44 @@ class CException {
...
+ *  @endcode
+ *
+ *  This will yield the following (Note that the header is omitted; asterisks
+ *  denote bold text):
+ *  @verbatim
+*hello*: world@endverbatim
  */
 class CTable : protected std::vector<std::vector<CString>> {
   public:
+    typedef enum { GridStyle, ListStyle } EStyle;
```
DarthGandalf: `enum EStyle { GridStyle, ListStyle };`

src/ClientCommand.cpp
```diff
@@ -1670,7 +1671,8 @@ void CClient::HelpUser(const CString& sFilter) {
         if (sFilter.empty() || sCmd.StartsWith(sFilter) ||
             sCmd.AsLower().WildCmp(sFilter.AsLower())) {
             Table.AddRow();
-            Table.SetCell(t_s("Command", "helpcmd"), sCmd + " " + sArgs);
+            Table.SetCell(t_s("Command", "helpcmd"),
+                          sCmd + (sArgs == "" ? "" : " ") + sArgs);
```
DarthGandalf: sArgs.empty()

include/znc/Utils.h
```diff
@@ -152,20 +152,44 @@ class CException {
...
+     *  style does not allow this many columns.
      */
     bool AddColumn(const CString& sName);
 
+    /** Selects the output style of the table.
+     *  Select between different styles for printing. Default is GridStyle.
+     *  @param eNewStyle Table style type.
+     *  @return false if the style cannot be applied (usually too many columns).
+     */
+    bool SetStyle(const EStyle eNewStyle);
```
DarthGandalf: https://abseil.io/tips/109

src/Utils.cpp
```diff
@@ -772,6 +774,18 @@ bool CTable::AddColumn(const CString& sName) {
     return true;
 }
 
+bool CTable::SetStyle(const EStyle eNewStyle) {
+    switch (eNewStyle) {
```
DarthGandalf: maybe "if"? switch here looks useless

girst: I think a switch is neater. It makes the intention of what SetStyle() does clearer: First, it checks if the requested style has any requirements (Gridstyle has none, ListStyle has 1). if all are fulfilled, it sets the style and returns true. 

If another style would get added to Estyle, it would be obvious that it should be added to the switch, but if there's only an `if (e==liststyle)`, it will look like liststyle is a special case.

girst:
> Do you know why on your screenshot on some rows there is space before colon?

That's because the HexChat screenshot was made before the `(sArgs == "" ? "" : " ")` and I didn't remember updating it. There are no stray spaces anymore.

girst:
Thanks for merging!

(perfectly timed during the week-long outage of my vps :| )

PR #1658, opened 2019-05-13
Fix: Update modtcl -> adding GetServerName function and network varia…

MalaGaM:
…ble with network name value.

Modification to have the variable ::network such as an eggdrop.

Example usage:
<MalaGaM> return $network
<*modtcl> artis
<MalaGaM> return $::network
<*modtcl> artis
<MalaGaM> return [GetNetworkName]
<*modtcl> artis

### DIFF
--- a/modules/modtcl/modtcl.tcl
+++ b/modules/modtcl/modtcl.tcl
@@ -34,7 +34,7 @@ set ::botnet-nick ZNC_[GetUsername]
 set ::botnick [GetCurNick]
 set ::server [GetServer]
 set ::server-online [expr [GetServerOnline] / 1000]
-set ::network	[GetServerName]
+set ::network	[GetNetworkName]
 
 # add some eggdrop style procs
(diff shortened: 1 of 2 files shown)


modules/modtcl/modtcl.tcl
```diff
@@ -34,6 +34,7 @@ set ::botnet-nick ZNC_[GetUsername]
 set ::botnick [GetCurNick]
 set ::server [GetServer]
 set ::server-online [expr [GetServerOnline] / 1000]
+set ::network	[GetServerName]
```
DarthGandalf: What exactly did you change compared to the previous PR? Looks exactly the same for me: GetServerName vs GetNetworkName.

I'd prefer GetNetworkName, because server name is probably a domain name, or IP address.

DarthGandalf:
You could force-push to the previous PR instead of opening a new one...

PR #1659, opened 2019-05-15
route_replies: Update to support freenode

glguy:


modules/route_replies.cpp
```diff
@@ -65,6 +65,7 @@ static const struct {
         {"313", false}, /* rfc1459 RPL_WHOISOPERATOR */
         {"317", false}, /* rfc1459 RPL_WHOISIDLE */
         {"319", false}, /* rfc1459 RPL_WHOISCHANNELS */
+        {"320", false}, /* RPL_WHOISSPECIAL */
```
DarthGandalf: Could you add a comment where this is defined? same for RPL_WHOISACTUALLY

glguy: `RPL_WHOISSPECIAL` introduced in https://github.com/unrealircd/unrealircd/blame/unreal42/include/numeric.h#L179

According to https://www.alien.net.au/irc/irc2numerics.html `RPL_WHOISACTUALLY` comes from `ircu`

I've updated the PR to include the sources of these numerics.

PR #1670, opened 2019-06-28
TrustPKI: Better wording in webadmin

Zarthus:
Per IRC, this wording is confusing and doesn't clearly denote the consequences, a less knowledgeable user may disable this not knowing what a PKI is

### DIFF
--- a/modules/data/webadmin/tmpl/add_edit_network.tmpl
+++ b/modules/data/webadmin/tmpl/add_edit_network.tmpl
@@ -80,9 +80,9 @@
 				</div>
 
 				<div class="subsection">
-					<div class="inputlabel"><? FORMAT "Trust the PKI:" ?></div>
+					<div class="inputlabel"><? FORMAT "Automatically detect trusted certificates (Trust the PKI):" ?></div>
 					<div class="checkbox"><input type="checkbox" name="trustpki" id="trustpki_checkbox"<? IF TrustPKI ?> checked="checked"<? ENDIF ?> />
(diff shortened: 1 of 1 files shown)


modules/po/webadmin.de_DE.po
```diff
@@ -200,13 +200,13 @@ msgstr ""
 "UNSICHER!"
 
 #: modules/po/../data/webadmin/tmpl/add_edit_network.tmpl:83
-msgid "Trust the PKI:"
+msgid "Automatically detect trusted certificates (Trust the PKI):"
 msgstr "Vertraue dem PKI:"
```
psychon: Shouldn't this be somehow marked as "not really translated", or something like that? If I understand this correctly, this is still considered a proper translation.

Zarthus: Yes, it depends on the i18n tool. Some of them automatically unverify translations on upload, some require you to clear the string, some of them require manual work

modules/po/webadmin.bg_BG.po
```diff
@@ -192,13 +192,13 @@ msgid ""
 msgstr ""
 
 #: modules/po/../data/webadmin/tmpl/add_edit_network.tmpl:83
-msgid "Trust the PKI:"
+msgid "Automatically detect trusted certificates (Trust the PKI):"
```
DarthGandalf: I don't think you should touch .po files at all in your PR. Automation will update po and pot automatically

Zarthus: Got it. Will revise.

Issue #1671, opened 2019-06-30
Compile from tarballs


zc0nf:
Hello

When I try to compile from tarball sources, it is not currently possible to do so.

Indeed, the dependency for the third_party is only present when I make a git clone, an issue had been opened #1521 , is it possible for each release and thus zip and tar.gz archive to add third_party missing to compile sources without problems?

The readme does not mention anywhere that it is not possible to compile from tarballs.

thank you,

zc0nf

psychon:
Uhm, the files in `third_party` are there for me in the release tarballs:
```
$ wget https://znc.in/releases/znc-1.7.4.tar.gz
$ tar -xvzf znc-1.7.4.tar.gz | grep third_party
znc-1.7.4/third_party/
znc-1.7.4/third_party/googletest/
znc-1.7.4/third_party/Csocket/
znc-1.7.4/third_party/Csocket/Csocket.cc
znc-1.7.4/third_party/Csocket/Csocket.h
```
Are you using https://github.com/znc/znc/archive/znc-1.7.4.tar.gz to download znc? If so, that file is automatically generated by GitHub and we cannot influence its content (well, it contains exactly the same files as are in git). That's exactly the same issue as #1521.

zc0nf:
Hi psychon,

Are you using https://github.com/znc/znc/archive/znc-1.7.4.tar.gz to download znc?
Yes

It is therefore necessary in this case, to update the readme evoking the impossibility to compile from the files made available in the Release section.

Unless someone can find a solution to fix or work around the problem?

For example, if I want to upgrade my znc to the latest stable version, which is the easiest and fastest method?

If I have to go through a git clone, I will have to position myself on the commit that speaks of the release in 1.7.4 which is forbidding and somehow imposes to be always on a development version?

thank you,

zc0nf

psychon:
> Are you using https://github.com/znc/znc/archive/znc-1.7.4.tar.gz to download znc?
> Yes

Then use https://znc.in/releases/znc-1.7.4.tar.gz

zc0nf:
Perfect,

Its possible to update readme ? :-)

zc0nf

PR #1672, opened 2019-06-30
Update error message for #1671

greenxx:
If it's not a git repo and modules are missing, you have to fetch them manually. 
Fix/enhance explanation for future users experiencing #1671

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,6 +251,8 @@ if(csocket_files STREQUAL "")
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
 			" Either this is not a git clone, or you don't have git installed")
+			" If you downloaded the source file from git submodules are not included \n"
+			" you'll need to fetch the modules in \"third_party\" manually ")
 	else()
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,9 +250,9 @@ if(csocket_files STREQUAL "")
 	if(git_status_var)
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
-			" Either this is not a git clone, or you don't have git installed")
-			" If you downloaded the source file from git submodules are not included \n"
-			" you'll need to fetch the modules in \"third_party\" manually ")
+			" Either this is not a git clone, or you don't have git installed."
+			" If you downloaded the source file from git, submodules are not included \n"
+			" you need to fetch the modules in \"third_party\" manually ")
 	else()
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,10 +250,8 @@ if(csocket_files STREQUAL "")
 	if(git_status_var)
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -250,7 +250,9 @@ if(csocket_files STREQUAL "")
 	if(git_status_var)
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
-			" Either this is not a git clone, or you don't have git installed")
+			" Either this is not a git clone, or you don't have git installed."
+			" If you downloaded the source file from git, submodules are not included \n"
```
DarthGandalf: the tarball from github, submodules...

CMakeLists.txt
```diff
@@ -250,7 +250,9 @@ if(csocket_files STREQUAL "")
...
+			" you need to fetch the modules in \"third_party\" manually ")
```
DarthGandalf: Also it probably should point at znc.in/releases/

CMakeLists.txt
greenxx: Is that better?

CMakeLists.txt
greenxx: Link added

CMakeLists.txt
```diff
@@ -250,7 +250,10 @@ if(csocket_files STREQUAL "")
 	if(git_status_var)
 		message(FATAL_ERROR
 			" It looks like git submodules are not initialized.\n"
-			" Either this is not a git clone, or you don't have git installed")
+			" Either this is not a git clone, or you don't have git installed. \n"
+			" If you downloaded the tarball or zip from git, third_party modules are not included. \n"
```
DarthGandalf: 1. submodules.
ZNC modules are something completely different.

2. github
How do you download it from "git"?

greenxx: 1: https://github.com/znc/znc/tree/master/third_party as in "third_party modules". Would `external dependencies` be better wording? Suggestions?

2: Front-page of project, `Clone or Download` and you can download as zip. 
Or from [releases page on GitHub](https://github.com/znc/znc/releases)

DarthGandalf: 1. I'm not sure what do you mean. "External dependencies" are more external than this, e.g. openssl is external, and we don't ship it together with ZNC, even in the actual release tarballs. In normal ZNC distribution third_party/ is supposed to be filled with 2 particular libraries, but in the downloaded .zip, it's empty. "git submodule" is the correct term here. If your concern is that "git submodule" is confusing, I don't know what to advise then.

2. Sure. That's specific to GitHub though, not Git. I'm sad people these days are confusing git and github...

DarthGandalf: 2. Moreover, if you use `git clone` instead of Github UI, the build works and/or tells you how exactly to get the submodules using `git` command.

greenxx: 1. That's why I asked what to call them. Sure they are git submodules that you include into your code (external repo). Most of your users probably don't know what `git submodules` are, but they are probably not the ones to compile from source either on the other hand.
Anyways, let's just encourage them to leech the tarball from the website.


2. Actuall it's not only GitHub, most of the versioning control systems with a WebUI have this feature like GitLab, BitBucket, Stash, Azure DevOps and more.. but that's not to focus on here. 

Let's just tell the users submodules are not initialized, if they know what to do wit that, good. 
If they don't know what to do, encourage them to leech the tarball from znc website.


Are you happy with that?

DarthGandalf: 2. Do they also silently fail to include submodules in the resulting zip file?

PR #1699, opened 2020-01-01
Welcome to 2020

MAGICCC:
By the way, the same procedure as last year, Miss Sophie?
The same procedure as every year, James!
Well, I'll do my very best!

[Skip CI]


DarthGandalf:
I think [skip ci] would need to be in commit, not in PR

PR #1700, opened 2020-01-02
add missing zncconfig include

SaveTheRbtz:


DarthGandalf:
Why it should matter? zncconfig is included automatically for every .cpp, via https://github.com/znc/znc/blob/0af3c4b234b7839bb1162e74b01b66e1a05e6fa1/src/CMakeLists.txt#L87

SaveTheRbtz:
Oh, that explains why I'm the only one seeing it.  Some context: this was found when I was trying to build `znc` with an alternative build system (`bazel`.)

DarthGandalf:
I'd be curious to see how you did it. What's the benefit there?

SaveTheRbtz:
Bazel benefits are mostly realized on large monorepos (e.g. google) because of the caching, distributed remote builds, and fast iterative builds.
As for the include itself main benefit is consistency since all other includes have it.

DarthGandalf:
> As for the include itself main benefit is consistency since all other includes have it.

Only some have it. There were various build failures caused by forgetting to add it, and that is bound to happen again, even if it's added to more files when discovering such failures. So now the build system adds it instead (and instances of the include can be removed from files as redundant, but that doesn't matter too much).

> Bazel benefits are mostly realized on large monorepos (e.g. google)

Yeah, I use it in my job, but ZNC is not a monorepo. What would be benefit for you to build ZNC with bazel?

> caching

ccache

> distributed remote builds

distcc

> fast iterative builds

cmake/make itself should do it just fine? I mean, running `make` doesn't rebuild files which were not changed

DarthGandalf:
For bazel support, you can pass the same `-include` flag to https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.copts and not depend on someone to keep remembering to add the `#include` everywhere

DarthGandalf:
Another way to get this in is if you have bazel working, add the whole bazel support (without breaking cmake), and add a test to CI which tests it. The test would prevent breaking it.

PR #1713, opened 2020-04-05
Web: fix small css errors after #1711

m4luc0:
Css was not completely right for all browsers after #1711.

1. A little less width for the checkbox labels works for all browsers
2. max-width instead of width (conflict with "Global Settings" -> first table)
3. More spacing between "Trust the PKI" and subsequent headline "Servers of this IRC network:" - https://imgur.com/a/8zlBYNI
4. Proper way to "clear: both" for subsection (:after pseudo element)

### DIFF
--- a/webskins/_default_/pub/_default_.css
+++ b/webskins/_default_/pub/_default_.css
@@ -293,8 +293,10 @@ tr.evenrow td {
 	font-size: 80%;
 }
 
-.subsection {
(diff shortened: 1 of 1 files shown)


DarthGandalf:
\<MetaNova\> the layout of 1713 results in https://i.imgur.com/H940TBQ.png

PR #1715, opened 2020-04-13
Fix incorrect html entities parsing in ZNCString.cpp

disconnect3d:
This PR fixes wrong size argument passed to `strncasecmp` function when it was invoked to check if the string contains HTML entities.

### DIFF
--- a/src/ZNCString.cpp
+++ b/src/ZNCString.cpp
@@ -294,13 +294,13 @@ CString CString::Escape_n(EEscape eFrom, EEscape eTo) const {
                     }
 
                     if (ch == 0) {
-                        if (!strncasecmp((const char*)&pTmp, "&lt;", 2))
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Can you add a test to StringTest.cpp?

disconnect3d:
> Can you add a test to StringTest.cpp?

Unfortunately, no. I found this issue in an automated way and I don't have time to add tests, sorry.

DarthGandalf:
How did you find it?

disconnect3d:
By grepping all GitHub repositories through Google BigQuery for certain function calls and then parsing them manually to find those where the string literal argument length is bigger than the `size` argument (number literal). I am going to write a blog post about it too :P.

DarthGandalf:
Sounds interesting, thanks

PR #1723, opened 2020-05-17
Web: remove legacy xhtml syntax

gasp:
according to w3c,

* `<?xml` syntax is deprecated in html: XML processing instructions are not supported in HTML
* The `align` attribute on the `td` element is obsolete and layout be done in css
* [polyglot html](https://www.w3.org/TR/html-polyglot/) is being deprecated


webskins/_default_/tmpl/BaseHeader.tmpl
```diff
@@ -6,7 +6,7 @@
 <? AddRow JSLoop HREF=/pub/selectize-standalone-0.12.1.min.js ?>
 <? AddRow CSSLoop HREF=/pub/jquery-ui-sortable.1.11.4.min.css ?>
 <? AddRow CSSLoop HREF=/pub/selectize-0.12.1.css ?>
-<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+<html lang="en" dir="ltr">
```
DarthGandalf, line 9: Why did you make it ltr? Do you think for Arabic/Hebrew support something needs to be changed in the markup?

PR #1725, opened 2020-05-31
Add route_monitor module.

### LINKED ISSUE #1036: route_replies doesn't route monitor
* Moved from https://github.com/communi/communi-sailfish/issues/106
* http://ircv3.net/specs/core/monitor-3.2.html

I usually have two clients in ZNC, WeeChat and Communi Sailfish. WeeChat always puts some users to monitor, but Communi Sailfish gets them too.

The numerics I see are 731 (offline) and 730 (online).

* * * * *

* ZNC 1.7.x-git-250-645fa85 - http://znc.in
* Debian GNU/Linux 7.8 (wheezy)

ajaspers:
This module implements the MONITOR protocol.
https://ircv3.net/specs/core/monitor-3.2

route_monitor keeps track of which clients have subscribed to individual targets, and routes 730/731 replies accordingly. MONITOR L is served from internal state (so 732/733 replies don't need to be routed) and the module keeps track of the maximum number of subscriptions so 734 doesn't need to be routed.

Resolves #1036. Per discussion in #znc-dev, I implemented this as a separate module rather than put it into route_replies.

Questions

- [ ] After looking at other ZNC modules, it seems that we can assume that only a single thread will be calling into the network module at any given time. Is that correct, or do I need to add a mutex when internal state is read/written?

- [ ] Is it okay to always use :irc.znc.in as the server name, or should this be extracted somehow? Other modules do this in certain circumstances, so it seemed okay.

Remaining tasks
- [ ] Sync MONITOR state with server when the module is initially loaded. Since we don't know which clients requested a specific target, I was planning on subscribing all connected clients to all targets upon initial load. /quote MONITOR C can be issued from a client that doesn't understand MONITOR.

- [ ] When the IRC connection is lost and reestablished, MONITOR + should be replayed from internal state.

- [ ] Respect TARGMAX for monitor. The overall limit is respected, and commands/replies will be kept <= 512 octets but the TARGMAX could be small even when the overall limit is big.

- [ ] Create wiki page

Hoping to get some initial feedback before I proceed further. Thanks!

### DIFF
--- a/modules/route_monitor.cpp
+++ b/modules/route_monitor.cpp
@@ -1,0 +1,460 @@
+/*
+ * Copyright (c) 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <znc/IRCNetwork.h>
+#include <znc/IRCSock.h>
+#include <znc/Modules.h>
+
+namespace {
+
+// SplitAndSend constructs commands and replies <= 512 octets, when the last
+// parameter is a comma-separated list.
+//
+// Precondition: The length of base plus the length of any individual member of
+// the list must be <= 512.
+template <typename Iterator, typename ConvertToCStringFn>
+void SplitAndSend(const CString &base, Iterator i_start, const Iterator &i_end,
+                  ConvertToCStringFn asString,
+                  const std::function<void(const CString &)> &send) {
+    while (i_start != i_end) {
+        CString command = base + asString(*i_start);
+        ++i_start;
+        while (i_start != i_end &&
+               (command.length() + 1 + asString(*i_start).length()) <= 512) {
+            command += "," + asString(*i_start);
+            ++i_start;
+        }
+        send(command);
+    }
+}
+
+// CasefoldedNick uses lowercased nick for comparisons.
+// NOTE: Only supports ASCII casefolding since that is all CString::AsLower
+// supports.
+class CasefoldedNick {
+  public:
+    explicit CasefoldedNick(const CString &nick) {
+        m_casefoldedNick = nick.AsLower();
+    }
+
+    const CString &GetCasefoldedNick() const { return m_casefoldedNick; }
+
+    bool operator<(const CasefoldedNick &rhs) const {
+        return m_casefoldedNick < rhs.m_casefoldedNick;
+    }
+
+    bool operator==(const CasefoldedNick &rhs) const {
+        return m_casefoldedNick == rhs.m_casefoldedNick;
+    }
+
+  private:
+    CString m_casefoldedNick;
+};
+
+// MonitorTarget stores state about a nick.
+struct MonitorTarget {
+    CString actualNick; // non-casefolded nick
+    bool online;
+};
+
+// AsActualNick converts casefolded nicks into actual nicks, using the
+// targetState map.
+class AsActualNick {
+  public:
+    // targetState map must outlive this class.
+    explicit AsActualNick(
+        const std::map<CasefoldedNick, MonitorTarget> *targetState)
+        : m_targetState(targetState) {}
+
+    const CString operator()(const CasefoldedNick &nick) const {
+        auto it = m_targetState->find(nick);
+        if (it == m_targetState->end()) {
+            // This shouldn't happen, but we just fallback to the casefolded
+            // nick in this case.
+            return nick.GetCasefoldedNick();
+        }
+        return it->second.actualNick;
+    }
+
+  private:
+    const std::map<CasefoldedNick, MonitorTarget> *m_targetState; // not owned
+};
+
+// Identity returns the string unchanged, when the nick is already not
+// casefolded.
+const CString &Identity(const CString &nick) { return nick; }
+
+const int kMonitorSupportUnknown = -2;
+const int kMonitorNotSupported = -1;
+
+} // namespace
+
+class CRouteMonitorMod : public CModule {
+  public:
+    MODCONSTRUCTOR(CRouteMonitorMod) {}
+
+    bool OnLoad(const CString &, CString &) override {
+        // TODO: Sync state with server. If the server supports MONITOR and
+        // there are existing subscriptions, subscribe all clients to them.
+        return true;
+    }
+
+    void OnIRCConnected() override {
+        // When OnIRCConnected() is called, ISUPPORT may not be available yet.
+        m_monitorLimit = kMonitorSupportUnknown;
+
+        // TODO: Replay MONITOR + to server instead of clearing state.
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Another remaining task:
* [ ] add a test

modules/route_monitor.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include <znc/Modules.h>
+
+namespace {
+
+// SplitAndSend constructs commands and replies <= 512 octets, when the last
+// parameter is a comma-separated list.
+//
+// Precondition: The length of base plus the length of any individual member of
+// the list must be <= 512.
+template <typename Iterator, typename ConvertToCStringFn>
```
DarthGandalf, line 28: Is this template needed?

modules/route_monitor.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+    void OnIRCDisconnected() override {
+        // TODO: Preserve state so we can replay MONITOR + on reconnect.
+        ClearState();
+    }
+
+    void OnClientDisconnect() override { ClearClientMonitorSubscriptions(); }
+
+    EModRet OnRawMessage(CMessage &message) override {
+        if (!GetNetwork()->GetIRCSock() ||
```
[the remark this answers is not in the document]
ajaspers: I'm not sure so I added this check. If the server disconnects and there are incoming messages still queued, would they get sent to the module?

modules/route_monitor.cpp
ajaspers: Iterator is needed since SplitAndSend is called with a variety of container iterators (std::set<CasefoldedNick>, std::vector<CString>,  std::set<CString>).

ConvertToCStringFn also needs to be templated since it can be called with a CasefoldedNick or a CString.

modules/route_monitor.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    void OnIRCDisconnected() override {
+        // TODO: Preserve state so we can replay MONITOR + on reconnect.
+        ClearState();
+    }
+
+    void OnClientDisconnect() override { ClearClientMonitorSubscriptions(); }
+
+    EModRet OnRawMessage(CMessage &message) override {
```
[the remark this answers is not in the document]
ajaspers: Done.

ajaspers:
This should be fully ready to review now. I completed the remaining tasks aside from the wiki page, and added a test which exercises most of the module's functionality.

ajaspers:
> Why do you need a concept of case-folding? It makes code much more complicated

IRC servers do casefolding internally; if a monitored user "asdf" changes their nick to "aSdF", the server considers this to be the same nick and the subscription for that user remains. If route_monitor didn't do the same, it would lose track of these subscriptions.

I've tested this with InspIRCd and Oragono. Here's an example:

User 1 connects with nick "asdf"
```
:testnet.inspircd.org NOTICE asdf :*** You are connected to testnet.inspircd.org using TLS (SSL) cipher 'TLS1.2-ECDHE-RSA-AES-256-GCM-AEAD'
```

User 2 connects with nick "test", subscribes to "asdf"
```
:testnet.inspircd.org NOTICE test :*** You are connected to testnet.inspircd.org using TLS (SSL) cipher 'TLS1.2-ECDHE-RSA-AES-256-GCM-AEAD'
MONITOR + asdf
:testnet.inspircd.org 730 test :asdf
```

User 1 changes their nick to "aSdF" and quits.
```
NICK aSdF
:asdf!a@<redacted> NICK :aSdF
QUIT
```

User 2 receives the 731 (offline) reply for aSdF.
```
:testnet.inspircd.org 731 test :aSdF
```

test/integration/tests/modules.cpp
```diff
@@ -188,6 +188,62 @@ TEST_F(ZNCTest, KeepNickModule) {
...
+        ":server 005 user MONITOR=100 TARGMAX=MONITOR:100 :are supported by "
+        "this server");
+    client1.ReadUntil(":server 005");
+
+    client1.Write("znc loadmod route_monitor");
+    client1.ReadUntil("Loaded module");
+
+    // route_monitor should sync existing subscriptions.
+    ircd.ReadUntil("MONITOR L");
+    ircd.Write(":server 732 user :existing");
```
DarthGandalf, line 208: Does the user monitor himself?

test/integration/tests/modules.cpp
```diff
@@ -188,6 +188,62 @@ TEST_F(ZNCTest, KeepNickModule) {
...
+    ircd.Write(":server 733 user :End of list");
+    ircd.ReadUntil("MONITOR S");
+    ircd.Write(":server 731 user :existing");
```
DarthGandalf, line 211: ...but that same user is offline

test/integration/tests/modules.cpp
```diff
@@ -188,6 +188,62 @@ TEST_F(ZNCTest, KeepNickModule) {
...
+    // MONITOR L should only return the targets subscribed to by the client.
+    client1.Write("MONITOR L");
+    client1.ReadUntil(":irc.znc.in 732 user :a,c,existing");
+    client1.ReadUntil(":irc.znc.in 733 user :End of MONITOR list");
+
+    // Issuing MONITOR C from client2 should only unsubscribe from b, since
+    // client1 is still subscribed to a and c.
+    client2.Write("MONITOR C");
+    ircd.ReadUntil("MONITOR - b");
+}
```
DarthGandalf, line 245: Add test for reconnecting, for casefolding

DarthGandalf:
> > Why do you need a concept of case-folding? It makes code much more complicated
> 
> IRC servers do casefolding internally; if a monitored user "asdf" changes their nick to "aSdF", the server considers this to be the same nick and the subscription for that user remains. If route_monitor didn't do the same, it would lose track of these subscriptions.
> 
> I've tested this with InspIRCd and Oragono. Here's an example:
> 
> User 1 connects with nick "asdf"
> 
> ```
> :testnet.inspircd.org NOTICE asdf :*** You are connected to testnet.inspircd.org using TLS (SSL) cipher 'TLS1.2-ECDHE-RSA-AES-256-GCM-AEAD'
> ```
> 
> User 2 connects with nick "test", subscribes to "asdf"
> 
> ```
> :testnet.inspircd.org NOTICE test :*** You are connected to testnet.inspircd.org using TLS (SSL) cipher 'TLS1.2-ECDHE-RSA-AES-256-GCM-AEAD'
> MONITOR + asdf
> :testnet.inspircd.org 730 test :asdf
> ```
> 
> User 1 changes their nick to "aSdF" and quits.
> 
> ```
> NICK aSdF
> :asdf!a@<redacted> NICK :aSdF
> QUIT
> ```
> 
> User 2 receives the 731 (offline) reply for aSdF.
> 
> ```
> :testnet.inspircd.org 731 test :aSdF
> ```

This should be mentioned in a comment for the CasefoldedNick class

modules/route_monitor.cpp
```diff
@@ -0,0 +1,594 @@
...
+        }
+
+        unsigned int numeric = message.GetCode();
+
+        // Refresh ISUPPORT information if this is a monitor numeric or we're
+        // waiting to replay subscriptions.
+        // Don't refresh on numeric 5 (ISUPPORT) since this hasn't been
+        // processed by ZNC core yet, and we get the information from ZNC core.
+        if ((numeric >= 730 && numeric <= 733) ||
+            (numeric != 5 && !m_replayNeeded.empty())) {
```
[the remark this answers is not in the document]
kylef: 👍 on this, also note that a server can disable monitor by sending something along the lines of `:server 005 guest -MONITOR :are supported by this server` at any time which I do not think it being handled correctly.

PR #1734, opened 2020-06-28
Extend port warning to 6697

ryzokuken:
Reworded the warning just a little bit to accommodate both ports. Let me know if you want the specific messages back, and I would happily do that with some string templating magic.


DarthGandalf:
Looks fine. Could you rebase this for 1.8.x branch?

PR #1736, opened 2020-07-05
Allow ISUPPORT parameters to be negated

kylef:
ISUPPORT allows negating parameters via `-PARAMETER` which allows servers to change functionality without disconnecting clients. As per https://modern.ircdocs.horse/#rplisupport-005:

> Tokens of the form -PARAMETER are used to negate a previously specified parameter. If the client receives a token like this, the client MUST consider that parameter to be removed and revert to the behaviour that would occur if the parameter was not specified. The client MUST act as though the paramater is no longer advertised to it. These tokens are intended to allow servers to change their features without disconnecting clients. Tokens of this form MUST NOT contain a value field.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -1356,7 +1356,11 @@ void CIRCSock::ParseISupport(const CMessage& Message) {
             break;
         }
 
-        m_mISupport[sName] = sValue;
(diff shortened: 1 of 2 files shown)


kylef:
Please see https://github.com/znc/znc/pull/1737

PR #1737, opened 2020-07-05
Allow ISUPPORT parameters to be negated

kylef:
ISUPPORT allows negating parameters via `-PARAMETER` which allows servers to change functionality without disconnecting clients. As per https://modern.ircdocs.horse/#rplisupport-005:

> Tokens of the form -PARAMETER are used to negate a previously specified parameter. If the client receives a token like this, the client MUST consider that parameter to be removed and revert to the behaviour that would occur if the parameter was not specified. The client MUST act as though the paramater is no longer advertised to it. These tokens are intended to allow servers to change their features without disconnecting clients. Tokens of this form MUST NOT contain a value field.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -1356,7 +1356,11 @@ void CIRCSock::ParseISupport(const CMessage& Message) {
             break;
         }
 
-        m_mISupport[sName] = sValue;
+        if (sName.StartsWith("-")) {
+            m_mISupport.erase(sName.TrimPrefix_n("-"));
+        } else {
+            m_mISupport[sName] = sValue;
+        }
 
         if (sName.Equals("PREFIX")) {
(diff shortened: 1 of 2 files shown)


src/IRCSock.cpp
```diff
@@ -1391,10 +1395,16 @@ void CIRCSock::ParseISupport(const CMessage& Message) {
             if (m_bNamesx) continue;
             m_bNamesx = true;
             PutIRC("PROTOCTL NAMESX");
+        } else if (sName.Equals("-NAMESX")) {
+            if (!m_bNamesx) continue;
```
DarthGandalf, line 1399: `continue` here is useless. In the true branch it prevented from sending `PROTOCTL` again, but here you don't send anything like `PROTOCTL -NAMESX`

DarthGandalf:
Docker on Travis choked on the name of your branch...

PR #1739, opened 2020-07-11
Bugfix adding listener and Small Improvements on Strings

brunodOut:
The Port number of a listener, either in the Config file or coming from the AddPort command (among other possible means) always comes as a string. To listen to a port, on the other hand, an unsigned short int is necessary, so the code makes a conversion using CString's method ToUShort.

The method converts the string into an unsigned long int, and then converts it into an unsigned short int returning to the caller the number. Due to this conversion, the exceeding left bits of the number are lost, transforming, for instance (assuming a 16bit unsigned short int), 65536 in 0, 65537 in 1, and so forth.

This causes unexpected behaviors when a port > 65535 is passed. When added to the znc.conf file, the program tries to load the config file and opens whatever port it gets in the conversion process. The impact of AddPort command is minimized, since when the Listener cannot be added it does not get saved. The first scenario, however, leads to strange and hard to find issues, like listening to a seemingly random port which ZNC is not configured for. In both cases, the error messages are also unclear about the issue.

The proposed pull request contains 2 commits, detailing how the issue was fixed, and also potential impacts and limitations and the principles governing the changes.


src/znc.cpp
```diff
@@ -1738,12 +1748,21 @@ bool CZNC::AddListener(CConfig* pConfig, CString& sError) {
...
-    if (!pConfig->FindUShortEntry("port", uPort)) {
+    unsigned long int ulPort; //read the port with some extra bits for validation
+    if (!pConfig->FindStringEntry("port", sPort)) {
         sError = "No port given";
         CUtils::PrintError(sError);
         return false;
     }
+    ulPort = sPort.ToULong();
+    if (ulPort > 65535 || ulPort < 1) {
+	sError = t_f("Port [{1}] is out of the 1-65535 range")(sPort);
```
DarthGandalf, line 1761: could you run clang-format?

src/ZNCString.cpp
```diff
@@ -1239,7 +1239,18 @@ short CString::ToShort() const {
     return (short int)strtol(this->c_str(), (char**)nullptr, 10);
 }
 unsigned short CString::ToUShort() const {
-    return (unsigned short int)strtoul(this->c_str(), (char**)nullptr, 10);
+    unsigned short int ret;
+    unsigned long int val;
+    val = strtoul(this->c_str(), (char**)nullptr, 10);
+
+    // We need short ints greater than 16bits as much as men need nipples
```
DarthGandalf, line 1246: be more respectful of the source code

brunodOut: Ohh, sorry, it was an internal note to remember where it was, I forgot to remove it. I'll surely remove it!

src/znc.cpp
brunodOut: @DarthGandalf sure, I'll run, my fault, I'm note used to writing code anymore... Any project specific usage? Or just standard?

DarthGandalf: It's already configured via https://github.com/znc/znc/blob/master/.clang-format so either just run it from command line, or if your IDE has some way to invoke it for the changed lines only, that would be best

(unfortunately clang-format has/had bugs in how it handles tabs :( )

PR #1744, opened 2020-07-31
Allow ordering of channels

### LINKED ISSUE #104: Allow ordering channels [$15]

njhanley:
Fixes #104

### DIFF
--- a/modules/data/webadmin/tmpl/add_edit_network.tmpl
+++ b/modules/data/webadmin/tmpl/add_edit_network.tmpl
@@ -184,6 +184,7 @@
 					<tr>
 						<td>[<a href="<? VAR URIPrefix TOP ?><? VAR ModPath TOP ?>addchan?user=<? VAR Username ESC=URL ?>&amp;network=<? VAR Name ESC=URL ?>"><? FORMAT "Add" ?></a>]</td>
 				<? IF ChannelLoop ?>
+						<th><? FORMAT "Index" ?></th>
 						<th><? FORMAT "Save" ?></th>
 						<th><? FORMAT "Name" ?></th>
 						<th><? FORMAT "CurModes" ?></th>
@@ -196,13 +197,14 @@
 					</tr>
 					</thead>
 
-					<tbody>
+					<tbody id="channels">
 			<? LOOP ChannelLoop ?>
 					<tr class="<? IF __EVEN__ ?>evenrow<? ELSE ?>oddrow<? ENDIF ?>">
 						<td>
 							<input type="hidden" name="channel" value="<? VAR Name ?>" />
 							[<a href="<? VAR URIPrefix TOP ?><? VAR ModPath TOP ?>editchan?user=<? VAR Username ESC=URL ?>&amp;network=<? VAR Network ESC=URL ?>&amp;name=<? VAR Name ESC=URL ?>"><? FORMAT "Edit" ?></a>] [<a href="<? VAR URIPrefix TOP ?><? VAR ModPath TOP ?>delchan?user=<? VAR Username ESC=URL ?>&amp;network=<? VAR Network ESC=URL ?>&amp;name=<? VAR Name ESC=URL ?>"><? FORMAT "Del" ?></a>]
 						</td>
+						<td><input class="channel_index" type="number" name="index_<? VAR Name ?>" min="1" max="<? VAR MaxIndex ?>" value="<? VAR Index ?>"/></td>
 						<td><input type="checkbox" name="save_<? VAR Name ?>"<? IF InConfig ?> checked="checked"<? ENDIF ?> /></td>
 						<td><? VAR Name ?></td>
 						<td><? VAR CurModes ?></td>
@@ -213,6 +215,7 @@
 			<? ENDLOOP ?>
 					</tbody>
 				</table>
+				<script>channellist_init(jQuery);</script>
 			</div>
 		</div>
 		<? ENDIF ?>
--- a/modules/data/webadmin/tmpl/add_edit_network.tmpl
+++ b/modules/data/webadmin/tmpl/add_edit_network.tmpl
@@ -197,7 +197,7 @@
 					</thead>
 
 					<tbody>
-			<? LOOP ChannelLoop SORTASC=Name ?>
+			<? LOOP ChannelLoop ?>
 					<tr class="<? IF __EVEN__ ?>evenrow<? ELSE ?>oddrow<? ENDIF ?>">
 						<td>
 							<input type="hidden" name="channel" value="<? VAR Name ?>" />
--- a/src/ClientCommand.cpp
+++ b/src/ClientCommand.cpp
@@ -453,6 +453,46 @@ void CClient::UserCommand(CString& sLine) {
             PutStatus(t_p("Disabled {1} channel", "Disabled {1} channels",
                           uDisabled)(uDisabled));
         }
+    } else if (sCommand.Equals("MOVECHAN")) {
+        if (!m_pNetwork) {
+            PutStatus(t_s(
+                "You must be connected with a network to use this command"));
+            return;
+        }
+
+        const auto sChan = sLine.Token(1);
+        const auto sTarget = sLine.Token(2);
+        if (sChan.empty() || sTarget.empty()) {
+            PutStatus(t_s("Usage: MoveChan <#chan> <index>"));
+            return;
+        }
+
+        unsigned int uIndex = sTarget.ToUInt();
+
+        CString sError;
+        if (m_pNetwork->MoveChan(sChan, uIndex - 1, sError))
+            PutStatus(t_f("Moved channel {1} to index {2}")(sChan, uIndex));
+        else
+            PutStatus(sError);
+    } else if (sCommand.Equals("SWAPCHANS")) {
+        if (!m_pNetwork) {
+            PutStatus(t_s(
+                "You must be connected with a network to use this command"));
+            return;
+        }
+
+        const auto sChan1 = sLine.Token(1);
+        const auto sChan2 = sLine.Token(2);
+        if (sChan1.empty() || sChan2.empty()) {
+            PutStatus(t_s("Usage: SwapChans <#chan1> <#chan2>"));
+            return;
+        }
+
+        CString sError;
+        if (m_pNetwork->SwapChans(sChan1, sChan2, sError))
+            PutStatus(t_f("Swapped channels {1} and {2}")(sChan1, sChan2));
+        else
+            PutStatus(sError);
     } else if (sCommand.Equals("LISTCHANS")) {
         if (!m_pNetwork) {
             PutStatus(t_s(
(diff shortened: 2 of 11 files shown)


psychon:
Could you briefly outline what this change does? From a quick look, this seems only related to the config, but it does not actually do anything on the IRC wire protocol.

So, how exactly do you understand "order of channels" and how is this order communicated to IRC clients? What happens if channel number two cannot be joined (e.g. because one is banned)? Would this mean that channel number three cannot be attempted to be joined since that would mess up the order that the client sees?

Basically, I always understood #104 as something that is just not supported by the IRC protocol and so I am curious to understand how you solved these issues, but too lazy to read the patch carefully to understand this.

njhanley:
The first commit makes CConfig preserve the order of subconfigs as they appear in znc.conf. The second adds commands to manipulate the order of channels in a network's channel vector.

> From a quick look, this seems only related to the config, but it does not actually do anything on the IRC wire protocol.

No changes to the wire protocol are needed. ZNC already joins clients to channels in a defined order, specifically lexicographically by name. These changes put that order under the user's control.

> So, how exactly do you understand "order of channels" and how is this order communicated to IRC clients?

The order of channels in the context of #104 is the order in which ZNC will attach a client to each channel. There is nothing new to communicate to clients.

> What happens if channel number two cannot be joined (e.g. because one is banned)?

The same behavior as current; the channel will be skipped. It will still occupy a place in the network's channel vector and can be ordered with the two new commands.

> Basically, I always understood #104 as something that is just not supported by the IRC protocol and so I am curious to understand how you solved these issues, but too lazy to read the patch carefully to understand this.

I'm not sure what you mean by unsupported by the IRC protocol.

psychon:
So, if znc fails to join a channel for whatever reason and the user then later joins that channel "by hand", the resulting order in the irc client will be "wrong". Okay.

Still, doing what you did might be enough for some people. Thanks.

test/ConfigTest.cpp
```diff
@@ -146,6 +145,10 @@ TEST_F(CConfigSuccessTest, SubConf8) {
     TEST_SUCCESS(" \t <A B>\nfoo = bar\n\tFooO = bar\n</a>",
                  "->a/B\nfoo=bar\nfooo=bar\n<-\n");
 }
+TEST_F(CConfigSuccessTest, SubConf9) {
+    TEST_SUCCESS("<foo b>\n</foo>\n<foo a>\n</foo>",
```
DarthGandalf, line 149: Please add comment that it's supposed to test not sorting the subconfs by name

src/Config.cpp
```diff
@@ -122,14 +123,17 @@ bool CConfig::Parse(CFile& file, CString& sErrorMsg) {
                 else
                     pActiveConfig = &ConfigStack.top().Config;
 
-                SubConfig& conf = pActiveConfig->m_SubConfigs[sTag.AsLower()];
-                SubConfig::const_iterator it = conf.find(sName);
+                const auto sTagLower = sTag.AsLower();
+                auto& nameset = pActiveConfig->m_SubConfigNameSets[sTagLower];
 
-                if (it != conf.end())
+                const auto& it = nameset.find(sName);
```
DarthGandalf, line 129: auto

Reference is not appropriate here, even though technically it works due to lifetime extension

modules/data/webadmin/tmpl/add_edit_network.tmpl
```diff
@@ -197,7 +197,7 @@
 					</thead>
 
 					<tbody>
-			<? LOOP ChannelLoop SORTASC=Name ?>
```
DarthGandalf, line 200: This belongs to separate commit

src/ClientCommand.cpp
```diff
@@ -438,6 +438,50 @@ void CClient::UserCommand(CString& sLine) {
                       "] channels matching [" + sPatterns + "]");
             PutStatus("Disabled [" + CString(uDisabled) + "] channels");
         }
+    } else if (sCommand.Equals("MOVECHAN")) {
+        if (!m_pNetwork) {
+            PutStatus("You must be connected with a network to use this command");
```
DarthGandalf, line 443: Missing `t_s()`, and below

I wonder how I missed it some other strings in this file

src/ClientCommand.cpp
```diff
@@ -438,6 +438,50 @@ void CClient::UserCommand(CString& sLine) {
...
+        const auto sTarget = sLine.Token(2);
+        if (sChan.empty() || sTarget.empty()) {
+            PutStatus("Usage: MoveChan <#chan> <index>");
+            return;
+        }
+
+        // use one-based indexing
+        unsigned int uIndex = sTarget.ToUInt();
+        if (uIndex == 0 || uIndex > m_pNetwork->GetChans().size()) {
+            PutStatus(t_f("Invalid index {1}")(uIndex));
```
DarthGandalf, line 457: It's not obvious what number user should pass to the command. Probably ListChans should show the current index?

src/IRCNetwork.cpp
```diff
@@ -942,6 +942,38 @@ bool CIRCNetwork::DelChan(const CString& sName) {
...
+bool CIRCNetwork::MoveChan(const CString& sName, unsigned int uIndex) {
+    if (uIndex >= m_vChans.size()) return false;
+
+    unsigned int i;
+    for (i = 0; i < m_vChans.size(); i++)
+        if (m_vChans[i]->GetName().Equals(sName)) break;
+    if (i == m_vChans.size()) return false;
+
+    const auto pChan = m_vChans[i];
+    for (int delta = i < uIndex ? 1 : -1; i != uIndex; i += delta)
```
DarthGandalf, line 954: erase() followed by insert() would be much simpler

src/IRCNetwork.cpp
```diff
@@ -942,6 +942,38 @@ bool CIRCNetwork::DelChan(const CString& sName) {
...
+    for (i = 0; i < m_vChans.size(); i++)
+        if (m_vChans[i]->GetName().Equals(sChan1)) break;
+    if (i == m_vChans.size()) return false;
+
+    unsigned int j;
+    for (j = 0; j < m_vChans.size(); j++)
+        if (m_vChans[j]->GetName().Equals(sChan2)) break;
+    if (j == m_vChans.size()) return false;
+
+    const auto pChan = m_vChans[i];
```
DarthGandalf, line 971: std::swap?

njhanley:
I can't find any tests for the existing commands. Where might I add the new ones?

Yes, reordering channels in the webadmin would be useful. I'll see what I can do.

DarthGandalf:
> I can't find any tests for the existing commands. Where might I add the new ones?

It's in https://github.com/znc/znc/tree/master/test/integration/tests

src/IRCNetwork.cpp
```diff
@@ -942,6 +942,35 @@ bool CIRCNetwork::DelChan(const CString& sName) {
     return false;
 }
 
+bool CIRCNetwork::MoveChan(const CString& sName, unsigned int uIndex) {
```
DarthGandalf, line 945: Every time this function is called, you're checking first whether the channel exists.

Maybe pass the `CChannel*` instead, to avoid looking it up again? Same for swap

Another option is to return a better error message from here instead of just "false", then you won't need to check it before calling, but just need to show the error message returned from here

src/IRCNetwork.cpp
```diff
@@ -942,6 +942,35 @@ bool CIRCNetwork::DelChan(const CString& sName) {
...
+    if (uIndex >= m_vChans.size()) return false;
+
+    auto it = m_vChans.begin();
+    for (; it != m_vChans.end(); ++it)
```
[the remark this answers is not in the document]
njhanley: Initially I thought this would require `std::find_if` with a rather ugly lambda. I see now that passing a `CChan*` from `CIRCNetwork::FindChan` would work.

modules/data/webadmin/tmpl/add_edit_network.tmpl
```diff
@@ -213,6 +213,15 @@
 			<? ENDLOOP ?>
 					</tbody>
 				</table>
+				<script>
+				$("#channels").sortable({
```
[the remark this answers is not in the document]
njhanley: That's correct. I will implement a non-JS interface.

src/IRCNetwork.cpp
njhanley: Passing the `CChan*` would be cleaner. This and removing the redundant check on `uIndex` would make both `MoveChan` and `SwapChan` void functions and leave all input validation at the call sites.

modules/data/webadmin/tmpl/add_edit_network.tmpl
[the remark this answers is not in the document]
njhanley: I'd like to rework the channels interface by copying the design of the network servers list.

Instead of the current table, the non-JS UI would be a textarea where each line specifies the configuration of a channel. With JS, the textarea would be replaced by a table with more user friendly inputs and draggable rows. I would also remove the add/edit channel page and move its options to the new interface.

Alternatively I could add a textarea (hidden with JS) to specify the channels in the desired order. This approach would not change the existing UI.

I believe the first approach is cleaner. What are your thoughts?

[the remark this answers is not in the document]
DarthGandalf: > I'd like to rework the channels interface by copying the design of the network servers list.

Tbh, I'm not entirely happy with what I did there. Currently the JS editor edits the (hidden) text area, while it would probably be cleaner to send a flag to the server whether to take input from the text area or from separate fields resulting from JS editor directly. Not sure.

> the non-JS UI would be a textarea where each line specifies the configuration of a channel

The channel config is more complicated than list of servers, so this would make webadmin essentially unusable for editing channels without JS. Maybe you have a mock of the new UI if I misunderstood what do you mean?

> I would also remove the add/edit channel page and move its options to the new interface.

This will break modules which add parts of UI to that page, e.g. stickychan. How would you present that in a single text field?

src/IRCNetwork.cpp
DarthGandalf: I've just realized that `CChan*` wouldn't be enough, because it will still require searching for it in the vector.

There are other functions here already which return the error via `CString& sError`

src/IRCNetwork.cpp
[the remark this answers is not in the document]
njhanley: I don't see how this would be an improvement. Also `std::find_if` would be needed.

modules/data/webadmin/tmpl/add_edit_network.tmpl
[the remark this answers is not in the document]
njhanley: > Maybe you have a mock of the new UI if I misunderstood what do you mean?

Something like this.

channel serialization mock

Here's a possible grammar in EBNF.
```ebnf
channel = name , space ,
          [ "key:" , key , space ] ,
          buffer size , space ,
          auto clear buffer , space ,
          save , space ,
          enabled , space ,
          detached , space ,
          default modes , newline ;
name = ? irc channel name ? ;
key = ? irc channel key ? ;
buffer size = "default" | integer ;
auto clear buffer = "default" | boolean ;
save = boolean ;
enabled = boolean ;
detached = boolean ;
default modes = ? irc channel modes, may contain spaces ? ;

space = " " , { " " } ;
newline = "\n" ;
boolean = "false" | "true" ;
integer = "0" | "1"..."9" , { "0"..."9" } ;
```

> This will break modules which add parts of UI to that page, e.g. stickychan.

I overlooked that section of the page, my bad. I'll abandon this idea.

njhanley:
I've added `sError` parameters to MoveChan and SwapChans and removed the redundant checks/searches.
Also I wrote a different interface for rearranging channels in the webadmin. It works without JS.

channel indexes mock

DarthGandalf:
Thanks!
I've added similar CSS to other skins to not make the index field so wide

PR #1746, opened 2020-08-03
Add support for cap account-tag

NuclearW:
The account-tag capability is now requested when ZNC connects to an IRC server, the same is then offered to clients connecting.
This permits the tag to then pass through to account-tag aware clients.


kylef:
That's a great start, there's a couple of changes that would be needed to push this forward:

- Only advertising to clients when the upstream server supports account-tag. We cannot guarentee support on all servers which 
- Using CAP NEW / CAP DEL to update client state
- The account tag should be filtered out for clients which did not enable account-tag. One client enabling the tag can break other clients which have no idea about message tags.

I'd like to see some tests to count for these cases.

kylef:
Nice, that's a great start! I've given this some manual testing and it appears to work as expected. I think I've found one bug regarding the behaviour of removing the capability (CAP DEL) still leaves the capability enabled according to ZNC which leaves the client thinking its disabled while the capability is enabled (partially).

```
< setup: connected to server with account-tag, client negotiated account-tag (used CAP LS 302 thus cap-notify is implicitly enabled) >
< znc disconnects after we had account-tag enabled >
S: :irc.znc.in CAP Guest59998 DEL :account-notify account-tag away-notify extended-join

C: CAP LIST
S: :irc.znc.in CAP Guest59998 LIST :account-tag
```

Perhaps around [CClient::ClearServerDependentCaps](https://github.com/znc/znc/blob/c79961c514a172e07821b2dafe81483569118d98/src/Client.cpp#L845) would need to be updated to unset the internal state of CClient.

It would be great to see some tests for this introduced behaviour, you can find similar tests in https://github.com/znc/znc/blob/c79961c514a172e07821b2dafe81483569118d98/test/ClientTest.cpp#L158. It would be good to test functionality such as when account-tag is enabled with server (`m_pTestSock`) and reciving account tag isn't propergated to client which didn't enable account-tag. Along with when it account-tag is enabled on both sides and the tag is presented to client.

NuclearW:
Thanks for taking a look.  I've tried to recreate the bug you've pointed out and I'm having trouble doing so.

Below is the debug log showing `CAP LS` and `CAP LS 302` (how weechat sends `/cap ls` by default) before and after disconnecting znc from an account-tag enabled server.  I had set the flag for account-tag to be server dependent where it is defined as part of `m_mCoreCaps` in Client.h so it shows up in the `CAP DEL` sent to the client.  I see that in your example output as well.  Just that it appears to be actually removed from the list right after in my testing.

```
[2020-08-03 08:44:45.171543] (nw/default) CLI -> ZNC [CAP LS 302]
[2020-08-03 08:44:45.171766] (nw/default) ZNC -> CLI [@time=2020-08-03T12:44:45.171Z :irc.znc.in CAP nw` LS :account-notify account-tag away-notify batch cap-notify echo-message extended-join multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso]
[2020-08-03 08:45:08.089625] (nw/default) CLI -> ZNC [CAP LS]
[2020-08-03 08:45:08.089764] (nw/default) ZNC -> CLI [@time=2020-08-03T12:45:08.089Z :irc.znc.in CAP nw` LS :account-notify account-tag away-notify batch cap-notify echo-message extended-join multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso]
[2020-08-03 08:45:36.309492] (nw/default) CLI -> ZNC [PRIVMSG *status :disconnect]
[2020-08-03 08:45:36.309653] (nw/default) ZNC -> IRC [QUIT :ZNC - http://znc.in]
[2020-08-03 08:45:36.309708] (nw/default) ZNC -> CLI [:*status!znc@znc.in PRIVMSG nw` :Disconnected from IRC. Use 'connect' to reconnect.]
[2020-08-03 08:45:36.309949] IRC::nw::default == Disconnected()
[2020-08-03 08:45:36.310121] (nw/default) ZNC -> CLI [@time=2020-08-03T12:45:36.310Z :irc.znc.in CAP nw` DEL :account-notify account-tag away-notify extended-join]
[2020-08-03 08:45:47.176368] (nw/default) CLI -> ZNC [CAP LS]
[2020-08-03 08:45:47.176507] (nw/default) ZNC -> CLI [@time=2020-08-03T12:45:47.176Z :irc.znc.in CAP nw` LS :batch cap-notify echo-message multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso]
[2020-08-03 08:47:14.898958] (nw/default) CLI -> ZNC [CAP LS 302]
[2020-08-03 08:47:14.899090] (nw/default) ZNC -> CLI [@time=2020-08-03T12:47:14.899Z :irc.znc.in CAP nw` LS :batch cap-notify echo-message multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso]
```

I will also look into seeing if a test can be added.  Off hand I don't see any documentation on running the tests, so I can test my tests, is there a page outlining this?

kylef:
The problem is with [`CAP LIST`](https://ircv3.net/specs/core/capability-negotiation#the-cap-list-subcommand) which returns the list of enabled capabilities. `CAP LS` returns the available capabilities. The enabled capabilities include `account-notify` after reciving DEL. After digging into this in more detail, it seems this would apply to other existing capabilities too, `CClient::ClearServerDependentCaps` does not appear to cleanup `m_ssAcceptedCaps` thus these deleted capabilities are still marked as accepted. Thus `CAP LIST` (which returns `m_ssAcceptedCaps`) includes `account-notify`. So not directly a problem introduced in this branch, just discovered it while reviewing these changes. Could be fixed outside this PR given it isn't introduced here.

---

As for the tests, there is a `unittest` target, if you're using CMake with the makefile generator you can run `make unittest` to invoke the mentioned tests.

NuclearW:
A ha, my bad, I misread `CAP LIST` as `CAP LS`

Following your logic, I added an erase in `CClient::ClearServerDependentCaps` to remove the capabilities from `m_ssAcceptedCaps`, since that seems to be the fix needed.  If you would prefer the fix for the issue uncovered but not introduced here come in another PR or in another fashion let me know.

I will also now look into the `unittest` target and see if I can't also make a test for the `account-tag` capability.

NuclearW:
I have added an `AccountTag` test at this time.

This test enables the capability on the server (`m_pTestSock`) then presents a message with the `@account=account-name` tag present to a client (`m_pTestClient`) which has not enabled the `account-tag` capability.  The capability is then turned on and the same message presented.  The test checks that the tag will only pass through to the client when it is enabled.

include/znc/IRCSock.h
```diff
@@ -150,6 +150,7 @@ class CIRCSock : public CIRCSocket {
     bool HasUHNames() const { return m_bUHNames; }
     bool HasAwayNotify() const { return m_bAwayNotify; }
     bool HasAccountNotify() const { return m_bAccountNotify; }
+    bool HasAccountTag() const { return m_bAccountTag; }
```
DarthGandalf, line 153: Unused? I don't think it worth making all caps to have such accessors, if there is a way to query cap support by name.

kylef:
I've realised theres some cases where account-tag support isn't quite as simple in a bouncer. This implementation breaches the account-tag specification, there are some MUST clauses which are not fulfilled (and I am not sure we can fulfil them without adding a lot of complexity).

> The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG, MODE, NOTICE, and all others).

If account-tag is enabled by the client, then all PRIVMSG/MODE/NOTICE and other messages must also contain an account tag. We might have messages stored in the our buffers while account-tag wasn't enabled with the upstream server which we may replay to a client which has enabled account-tag, the replay would breach that clause.

ZNC may have offered the end user `echo-message`, where ZNC itself will be echoing the users message *without* their account tag.

> The tag also MUST be added by the ircd to all numerics directly caused by the sender

There's many cases where ZNC would directly reply to messages by producing the numeric response (which won't have an account tag), for example handling a `MODE` command can produce 324 and 329 numerics (https://github.com/znc/znc/blob/f5c3ea4/src/Client.cpp#L1104).

---

I've got some thoughts on being able to solve some of these MUST breaches, although I'm not to keen on all the complexity that these would bring:

- znc might need to always enable account-tag with the server so that message buffers contain account-tag when future clients might enable the capability.
- Should ZNC only offer account-tag if all messages in ZNC buffers have been received with account-tag? Would ZNC send `CAP NEW` when buffers are cleared or new messages push old messages out of buffers?
- can/should we use account-notify to be able to know about account-tag's for messages when account-tag is not enabled with upstream server.
- Would we need to always know about our own account so that ZNC produced messages and numerics are able to contain the account-tag? If so, would account-notify facilitate that?

DarthGandalf:
> I've realised theres some cases where account-tag support isn't quite as simple in a bouncer.

Shall we roll this back for now?

> znc might need to always enable account-tag with the server

Definitely. Why would it not do so? Doesn't it do it now?

> Should ZNC only offer account-tag if all messages in ZNC buffers have been received with account-tag? Would ZNC send CAP NEW when buffers are cleared or new messages push old messages out of buffers?

I'd prefer to avoid such dependency.

> There's many cases where ZNC would directly reply to messages by producing the numeric response

This might be ~easy to fix by adding the tag in something like OnSendToClientMessage(), though need to be careful with multiple clients connected.

> Would we need to always know about our own account so that ZNC produced messages and numerics are able to contain the account-tag? If so, would account-notify facilitate that?

That could work with CAP NEW - start advertising account-tag to clients after ZNC learnt its account.

PR #1750, opened 2020-08-21
Unix Listener

cheako:
#1363 + shared ssl and merge master.

Check CClient::UserPortCommand in src/ClientCommand.cpp, I don't know about the "listports" entries.


cheako:
Also, I think the integration tests were removed.

DarthGandalf:
Why? don't remove them

I'll let you fix the build before starting reviewing this.

PR #1760, opened 2020-10-17
Make renaming a network case-sensitive

### LINKED ISSUE #1454: Network renaming is case-insensitive
Discovered by changing the capitalisation for a network name within the web interface and resulted in `Network [network] already exists` message.

RotationMatrix:
This PR modifies the logic for detecting duplicate network names to resolve #1454 .

I'm also participating in hacktoberfest. Please let me know if there's anything you'd like me to improve! 😄

### DIFF
--- a/modules/webadmin.cpp
+++ b/modules/webadmin.cpp
@@ -1077,7 +1077,10 @@ class CWebAdminMod : public CModule {
             WebSock.PrintErrorPage(t_s("Network name is a required argument"));
             return true;
         }
-        if (!pNetwork || !pNetwork->GetName().Equals(sName)) {
+        if (pNetwork && pNetwork->GetName().Equals(sName)) {
+            // Set the network name to pickup changes to capitalization.
+            pNetwork->SetName(sName);
+        } else {
             CString sNetworkAddError;
             CIRCNetwork* pOldNetwork = pNetwork;
             pNetwork = pUser->AddNetwork(sName, sNetworkAddError);
@@ -1094,8 +1097,6 @@ class CWebAdminMod : public CModule {
                 pNetwork->Clone(*pOldNetwork, false);
                 pUser->DeleteNetwork(pOldNetwork->GetName());
             }
-        } else {
-            pNetwork->SetName(sName);
         }
 
         CString sArg;
--- a/modules/webadmin.cpp
+++ b/modules/webadmin.cpp
@@ -1077,7 +1077,7 @@ class CWebAdminMod : public CModule {
             WebSock.PrintErrorPage(t_s("Network name is a required argument"));
             return true;
         }
-        if (!pNetwork || pNetwork->GetName() != sName) {
+        if (!pNetwork || !pNetwork->GetName().Equals(sName)) {
             CString sNetworkAddError;
             CIRCNetwork* pOldNetwork = pNetwork;
             pNetwork = pUser->AddNetwork(sName, sNetworkAddError);
@@ -1094,6 +1094,8 @@ class CWebAdminMod : public CModule {
                 pNetwork->Clone(*pOldNetwork, false);
                 pUser->DeleteNetwork(pOldNetwork->GetName());
             }
+        } else {
+            pNetwork->SetName(sName);
         }
 
         CString sArg;
--- a/src/ClientCommand.cpp
+++ b/src/ClientCommand.cpp
@@ -726,7 +726,8 @@ void CClient::UserCommand(CString& sLine) {
             return;
         }
 
-        if (pNewUser->FindNetwork(sNewNetwork)) {
+        if (pNewUser->FindNetwork(sNewNetwork) && (!sOldUser.Equals(sNewUser) ||
+            !sOldNetwork.Equals(sNewNetwork))) {
             PutStatus(t_f("User {1} already has network {2}.")(sNewUser,
                                                                sNewNetwork));
             return;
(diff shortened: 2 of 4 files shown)


src/User.cpp
```diff
@@ -526,7 +526,7 @@ CIRCNetwork* CUser::AddNetwork(const CString& sNetwork, CString& sErrorRet) {
 }
 
 bool CUser::AddNetwork(CIRCNetwork* pNetwork) {
-    if (FindNetwork(pNetwork->GetName())) {
+    if (FindNetwork(pNetwork->GetName(), CaseSensitivity::CaseSensitive)) {
```
DarthGandalf, line 529: If you do it like this, there will be two networks with almost the same name, the difference only in case of letters, no?

DarthGandalf: And then a future `FindNetwork()` will return arbitrarily one or the other

RotationMatrix: Hmm, that is a good point. I initially did it this way to try to restrict the change size, but that is a nasty side effect. 🤔

teward:
I'm curious, why exactly do we *need* to have two networks of the same/near-same name?  Making things case-sensitive makes sense in *rare* cases, but I don't think so for IRC network names.  Consider an email address - `foo@bar.baz` is identical to `FOO@BAR.BAZ` and having capitalization in that way would be VERY confusing, both at the domain level and the left-hand-side recipient name... which is why case insensitivity in this manner is more preferred.

This gets a -1 from me on those bases.

DarthGandalf:
@teward have you read the linked issue? #1454

teward:
Only peripherally.  However, I think this needs the more verbose check - `if Renaming, then: if NewName == AnyOtherNetworkName.Lowercase, then Fail)` - for the same logic of my last message.  Message I made was supposed to be a comment on the commit you poked at @DarthGandalf ("If you do it like this, there will be two networks with almost the same name, the difference only in case of letters, no?")

Now I"m caught up, so ignore my last message :)

RotationMatrix:
Definitely agree @teward this must be more verbose. And yeah upper vs lower is basically the same. Was concerned myself about how specific one needs to be when typing, and what happens when one forgets about caps lock etc.

Going to see if I can locate where the edit path originates. Somehow it doesn't go through `AddNetwork()` if the network name isn't changed, so that's probably where the issue really lies.

RotationMatrix:
Closing this cause there's nothing to salvage. Going to open a new one with the improved approach.

DarthGandalf:
Okay. Though you could do `git push -f` to the same branch.

RotationMatrix:
Your right about force pushing. Thinking it will be cleaner/clearer to separate it. But if you prefer I continue in this one, I'll reopen this.

RotationMatrix:
Actually, looking back it's only one commit. Just going to reopen.

modules/webadmin.cpp
```diff
@@ -1094,6 +1094,8 @@ class CWebAdminMod : public CModule {
                 pNetwork->Clone(*pOldNetwork, false);
                 pUser->DeleteNetwork(pOldNetwork->GetName());
             }
+        } else {
+            pNetwork->SetName(sName);
```
DarthGandalf, line 1098: This will crash.

DarthGandalf: Ah, no, it uses || there

DarthGandalf: Reversing the condition to `pNetwork && Equals()` would make it less confusing

RotationMatrix: I think I'm following. So like this?

```cpp
if (pNetwork && pNetwork->GetName().Equals(sName)) {
    pNetwork->SetName(sName);
} else {
    // AddNetwork()
}
```

DarthGandalf: Yep, with a comment what SetName is supposed to do there

src/ClientCommand.cpp
```diff
@@ -726,7 +726,8 @@ void CClient::UserCommand(CString& sLine) {
             return;
         }
 
-        if (pNewUser->FindNetwork(sNewNetwork)) {
+        if (pNewUser->FindNetwork(sNewNetwork) &&
+            (!sOldUser.Equals(sNewUser) || !sOldNetwork.Equals(sNewNetwork))) {
```
DarthGandalf, line 730: Why are you comparing username?

src/ClientCommand.cpp
```diff
@@ -758,29 +759,38 @@ void CClient::UserCommand(CString& sLine) {
...
-        }
+            if (!pNewNetwork) {
+                PutStatus(t_f("Error adding network: {1}")(sNetworkAddError));
+                return;
+            }
 
-        pNewNetwork->Clone(*pOldNetwork, false);
+            pNewNetwork->Clone(*pOldNetwork, false);
+        } else {
+            CIRCNetwork* pOldNetwork = pNewUser->FindNetwork(sOldNetwork);
```
DarthGandalf, line 775: Same comment as in webadmin: this would be more understandable with the reverse condition and a comment

src/ClientCommand.cpp
```diff
@@ -758,29 +759,38 @@ void CClient::UserCommand(CString& sLine) {
...
+        if (!sOldUser.Equals(sNewUser) ||
+            !pOldNetwork->GetName().Equals(sNewNetwork)) {
+            if (!pOldUser->DeleteNetwork(sOldNetwork)) {
+                PutStatus(t_s(
+                    "Copied the network to new user, but failed to delete old "
                     "network"));
+            }
         }
+
+        PutStatus(t_s("Success."));
```
DarthGandalf, line 793: > Copied the network to new user, but failed to delete old network
> Success.

src/ClientCommand.cpp
RotationMatrix: It appears two users can have networks with equal names, and in the "movenetwork A znc B znc" case, the error should be given even though the network names are equal.

src/ClientCommand.cpp
RotationMatrix: 🤦

src/ClientCommand.cpp
DarthGandalf: The username is already converted to the pointer to CUser. In a case-sensitive way. Compare pointers instead of reimplementing the existing logic (but with subtly different semantics)

src/ClientCommand.cpp
```diff
@@ -758,29 +759,40 @@ void CClient::UserCommand(CString& sLine) {
             pMod->MoveRegistry(sNewModPath);
         }
 
-        CString sNetworkAddError;
-        CIRCNetwork* pNewNetwork =
-            pNewUser->AddNetwork(sNewNetwork, sNetworkAddError);
+        if (sOldUser.Equals(sNewUser) &&
+            pOldNetwork->GetName().Equals(sNewNetwork)) {
+            // Set the network name to pick up capitalization changes.
+            CIRCNetwork* pOldNetwork = pNewUser->FindNetwork(sOldNetwork);
```
DarthGandalf, line 765: This variable already defined above

modules/webadmin.cpp
```diff
@@ -1077,7 +1077,10 @@ class CWebAdminMod : public CModule {
             WebSock.PrintErrorPage(t_s("Network name is a required argument"));
             return true;
         }
-        if (!pNetwork || pNetwork->GetName() != sName) {
+        if (pNetwork && pNetwork->GetName().Equals(sName)) {
+            // Set the network name to pickup changes to capitalization.
+            pNetwork->SetName(sName);
```
DarthGandalf, line 1082: This skips the rest of logic of renaming network, like moving files to the new place or calling `OnAddNetwork`/`OnDeleteNetwork` of modules

RotationMatrix: 1. Copy [this](https://github.com/znc/znc/blob/6cd80535c064e91155468e79f50fc1c3f3c0f3e5/modules/webadmin.cpp#L1089) up here?
2. Should this mimic `AddNetwork`/`DeleteNetwork` behavior and make direct `USERMODULECALL`s?

DarthGandalf: Could it reuse the logic instead of copying or mimicking it?

RotationMatrix: Maybe with an optional `bool renaming = false` parameter on `AddNetwork`/`DeleteNetwork`.

If only capitalization changed, should that count as a new network?
Are module paths case-sensitive?

DarthGandalf:
File paths in Linux are generally case sensitive

RotationMatrix:
Closing this for now. Will reopen if I find a solution.

PR #1761, opened 2020-10-20
Expand log module with a prototype web interface for viewing logs

### LINKED ISSUE #1334: [Feature Request} Webadmin module that shows user logs
As the title says, a module that shows logs on the web interface. Obviously users can **only** view their own logs on their **own** networks, but admins can view all logs on all networks. Users will have the option to enable this module or disable it for themselves, but either way, admins can view the logs.

braewoods:
This only allows users to view their own logs so it should be fairly
secure by that metric. It also checks that the requested directory
or log file is within the module's Save Path so it should be impossible
to read external files as well. This should close #1334.


braewoods:
@DarthGandalf Any idea when you might be able to review this? Thanks.

modules/data/log/tmpl/index.tmpl
```diff
@@ -0,0 +1,40 @@
...
+<li><a href="<? VAR ModPath TOP ?>?path=<? VAR Long ESC=HTML,URL ?>"><? VAR Short ?></a></li>
+<? ENDLOOP ?>
+</ul>
+</div>
+</div>
+<? ENDIF ?>
+
+<? IF Log ?>
+<h3>Log Viewer (Page <? VAR Page ?>)</h3>
+<div class="sectionbg">
```
DarthGandalf, line 19: Could you indent the lines here? It's hard to understand which div is inside which div

modules/data/log/tmpl/index.tmpl
```diff
@@ -0,0 +1,40 @@
...
+<? ENDIF ?>
+<? IF NextPage || NextPage == 0 ?>
+<button><a style="text-decoration: none;" href="<? VAR ModPath TOP ?>?path=<? VAR Path ESC=HTML,URL ?>&page=<? VAR NextPage ESC=HTML,URL ?>">Next</a></button>
+<? ENDIF ?>
+</div>
+</div>
+<? ENDIF ?>
+
+<? IF !Files && !Log ?>
+<p><b>Logs cannot be accessed from the Global Module.</b></p>
```
DarthGandalf, line 37: That's not what I see in the condition one line above

modules/data/log/tmpl/index.tmpl
```diff
@@ -0,0 +1,40 @@
...
+<h3>Log Viewer (Page <? VAR Page ?>)</h3>
+<div class="sectionbg">
+<div class="sectionbody">
+<code style="display: block; margin-bottom: 0.5em;">
+<? LOOP Log ?>
+<? VAR Line ?><br>
+<? ENDLOOP ?>
+</code>
+<? IF PrevPage || PrevPage == 0 ?>
+<button><a style="text-decoration: none;" href="<? VAR ModPath TOP ?>?path=<? VAR Path ESC=HTML,URL ?>&page=<? VAR PrevPage ESC=HTML,URL ?>">Prev</a></button>
```
DarthGandalf, line 27: Prev/Next should be translateable. In .tmpl files it can be done via `<? FORMAT "Foo"  ?>`

Same for other strings

modules/data/log/tmpl/index.tmpl
```diff
@@ -0,0 +1,40 @@
+<? I18N znc-log ?>
+<? INC Header.tmpl ?>
+
```
DarthGandalf, line 3: As log module has some settings which are not accessible via this UI, please mention that here. Or you can implement them if you wish.

modules/log.cpp
```diff
@@ -551,6 +555,74 @@ CModule::EModRet CLogMod::OnChanMsg(CNick& Nick, CChan& Channel,
     return CONTINUE;
 }
 
+bool CLogMod::OnWebRequest(CWebSock& WebSock, const CString& sPageName, CTemplate& Tmpl) {
+    if(GetType() == CModInfo::EModuleType::GlobalModule) {
+        return true;
+    }
+
+    CString sPath(WebSock.GetParam("path", false));
```
DarthGandalf, line 563: Why not `=`?

modules/log.cpp
```diff
@@ -551,6 +555,74 @@ CModule::EModRet CLogMod::OnChanMsg(CNick& Nick, CChan& Channel,
...
+
+    const size_t LINES_PER_PAGE = 512;
+    size_t Page = 0;
+    WebSock.GetParam("page", false).Convert(&Page);
+    const size_t Lines = Page * LINES_PER_PAGE;
+
+    if (File.IsReg()) {
+        CString Line;
+        File.Open();
+        for (size_t i = 0; i < Lines && File.ReadLine(Line); i++)
```
DarthGandalf, line 607: If user has to go through pages, the file will be read again and again. Maybe it's easier to show the whole file without paging?

braewoods: That was what I originally wanted to do but the comments on the first attempt to implement this raised concerns about large log files (> 512KB) so I thought it would be better to page it by line numbers. If you would prefer it that way I can change it back and remove the paging parts from the template.

modules/log.cpp
braewoods: I saw other code using this syntax for triggering the CString constructor. I can switch to triggering it via assignment if that is preferred.

modules/data/log/tmpl/index.tmpl
braewoods: Are you referring to the arguments I saw mentioned in the module pages where you enable the log module? I didn't even think of those since they already had a method of activation.

modules/data/log/tmpl/index.tmpl
braewoods: Ok. I'll see about doing that once I know what I will be revising in the template and module code.

modules/data/log/tmpl/index.tmpl
braewoods: Ok. I'll redo the indentation here after I have done some other revisions. I just wasn't too sure how I should do it since other templates don't seem to follow a consistent pattern.

modules/data/log/tmpl/index.tmpl
braewoods: Ok. Fair enough. While it does what I needed it to do I guess I should add a variable from the module to signal this condition. Or is there another way to check the module type from the template?

modules/data/log/tmpl/index.tmpl
DarthGandalf: No, not the arguments. Check `/msg *log help`

modules/data/log/tmpl/index.tmpl
DarthGandalf: I don't think so. A variable would work. Did you test this with 2 other types?

modules/log.cpp
DarthGandalf: It's the copy constructor from another CString. So I think it's more obvious with `CString sPath = ...;`

modules/log.cpp
DarthGandalf: I remember it now, yes.

Probably showing logs in web interface should have a better source than a plain text file, but something where you can say LIMIT and OFFSET like in SQL…

modules/data/log/tmpl/index.tmpl
braewoods: No I did not but I didn't see a reason to as the other 2 module types work the same anyway for what I was doing. I will work on this now.

modules/log.cpp
braewoods: Ok. I've revised the instances of where this occurred in my new code.

modules/log.cpp
braewoods: But the plain text file is what is actually written. SQL isn't an option for this without changing how logs are managed. And since the files are line oriented I have to perform manual seeking since the lines can end pretty much anywhere, rendering byte offsets worthless except for forward paging only. Though I could make the lines per page configurable as a GET parameter if this would help. The way I had it setup was the best way I could think of to keep the memory usage from getting out of hand.

What would you have me do?

DarthGandalf: What about counting bytes instead of lines per page? And when it got in a middle of a line, round to the nearest line ending, which should be near enough. Then byte offset could be set via GET

modules/data/log/tmpl/index.tmpl
braewoods: I have revised it to be more obvious about what the conditional is actually doing. I also restructured the other modules to an ELSE branch so it flows more logically.

modules/data/log/tmpl/index.tmpl
braewoods: I redid the indentation while I was redoing the template flow for the Global module. This should be resolved now.

modules/log.cpp
braewoods: That would work more or less but it has problems for "going back" as there's no way to know where the previous page positions were. But it works fine for going forward in one direction. I think we could leave "previous" as a function of the browser back button perhaps?

DarthGandalf: Offsets of previous pages could be passed through browser in a hidden field?

modules/data/log/tmpl/index.tmpl
braewoods: Ok. I think I'll try to implement these.

modules/log.cpp
braewoods: That would require a switch to POST requests since GET can't really handle that much data.

braewoods: Ok. This has been implemented as the new pagination method for the log files.

modules/data/log/tmpl/index.tmpl
braewoods: All user-visible text should now be translateable in the template save for ones where no translation is feasible.

modules/data/log/tmpl/index.tmpl
braewoods: The rules and settings configuration should now be usable from all 3 module modes but the log related functionality is not available to the global module.

PR #1768, opened 2020-11-25
Route TOPIC requests to client in route_replies

kylef:
This is very similar to the behaviour for route_replies with MODE, in which MODE requests (not change) are routed. Here I am routing any TOPIC request excluding TOPIC change only back to the client that requested it.

### DIFF
--- a/modules/route_replies.cpp
+++ b/modules/route_replies.cpp
@@ -183,6 +183,18 @@ static const struct {
         {"502", true}, /* rfc1459 ERR_USERSDONTMATCH */
         {nullptr, true},
        }},
+       {"TOPIC",
+        {
+         {"461", true}, /* rfc1459 ERR_NEEDMOREPARAMS */
+         {"403", true}, /* rfc1459 ERR_NOSUCHCHANNEL */
+         {"442", true}, /* rfc1459 ERR_NOTONCHANNEL */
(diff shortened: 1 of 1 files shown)


modules/route_replies.cpp
```diff
@@ -183,6 +183,18 @@ static const struct {
...
        }},
+       {"TOPIC",
+        {
+         {"461", true}, /* rfc1459 ERR_NEEDMOREPARAMS */
+         {"403", true}, /* rfc1459 ERR_NOSUCHCHANNEL */
+         {"442", true}, /* rfc1459 ERR_NOTONCHANNEL */
+         {"482", true}, /* rfc1459 ERR_CHANOPRIVSNEEDED */
+         {"331", true}, /* rfc1459 RPL_NOTOPIC */
+         {"332", false}, /* rfc1459 RPL_TOPIC */
+         {"333", true}, /* ircu? RPL_TOPICWHOTIME */
```
kylef, line 194: Are there any server's that don't emit `333`? It might be better to alter this to make `332` an end reply, `333` isn't defined in rfc1459 and thus ending on it might be incorrect as we may not be able to expect it from a server. However that leads to `333` to not get routed, so if we can reasonable assume that all implementations will return `333` then we can leave it like this. I am not sure so open to feedback to change this part.

Issue #1779, opened 2021-02-21
Roundtripping server-time on non UTC breaks


kylef:
If the system's timezone is not configured to UTC, then roundtripping server-time does not behave correctly. This can cause issues with playback replay when both the server, and the client are using server-time and the system TZ is set to something that isn't UTC.

Internally, the following should produce same result as the input regardless of TZ.

```
FormatServerTime(ParseServerTime("2011-10-19T16:40:51.620Z"))
```

Given the following patch to demonstrate:

```diff
diff --git a/test/UtilsTest.cpp b/test/UtilsTest.cpp
index 44514d98..e2f37c29 100644
--- a/test/UtilsTest.cpp
+++ b/test/UtilsTest.cpp
@@ -86,15 +86,16 @@ TEST(IRC32, SetMessageTags) {
 }
 
 TEST(UtilsTest, ServerTime) {
-    char* oldTZ = getenv("TZ");
-    if (oldTZ) oldTZ = strdup(oldTZ);
-    setenv("TZ", "UTC", 1);
-    tzset();
 
     timeval tv1 = CUtils::ParseServerTime("2011-10-19T16:40:51.620Z");
     CString str1 = CUtils::FormatServerTime(tv1);
     EXPECT_EQ(str1, "2011-10-19T16:40:51.620Z");
 
+    char* oldTZ = getenv("TZ");
+    if (oldTZ) oldTZ = strdup(oldTZ);
+    setenv("TZ", "UTC", 1);
+    tzset();
+
     timeval now = CUtils::GetTime();
 
     // Strip microseconds, server time is ms only
```

and then run:

```
$ env TZ=Europe/Prague make unittest

[ RUN      ] UtilsTest.ServerTime
../test/UtilsTest.cpp:92: Failure
Expected equality of these values:
  str1
    Which is: "2011-10-19T15:40:51.620Z"
  "2011-10-19T16:40:51.620Z"
    Which is: 0x477667c0174
```

The test case looks like it was implemented incorrectly, hiding this problem. Z in ISO 8601 should be UTC on both sides.

szlend:
I've experienced this issue myself, causing gaps in the playback log. I worked around the issue by setting `Environment=TZ=Etc/UTC` in the ZNC systemd service. It only happens on servers with the `server-time` extension (e.g. oragono) and only for received messages. Messages sent from ZNC are unaffected. With the `playback` module you can see this issue by running `/msg *playback list`. The last message timestamp will be offset by your timezone.

[...]

PR #1780, opened 2021-02-21
Fix parsing server-time from server to respect UTC

kylef:
This fixes the problems described in #1779 which causes problems to parsing server-time for server's which support server-time when the system is configured for a non UTC timezone.

This is because we use `mktime` which we use to convert our UTC time in `struct tm` parses it as the system configuration or TZ time instead of UTC. In formatting, we use `gmtime` which returns struct tm in UTC correctly.

`timegm` is the inverse of `gmtime`:

> The  functions  timelocal()  and timegm() are the inverses of localtime(3) and gmtime(3).  Both functions take a broken-down time and convert it to calendar time (seconds since the Epoch, 1970-01-01 00:00:00 +0000, UTC).  The difference  between  the  two functions is that timelocal() takes the local timezone into account when doing the conversion, while timegm() takes the input value to be Coordinated Universal Time (UTC)

While this fix does work, I am not sure 100% sure on the portability and if this will compile on all our target platforms, so that is to be determined.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -584,15 +584,28 @@ timeval CUtils::ParseServerTime(const CString& sTime) {
     struct tm stm;
     memset(&stm, 0, sizeof(stm));
     const char* cp = strptime(sTime.c_str(), "%Y-%m-%dT%H:%M:%S", &stm);
+
     struct timeval tv;
     memset(&tv, 0, sizeof(tv));
     if (cp) {
-        tv.tv_sec = mktime(&stm);
+        time_t time = mktime(&stm);
+
+        struct tm stm_gm;
+        gmtime_r(&time, &stm_gm);
+        time_t time_gm = mktime(&stm_gm);
+
+        struct tm stm_local;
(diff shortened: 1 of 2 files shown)


src/Utils.cpp
```diff
@@ -581,13 +581,24 @@ CString CUtils::FormatServerTime(const timeval& tv) {
...
-        tv.tv_sec = mktime(&stm);
+        time = mktime(&stm);
+
+        gmtime_r(&time, &stm_gm);
+        time_gm = mktime(&stm_gm);
+
+        localtime_r(&time, &stm_local);
+        time_local = mktime(&stm_local);
+
+        tv.tv_sec = mktime(&stm_local) + time_local - time_gm;
```
DarthGandalf, line 600: Can you explain this formula? 2*local-gm

DarthGandalf (changes requested):

kylef:
Fixed by https://github.com/znc/znc/pull/1857

PR #1783, opened 2021-03-28
fix foreground message to not imply that debug mode is on

satyanash:
Right now, running with `-f` or `--foreground` prints:
```
Staying open for debugging [pid: 251]
```
Which is kinda confusing and I was worried if some debug functionality was exposed.

This PR will print the above message only if `-D` or `--debug` is given.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -472,7 +472,7 @@ int main(int argc, char** argv) {
 
     if (bForeground) {
         int iPid = getpid();
-        CUtils::PrintMessage("Staying open for debugging [pid: " +
+        CUtils::PrintMessage("Running in foreground [pid: " +
                              CString(iPid) + "]");
 
         pZNC->WritePidFile(iPid);


src/main.cpp
```diff
@@ -471,11 +471,13 @@ int main(int argc, char** argv) {
...
-        int iPid = getpid();
-        CUtils::PrintMessage("Staying open for debugging [pid: " +
-                             CString(iPid) + "]");
+        if(CDebug::Debug()) {
+            int iPid = getpid();
+            CUtils::PrintMessage("Staying open for debugging [pid: " +
+                                CString(iPid) + "]");
 
-        pZNC->WritePidFile(iPid);
+            pZNC->WritePidFile(iPid);
```
psychon, line 479: I think this should always write the pid file.

Why don't you just change the message instead? Something like `Running in foreground` or so.

satyanash:
> Something like Running in foreground or so.
@psychon I have changed the message to `Running in foreground`. It will also always right the Pidfile.

kylef (approved):

PR #1785, opened 2021-05-21
Rewrite message parsing using string_view

DarthGandalf:
It's a bit too early yet to require C++17 so the implementation from
~~abseil-cpp~~BackportCpp is used instead.

Fixes https://crbug.com/oss-fuzz/34413 - slow message parsing on huge
messages. In real word, messages can't be that big, because CSocket
enforces a line length limit.

This can be considered a regression of 1.7.0, because before it, instead
of gathering params into a vector, code was searching 1st word in the
string, then 2nd word, then 3rd word, starting from beginning each time.
It was not very efficient, but the number of passes over the string was
limited.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,7 +284,6 @@ if(append_git_version)
 endif()
 
 
-
 file(GLOB csocket_files LIST_DIRECTORIES FALSE
 	"${PROJECT_SOURCE_DIR}/third_party/Csocket/Csocket.*")
 if(csocket_files STREQUAL "")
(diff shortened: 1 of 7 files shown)


DarthGandalf:
@jpnurmi FYI

psychon:
Where's the regression test that was previously painfully slow and now isn't? :-P

DarthGandalf:
Do you want me to add 1MB file to the repo?

psychon:
Dunno, I can't access the bug. Is it just random data? If so, the test could just generate something semi-random itself. Like, repeating the string `"a "` one million times, or so.

DarthGandalf:
Yeah, that would work. But I need to find some other implementation of string_view, because abseil doesn't work on cygwin apparently

PR #1735, opened 2020-07-05
Allow capabilities to be handled after upstream IRC registration

kylef:
This is perhaps the first part in being able to support [cap-notify](https://ircv3.net/specs/core/capability-negotiation#cap-notify) and capability 3.2 by extending the capability handling to occur after registration. I would like to add some automated testing, having some problems getting those passing (debugging setup with qt/python etc) so some delays with that.

The first change removes the `m_bAuthed` check, and then the second change makes it so that subsequent `CAP LS` won't trigger re-requesting already negotiated capabilities, otherwise the following would occur:

```

ZNC *raw CAP LS
:*raw!znc@znc.in PRIVMSG kyle :YOU -> [ZNC *raw CAP LS]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.812Z :irc-eu-2.darkscience.net CAP kyle LS :account-notify account-tag away-notify batch cap-notify chghost echo-message extended-join inspircd.org/poison inspircd.org/standard-replies invite-notify labeled-response multi-prefix sasl server-time setname userhost-in-names ]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.822Z :irc-eu-2.darkscience.net CAP kyle ACK :account-notify]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.831Z :irc-eu-2.darkscience.net CAP kyle ACK :away-notify]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.840Z :irc-eu-2.darkscience.net CAP kyle ACK :extended-join]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.848Z :irc-eu-2.darkscience.net CAP kyle ACK :multi-prefix]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.857Z :irc-eu-2.darkscience.net CAP kyle ACK :server-time]
:*raw!znc@znc.in PRIVMSG kyle :IRC -> [@time=2020-07-05T11:39:07.865Z :irc-eu-2.darkscience.net CAP kyle ACK :userhost-in-names]

ZNC *raw CAP LS
:*raw!znc@znc.in PRIVMSG kyle :YOU -> [ZNC *raw CAP LS]
[...]

PR #1790, opened 2021-06-01
Make PauseCap/ResumeCap no-op after registration

kylef:
To allow cap-notify and capabilities after registration. This is part of the road to support cap-notify in znc<>server by disabling the pausing logic of capabilities if we're registered. After registration there isn't a pause for capabilities (that's a pause to allow caps like SASL to operate before completing registration).

Extracted from  #1735

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -255,11 +255,17 @@ void CIRCSock::SendNextCap() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


DarthGandalf (approved):

DarthGandalf:
I don't know why you still didn't merge this, but is it still needed with #1859 ?

Issue #1770, opened 2020-12-02
API for clients to get ZNC info


strugee:
It would be awesome if ZNC had some way to query all of the user's networks, including credentials to connect to them. (Probably ZNC would have to generate and keep track of a token that could be handed out to clients, unless it stores user passwords in plaintext? Which hopefully it doesn't.)

This would allow clients to have a "sign in with ZNC" button or something like that so users wouldn't have to type in all their network details again. I'm thinking maybe this could be a `*api` virtual user that clients could PM or something? That way they could use their existing IRC stack and not have to vendor in an HTTP library or anything like that. Having the API run over IRC would also allow pushing events to the client in realtime, e.g. "the user deleted/added a network" or whatever, so the client could update its UI immediately without having to regularly poll. Whatever this looks like should be extensible in case we end up wanting to expose more meta-information in the future.

DarthGandalf:
This is an alternative to #515 

Sending such events from `*api` is easy, subject to deciding how the protocol should look like, of course. But managing such tokens - not so. Where would they be stored? What happens to them if ZNC is configured to use an external auth provider (e.g. via Cyrus SASL), and user changes the password there? Should clients logging in via the token have access to change the ZNC password?

I think limiting the token to a single time login only would be acceptable answer to these questions, where client needs to login using usual password to one network to be able to login to other networks next.

strugee:
> I think limiting the token to a single time login only would be acceptable answer to these questions, where client needs to login using usual password to one network to be able to login to other networks next.

This makes sense to me. Long-term I think it would be great to store the token persistently somewhere because it would move ZNC closer to a model more like modern webapps, where 3rd-party clients don't have to save the user's password and users can review the list of authorized clients and revoke individual clients if they want. But the devil's in the details and there are a _lot_ of sticky details, as you note. And we can always add that persistence later.

I wonder if it would make sense to have a well-known virtual network too, so clients wouldn't have to know _any_ of the user's network names? I.e. they'd connect to a "network" called `api` or whatever and that network would _only_ have the `*api` user on it, so the client could perform network discovery. Does that make sense/seem like a reasonable idea to you?

If I find some time today I might try to hack up a prototype of this in a module, though my C++ is a little rusty... so no promises :P

DarthGandalf:
> they'd connect to a "network" called api or whatever and that network

If I not mistake, if the requested network doesn't exist, the client will connect with no network. So choose some random name

> though my C++ is a little rusty

Perl or Python maybe?

PR #1803, opened 2021-08-05
Add WIP API module

strugee:
This adds a work-in-progress `*api` user that provides programmatic access to many aspects of ZNC. I intend to add the ability to set many of these values as well as query and set network information before I consider this fully useful.

I don't know if anyone is interested in pulling this in upstream, so I figured I would open this PR to see. If not I will likely maintain it as an external module. I would also be interested in code style feedback, as it's been ~a decade since I seriously wrote C++ and I'm rusty to say the least :P

Ref #1770


strugee:
Also, a little bit about the API design: it's supposed to feel a little bit like SMTP. I like the all-caps style of SMTP so that's what things are written in, but input is case-insensitive (so far; obviously user-specified strings like network names won't be). There's a notion of a top-level command, which is generally just, the thing you're doing. Currently there's just `QUERY` and `PING`,  but I expect to add `SET` and a few others in the future, e.g. cloning networks will probably be available via a toplevel command `CLONE` (though I haven't thought about it that much).

`QUERY` has a notion of scopes, which is basically what you'd expect - it's the general area in ZNC where you want to query values. In the code that's currently written scopes are static but I'm planning for each network to define a dynamic scope, so you'll be able to do e.g. `QUERY libera NICK` to get the nickname the user uses specifically on Libera.chat, assuming they have a network configured named `libera`.

Errors are borrowed from Unix because people tend to grok those already. The first space-delimited word is the general error code and then an explanation is provided in the rest of the string. The intention is that clients will be able to look at the general error code and do something useful even if they don't recognize the specific error string. The intention is to keep these strings and error codes stable over time.

Whether or not this gets pulled into ZNC core or I end up maintaining it out-of-tree, I intend to document the whole API on the ZNC wiki. Feedback on the API design or style would be super helpful.

strugee:
LGTM alert is spurious.

strugee:
Fixed the LGTM alert; thanks @DarthGandalf on IRC.

I've done some stuff that returns lists now. Open question: should the `LIST` type just become `TUPLELIST 1`? I feel like yes since there seems to be no reason for these to be separate. (It doesn't really matter for programmatic consumers obviously, but one goal of mine is to make this API essentially intuitively understandable/explorable by manually sending commands to the `*api` user and this would make the API more consistent.)

PR #1807, opened 2021-08-20
Added DHE-RSA-CHACHA20-POLY1305 to default SSL Ciphers

DoctorD90:


DarthGandalf:
You've missed the "why?" part of the description

src/Socket.cpp
```diff
@@ -36,7 +36,10 @@ static CString ZNC_DefaultCipher() {
     return "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:"
            "ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:"
            "ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:"
-           "DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384";
+           "DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:"
```
DarthGandalf, line 39: So this is not a copypaste from mozilla page anymore?

DoctorD90: Chacha20 is a good one, under certain aspects, better then AES too.
It has been implemented for ECDHE version too.
I checked a bit, but I find out no good reasons to exclude it.
It is still a "copy/paste", with just the add of Chacha20 for the DHE.

RealKindOne:
Should this be closed? This was added in https://github.com/znc/znc/commit/43ffd33dd2e5351c8c11b002f426e8cb8417f1aa.

Issue #1806, opened 2021-08-20
DelPort doesnt check if there is only 1 active port remaining


DoctorD90:
Via IRC messages is possible to completely remove all listening ports.

In the code, [here](https://github.com/znc/znc/blob/9cd36055166164d359a34c0af835e597e558cc59/src/ClientCommand.cpp#L1719), should be there an other "IF", checking if the number of listeners is = 1; in that case it should give back error, telling to add first another port before to remove the **last** available port.

https://github.com/znc/znc/blob/9cd36055166164d359a34c0af835e597e558cc59/src/ClientCommand.cpp#L1718-L1723

csprr:
The if should be to check if it's the current active port, because that one should never be able to be deleted (which is already the case in webadmin iirc)

DarthGandalf:
This is mostly by design. It should be possible to e.g. switch SSL on on the port without adding a new port with different number: delport, then addport. Of course, there's a potential to disconnect after delport, which would make ZNC not accessible anymore.

Webadmin does have this check, because all HTTP connections are independent (we always use `Connection: close` header, and even without it, we can't rely on browser to not close the connection between requests), so deleting the port from there would result in a weird error instead of the "success page"

PR #1809, opened 2021-08-20
Add IRC status commands EDITPORT and restrict DELPORT

csprr:
Possible fix (if wanted) for issue #1806
In "danger" of locking yourself out by removing your active port, this should not be possible (same as to HTTP) in my opinion.
This will allow the modification of an active port instead.


DarthGandalf:
Any tests?

PR #1816, opened 2021-11-15
Add support for draft solanum caps

A-UNDERSCORE-D:
While normally I'd be a bit dubious doing so, these caps explicitly only add tags, so this is really just allowing them to be passed though to clients that support them.


DarthGandalf:
#1859 allows doing this via modules

DarthGandalf:
Modules can add custom tags since ZNC 1.10

PR #1825, opened 2022-02-26
Initial HalfOps features

skralg:
Almost entirely copy/paste with small edits to function names.
Added autohalfop module, nearly the same as the autoop module.


include/znc/Modules.h
```diff
@@ -603,6 +603,16 @@ class CModule {
                          CChan& Channel, bool bNoChange);
     virtual void OnDeop(const CNick& OpNick, const CNick& Nick, CChan& Channel,
                         bool bNoChange);
+    /** Called when a nick is halfopped on a channel */
+    virtual void OnHalfOp2(const CNick* pOpNick, const CNick& Nick, CChan& Channel,
+                       bool bNoChange);
+    virtual void OnHalfOp(const CNick& OpNick, const CNick& Nick, CChan& Channel,
```
DarthGandalf, line 609: No need for this "2". The other function has it only for back compatibility with old modules.

include/znc/Modules.h
```diff
@@ -603,6 +603,16 @@ class CModule {
...
+                      bool bNoChange);
+    /** Called when a nick is dehalfopped on a channel */
+    virtual void OnDeHalfOp2(const CNick* pOpNick, const CNick& Nick,
+                         CChan& Channel, bool bNoChange);
+    virtual void OnDeHalfOp(const CNick& OpNick, const CNick& Nick, CChan& Channel,
```
DarthGandalf, line 614: If adding new callbacks, need to update modperl and modpython

modules/autohalfop.cpp
```diff
@@ -0,0 +1,646 @@
...
+};
+
+void CAutoHalfOpTimer::RunJob() { m_pParent->ProcessQueue(); }
+
+template <>
+void TModInfo<CAutoHalfOpMod>(CModInfo& Info) {
+    Info.SetWikiPage("autohalfop");
+}
+
+NETWORKMODULEDEFS(CAutoHalfOpMod, t_s("Auto halfop the good people"))
```
DarthGandalf, line 646: We shouldn't copypaste/duplicate code like that. A better way would be to have a single module which supports ops, voices, halfops

include/znc/Modules.h
skralg: modperl and modpython updated

modules/autohalfop.cpp
skralg: Perhaps that shall be my second contribution.

skralg:
Ready for another look

include/znc/Modules.h
DarthGandalf: You've missed the `function.in` files.

modules/autohalfop.cpp
DarthGandalf: Perhaps it shall.

I believe merging this PR can wait until that's done though.

skralg:
I've added a new 'automode' module that can do automatic arbitrary mode changes.

skralg:
@DarthGandalf Have you looked at the new automode module yet?

modules/automode.cpp
```diff
@@ -0,0 +1,420 @@
...
+            CheckAutoMode(Nick, Channel);
+        }
+    }
+
+    void OnMode(const CNick* pOpNick, const CNick& Nick, CChan& Channel,
+               bool bNoChange) {
+        if (Nick.GetNick() == GetNetwork()->GetIRCNick().GetNick()) {
+            const map<CString, CNick>& msNicks = Channel.GetNicks();
+            for (const auto& it : msNicks) {
+                if (!it.second.HasPerm(CChan::HalfOp)) {
```
DarthGandalf, line 194: Are you sure?

modules/automode.cpp
```diff
@@ -0,0 +1,420 @@
...
+        CString sHost = sLine.Token(2);
+        CString sMode = sLine.Token(3);
+
+        if (sHost.empty()) {
+            PutModule(
+                t_s("Usage: AddUser <user> <hostmask>[,<hostmasks>...] <mode> "
+                    "[channels]"));
+        } else if (sMode != "q" && sMode != "a" && sMode != "o" &&
+                   sMode != "h" && sMode != "v") {
+            PutModule("<mode> must be one of the following: qaohv");
```
DarthGandalf, line 217: You missed t_s

DarthGandalf (commented):
To clarify: it doesn't check the key like autoop did?

modules/automode.cpp
skralg: It seems to work fine.. what is t_s anyway?

modules/automode.cpp
skralg: Ah, good catch. That needs to become more robust.

skralg:
> 

I removed the key mechanism altogether because I don't know a single person who doesn't just use __NOKEY__.  Having to request it sort of defeats the purpose of calling it 'auto'...

DarthGandalf:
t_s is for localisation support. Visible messages should be translateable

Key is for efnet etc. You and another person add each other with the same
key, and the module cryptographically checks that the key is the same
instead of only relying on hostmask

сб, 5 мар. 2022 г., 03:01 Jonathan Rudolph ***@***.***>:

DarthGandalf:
It's still auto, because module requests it itself when nick and host match

сб, 5 мар. 2022 г., 03:01 Jonathan Rudolph ***@***.***>:

Issue #198, opened 2012-07-23
AddCommand() in python [$5]


DarthGandalf:

crocket:
I encountered the same issue....

crocket:
Why don't you circumvent the issue by implementing an independent command system in python?

DarthGandalf:
That would be a possible solution, yes.
As for why, because noone did it?

pcolmer:
Does this mean it isn't currently possible to do something like this:

`/msg *module command`

because it isn't possible to call AddCommand in a Python module?

DarthGandalf:
OnModCommand() is still available, but you need to parse the line yourself.

вт, 19 мар. 2019 г. в 11:53, Philip Colmer <notifications@github.com>:

pcolmer:
@DarthGandalf thanks. If I add OnModCommand and send a private message to my module, I get:

`This module doesn't implement any commands.`

which seems to suggest that I still need to call AddCommand somehow, but I'm struggling with the parameters to give it. The call is insisting that I pass it a CModCOmmand::ModCmdFunc.

What am I misunderstanding here?

DarthGandalf:
You don't just add it. You need to implement it. The default implementation
uses commands which were added via AddCommand().

CModCommand::ModCmdFunc is exactly why AddCommand doesn't work currently,
this is what this issue is about - to make AddCommand work.

вт, 19 мар. 2019 г. в 13:52, Philip Colmer <notifications@github.com>:

pcolmer:
I'm sorry - I don't understand.

If AddCommand doesn't work then how do I implement OnModCommand? The code for the `pyeval` module just adds the function and has code to do processing when the function is called. I've added a definition for OnModCommand but it doesn't get called - I just get that error message displayed.

```
    def OnModCommand(self, message):
        print("OnModCommand")
```

pcolmer:
Hmm … I've just realised that my `OnModCommand` **is** getting called as the `print` statement is being executed.

So what do I need to do to tell ZNC that the call has been handled? It is a `void` function and this is partly why I'm confused about this.

pcolmer:
OK … if I call something like PutModule, I stop getting the error message about there not being any commands implementation. I think I'm OK with this now.

PR #1832, opened 2022-07-03
modules: modpython: Implement Module.AddCommand()

ernestask:
Currently, there is no usable wrapper for `CModCommand` for use within
bindings, so this commit adds a proxy class that adds itself as a
callback and allows implement Python classes to implement commands via
`__call__()`.

A completely synthetic example:
```python
import znc

class foo(znc.Module):
    module_types = [znc.CModInfo.UserModule]

    def OnLoad(self, args, message):
        self.AddHelpCommand()
        self.AddCommand(FooCmd)

        return True

class FooCmd(znc.Command):
    cmd = 'foo'
    args = foo.t_d('bar')
    desc = foo.t_d('baz')

    def __call__(self, line):
        self.GetModule().PutModule('I have been foo’d!')
```

Fixes https://github.com/znc/znc/issues/198


DarthGandalf:
Thanks! Can you add a test to https://github.com/znc/znc/blob/master/test/integration/tests/scripting.cpp with such synthetic example?

ernestask:
> Thanks! Can you add a test to https://github.com/znc/znc/blob/master/test/integration/tests/scripting.cpp with such synthetic example?

Added a test.

PR #1839, opened 2022-11-03
Send 'CAP END' to close off the 'CAP LS' command

lag-google:


DarthGandalf:
Did something not work for you?

dgw:
> Did something not work for you?

If I'm reading the patch right (possibly not because I'm looking in GitHub Mobile), this would send CAP END before any negotiations happen, which is probably not wanted.

DarthGandalf:
Yeah, I mean why the patch was deemed necessary at all.

lag-google:
An IRC server was prematurely disconnecting after receiving the USER command.

The RFC says that the CAP END command must be sent to close off CAP LS.

The Admin for the IRC server in question said that the server was hanging whilst waiting for CAP END.

This turned out not to be the reason for the disconnect, so the patch has been dropped.

dgw:
Yes, the [capability negotiation spec](https://ircv3.net/specs/extensions/capability-negotiation.html) says that the client SHOULD either send CAP LS _or_ blindly CAP REQ what it wants, and then MUST send the standard NICK and USER commands. A server that doesn't handle those while waiting for CAP END is violating the spec.

ZNC sends CAP END later in the process. Adding it in this place would have made ZNC noncompliant.

lag-google:
Understood.  Thank you for your valuable feedback.

PR #1841, opened 2022-11-05
Add support SWIG 4.1.0, drop support for < 4.0.2

DarthGandalf:
https://bugs.gentoo.org/878587

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,7 +173,7 @@ if(WANT_PYTHON AND NOT EXISTS
 	endif()
 endif()
 if(search_swig)
-	find_package(SWIG 3.0.0)
+	find_package(SWIG 4.0.1)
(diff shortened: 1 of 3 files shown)


PeGaSuS-Coder:
Unsure how this affects Ubuntu, but `Ubuntu 20.04 LTS` which has `End of Standard Support (ESS)` on April 2025 and `End of Life (EOL)` on April 2030 still has Swig 4.0.1.
```
# apt show swig4.0
Package: swig4.0
Version: 4.0.1-5build1
```

DarthGandalf:
Yeah, I've updated the number in CMakeLists to 4.0.1, because Ubuntu 20.04 is used in our CI.

I'm increasing this number only because I can't test what happens with older versions, and there's an easy escape hatch anyway, when using tarballs.

I just didn't update the title of the PR

PR #1843, opened 2022-11-07
`clientnotify.cpp`: add new option and test existing code

### LINKED ISSUE #1840: `clientnotify.cpp`: shouldnt notify for same ip & identifier but different network
I just installed ZNC and configured two networks, each time I connected to their fake hosts as [`chris@identifier/servername`](https://github.com/christian-heusel/dotfiles/blob/d14d72138b4c2cb1bf5004456eba796fbe4a6d02/roles/irc-configuration/templates/irc.conf#L197) as per documentation.

Now each time I am opening my IRC client I seem to get a notification that another client authenticated:

```
00:15 *status ╡ Another client authenticated as your user. Use the 'ListClients' command to see all 2 clients.
00:25    name ╡ ListClients
00:25 *status ╡ +-------------+-----------+------------+
00:25       ⤷ ╡ | Host        | Network   | Identifier |
00:25       ⤷ ╡ +-------------+-----------+------------+
00:25       ⤷ ╡ | 141.70.80.5 | libera    | pioneer    |
00:25       ⤷ ╡ | 141.70.80.5 | robustirc | pioneer    |
00:25       ⤷ ╡ +-------------+-----------+------------+
```

This seems a bit overly noisy to me, and I would propose to filter if two clients use the same Identifier there should not be another notification about them rejoining.

This issue is somewhat related to https://github.com/znc/znc/issues/755 but not a duplicate as I am also using the same Identifier for both clients.

The code in question is the following I guess, I am happy to code together a fix for this if I get positive feedback about the idea:

https://github.com/znc/znc/blob/e0ffdddd473e97cb843f2bc8ad4fa16cf47c65b4/modules/clientnotify.cpp#L77-L91

**ZNC Version:**
Docker image: https://hub.docker.com/_/znc/ 
Tag: latest
```
00:32       ⤷ ╡ ZNC 1.8.2 - https://znc.in
00:32       ⤷ ╡ IPv6: yes, SSL: yes, DNS: threads, charset: yes, i18n: yes, build: cmake
```

christian-heusel:
This PR adds:
- an integration test for [`clientnotify module`](https://github.com/znc/znc/blob/master/modules/clientnotify.cpp)
- a new option `NewNotifyOn` to the module which determines whether notifications should be sent when a client with a new IP, new identifier or for both cases should be sent
- the options to use identifiers in the integration tests

Questions for the reviewer:
- Should the Code be formatted given the `.clang-format` config? If yes I will push the reformatted code in another commit
- What do you think about the naming of the new option?
- Should the one [text] option be split into two boolean ones or stay as is?

Fixes #1840

### DIFF
--- a/modules/clientnotify.cpp
+++ b/modules/clientnotify.cpp
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#include <znc/znc.h>
 #include <znc/User.h>
+#include <znc/znc.h>
 
 using std::set;
 
@@ -57,10 +57,12 @@ class CClientNotifyMod : public CModule {
                    t_d("Turns notifications for unseen connections on or off"),
                    [=](const CString& sLine) { OnNewOnlyCommand(sLine); });
         AddCommand("NotifyOnNewIP", t_d("<on|off>"),
-                   t_d("Specifies whether you want to be notified about new connections with new IPs"),
+                   t_d("Specifies whether you want to be notified about new "
+                       "connections with new IPs"),
                    [=](const CString& sLine) { OnNotifyOnNewIP(sLine); });
         AddCommand("NotifyOnNewID", t_d("<on|off>"),
-                   t_d("Specifies whether you want to be notified about new connections with new IDs"),
+                   t_d("Specifies whether you want to be notified about new "
+                       "connections with new IDs"),
                    [=](const CString& sLine) { OnNotifyOnNewID(sLine); });
         AddCommand(
             "OnDisconnect", t_d("<on|off>"),
@@ -98,23 +100,27 @@ class CClientNotifyMod : public CModule {
         }
 
         auto sendLoginNotification = [&]() {
-            SendNotification(t_p("<This message is impossible for 1 client>",
-                                 "Another client ({1}) authenticated as your user. "
-                                 "Use the 'ListClients' command to see all {2} "
-                                 "clients.",
-                                 GetUser()->GetAllClients().size())(
-                sClientNameMessage, GetUser()->GetAllClients().size()));
+            SendNotification(
+                t_p("<This message is impossible for 1 client>",
+                    "Another client ({1}) authenticated as your user. "
+                    "Use the 'ListClients' command to see all {2} "
+                    "clients.",
+                    GetUser()->GetAllClients().size())(
+                    sClientNameMessage, GetUser()->GetAllClients().size()));
         };
 
         if (m_bNewOnly) {
             // see if we actually got a new client
-            // TODO: replace setName.find(...) == setName.end() with !setName.contains() once ZNC uses C++20
-            if ((m_bNotifyOnNewIP       && (m_sClientsSeenIP.find(sRemoteIP)       == m_sClientsSeenIP.end())) ||
-                (m_bNotifyOnNewClientID && (m_sClientsSeenID.find(sRemoteClientID) == m_sClientsSeenID.end()))) {
+            // TODO: replace setName.find(...) == setName.end() with
+            // !setName.contains() once ZNC uses C++20
+            if ((m_bNotifyOnNewIP && (m_sClientsSeenIP.find(sRemoteIP) ==
+                                      m_sClientsSeenIP.end())) ||
+                (m_bNotifyOnNewClientID &&
+                 (m_sClientsSeenID.find(sRemoteClientID) ==
+                  m_sClientsSeenID.end()))) {
                 sendLoginNotification();
             }
-        }
-        else {
+        } else {
             sendLoginNotification();
         }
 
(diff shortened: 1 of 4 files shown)


DarthGandalf:
That's not your fault, appveyor broke in master branch too, for some reason... I don't understand what broke there though

modules/clientnotify.cpp
```diff
@@ -22,14 +22,19 @@ using std::set;
 class CClientNotifyMod : public CModule {
   protected:
     CString m_sMethod;
+    CString m_sNewNotifyOn;
```
DarthGandalf, line 25: This var is not needed, as you have 2 bools for it already

modules/clientnotify.cpp
```diff
@@ -76,18 +89,37 @@ class CClientNotifyMod : public CModule {
 
     void OnClientLogin() override {
         CString sRemoteIP = GetClient()->GetRemoteIP();
-        if (!m_bNewOnly ||
-            m_sClientsSeen.find(sRemoteIP) == m_sClientsSeen.end()) {
+        CString sRemoteClientID = GetClient()->GetIdentifier();
+
+        CString& sClientNameMessage = sRemoteIP;
+        if (m_bNotifyOnNewClientID and sRemoteClientID != "") {
+            sClientNameMessage = sRemoteClientID;
```
DarthGandalf, line 96: I think it would be more useful to show IP always, and id if client sent one. Now id would hide the ip

modules/clientnotify.cpp
```diff
@@ -76,18 +89,37 @@ class CClientNotifyMod : public CModule {
...
-
-            // the set<> will automatically disregard duplicates:
-            m_sClientsSeen.insert(sRemoteIP);
+                sClientNameMessage, GetUser()->GetAllClients().size()));
+        };
+
+        if (m_bNewOnly) {
+            // see if we actually got a new client
+            // TODO: replace setName.find(...) == setName.end() with !setName.contains() once ZNC uses C++20
+            if ((m_bNotifyOnNewIP       && (m_sClientsSeenIP.find(sRemoteIP)       == m_sClientsSeenIP.end())) ||
```
DarthGandalf, line 111: Why so many spaces before &&?

modules/clientnotify.cpp
```diff
@@ -76,18 +89,37 @@ class CClientNotifyMod : public CModule {
...
+                (m_bNotifyOnNewClientID && (m_sClientsSeenID.find(sRemoteClientID) == m_sClientsSeenID.end()))) {
+                sendLoginNotification();
+            }
+        }
```
DarthGandalf, line 115: no \n (clang-format should fix this automatically)

test/integration/tests/modules.cpp
```diff
@@ -54,6 +54,50 @@ TEST_F(ZNCTest, NotifyConnectModule) {
...
+TEST_F(ZNCTest, ClientNotifyModule) {
+    auto znc = Run();
+    auto ircd = ConnectIRCd();
+    auto client = LoginClient();
+    client.Write("znc loadmod clientnotify");
+    client.ReadUntil("Loaded module");
+
+    auto check_not_sent = [](Socket& client, QString wrongAnswer){
+        auto result = client.ReadRemainder();
+        EXPECT_FALSE(result.contains(wrongAnswer.toUtf8())) << "Got an answer even though we didnt want one with the given configuration";
```
DarthGandalf, line 66: Either add << result, or use EXPECT_THAT(result, Not(HasSubstr()) for a better error message

modules/clientnotify.cpp
```diff
@@ -76,18 +89,37 @@ class CClientNotifyMod : public CModule {
...
-                GetUser()->GetAllClients().size()));
-
-            // the set<> will automatically disregard duplicates:
-            m_sClientsSeen.insert(sRemoteIP);
+                sClientNameMessage, GetUser()->GetAllClients().size()));
+        };
+
+        if (m_bNewOnly) {
+            // see if we actually got a new client
+            // TODO: replace setName.find(...) == setName.end() with !setName.contains() once ZNC uses C++20
```
DarthGandalf, line 110: Yeah, probably it's time to bump the C++ version further. C++20 is too early to use though

modules/clientnotify.cpp
christian-heusel: Will be fixed when running clang-format

modules/clientnotify.cpp
christian-heusel: Will be fixed when running clang-format

test/integration/tests/modules.cpp
christian-heusel: ✅

modules/clientnotify.cpp
christian-heusel: Do you create an Issue for this or should I?

modules/clientnotify.cpp
christian-heusel: I have now split upd the option `NewNotifyOn` into `NotifyOnNewIP` and `NotifyOnNewID`, so the string is removed!

modules/clientnotify.cpp
```diff
@@ -76,18 +92,41 @@ class CClientNotifyMod : public CModule {
...
+
+        if (m_bNewOnly) {
+            // see if we actually got a new client
+            // TODO: replace setName.find(...) == setName.end() with
+            // !setName.contains() once ZNC uses C++20
+            if ((m_bNotifyOnNewIP && (m_sClientsSeenIP.find(sRemoteIP) ==
+                                      m_sClientsSeenIP.end())) ||
+                (m_bNotifyOnNewClientID &&
+                 (m_sClientsSeenID.find(sRemoteClientID) ==
+                  m_sClientsSeenID.end()))) {
```
christian-heusel, line 120: TBH, i am pretty unhappy with the way clang-format formats this 🤔

PR #1845, opened 2022-11-23
CI: add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where ZNC already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and also moves the fuzzing closer into the development workflow. CIFuzz has a variety of features which can be found in the URL above. In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'znc'
+        dry-run: false
+        language: c++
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Thanks, how does it know to pass `cmake -DBUILD_SHARED_LIBS=OFF -DWANT_ICU=OFF -DWANT_OPENSSL=OFF -DWANT_ZLIB=OFF -DWANT_IPV6=OFF ..` ?

DavidKorczynski:
> Thanks, how does it know to pass `cmake -DBUILD_SHARED_LIBS=OFF -DWANT_ICU=OFF -DWANT_OPENSSL=OFF -DWANT_ZLIB=OFF -DWANT_IPV6=OFF ..` ?

It's based on the set up here: https://github.com/google/oss-fuzz/blob/master/projects/znc/build.sh

DarthGandalf:
I see. I've added a comment with that link to this file.

PR #1847, opened 2022-12-16
Check for github action updates monthly

jlosito:
This will use dependabot to check for updates against any github action which this project uses on a monthly basis and submit pull requests with version bumps.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


DarthGandalf:
What dependencies would that be? Everything is taken from the user system (or the distro builder... where ZNC is built)

jlosito:
@DarthGandalf This change would only manage these dependencies > https://github.com/znc/znc/blob/29142359cd8b2c2bd805776ce70ddc3bec8d2abe/.github/workflows/build.yml#L12

The latest version the checkout action is on is v3. Some of the actions are worth upgrading from my experience, especially the ones that add caching and such in later versions, specifically the setup-* actions.

PR #1853, opened 2022-12-21
Fix ParseServerTime to use UTC

### LINKED ISSUE #1773: znc.in/self-message not working
I've ran into a slight issue with znc.in/self-message. 

Lately, I've noticed that I'm failing to get messages from myself when I'm receiving playback using the clientbuffer mod. Originally, I noticed this bug because the feature documented here: https://wiki.znc.in/Query_buffers does not work at all on IRC networks running Oragono as an ircd. In addition, something lately has made it so the feature no longer works on Snoonet, but it works on Freenode. 

I've tested this using znc 1.8.2 on macOS Mojave, Ubuntu and Fedora with Textual and Palaver as my clients (both of which are listed to support znc.in/self-message)

Steps to reproduce:

1. Make a fresh ZNC installation (I've chosen 1.8.2 from Ubuntu repositories)
2. Set up a network (I've chosen Snoonet)
3. Disable the auto clearing of queries and buffers
4. Install the clientbuffer mod and configure 2 clients (Mine are `Textual` and `Textual2`)
5. Chat normally on client 1
6. Connect client 2 and fail to see your own messages

johnfzc:
I'm not at all certain if this fix will meet your standards, but it does resolve #1773. It's based on other timezone manipulation code in `Utils.cpp`.

Please let me know if changes are needed.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -587,7 +587,22 @@ timeval CUtils::ParseServerTime(const CString& sTime) {
     struct timeval tv;
     memset(&tv, 0, sizeof(tv));
     if (cp) {
+        char* oldTZ = getenv("TZ");
+        if (oldTZ) oldTZ = strdup(oldTZ);
+        setenv("TZ", "UTC", 1);
+        tzset();
+
         tv.tv_sec = mktime(&stm);
+
+        // restore old value
+        if (oldTZ) {
+            setenv("TZ", oldTZ, 1);
+            free(oldTZ);
+        } else {
+            unsetenv("TZ");
+        }
+        tzset();
+
         CString s_usec(cp);
         if (s_usec.TrimPrefix(".") && s_usec.TrimSuffix("Z")) {
             tv.tv_usec = s_usec.ToULong() * 1000;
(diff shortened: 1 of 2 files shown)


psychon:
Are you perhaps looking for `timegm`? That's `mktime` but for an UTC input.

Sadly, that's a GNU extension. `man timegm` says:

> These functions are nonstandard GNU extensions that are also present on the BSDs.

No idea whether using this is acceptable.

DarthGandalf:
@psychon From what I recall, some non-standard functions didn't exist on Cygwin's libc. I don't remember whether it was this function, or some other function

johnfzc:
After some discussion on IRC I will update this PR to include tests, move to better C++20 time will occur in a future version, perhaps 1.9.

johnfzc:
I've updated this PR to include unit tests. The new unit test failed without the ParseServerTime patch and now pass.

DarthGandalf:
If FormatServerTime and ParseServerTime are opposite of each other, shouldn't the other one have something similar too?

johnfzc:
I think the existing tests pass because of this code: https://github.com/znc/znc/pull/1853/files#diff-28bf8e015476ffb578a318f1d6afabf7007e4f1fb9943477003fa1ecf4f03e64R89-R92 . It's also possible we don't have problems as most IRC servers send the time in UTC.

I can make further improvements if you would like, but I'll need more specific instruction to do so. As far as I know from my testing this is the only timezone issue that exists. Timezone handling in C++ is a minefield, and there could easily be other problems that we don't know about.

Issue #1861, opened 2023-01-17
Crypt Module - IRC Clients that do not keep "sent" text and instead read the IRC response will not display properly


zach-snell:
IRC Clients such as [TheLounge](https://thelounge.chat/) will not properly show the sent message unless IRC responds to them with a correct response message.

This was causing several issues including messages never showing up when sent to crypt PM or showing up as encrypted in channel/un-prefixed PM.

This fix is quite simple: 
When sending to ZNC clients, ensure the response payload is decrypted and in the right channel when there is an encrypted payload.

I wrote a patch for this and while doing this - I also combined / clarified some logic and updated deprecated functions.
Feel free to use my fix or advise for changes/etc. Will link to issue.

If you don't like my other updates, could trim it down just to the band-aid in "OnSendToClients" with a check if its the right message (inbound, from self). It does have to decrypt for every client though because I can't edit the buffer in current implementation.

zach-snell:
In further reading - I may withdraw this. I think it violates [IRC spec](https://ircv3.net/specs/extensions/echo-message.html).
Since TheLounge is leveraging echo-message (which is what creates the self message) spec says the message must be exactly as sent to the IRC Server. Therefore, me decrypting it to send back to the client would be a violation.

I suppose there may need to be a fix on TheLounge instead. 

The remainder of the cleanup may be useful if desired. I can branch and open a PR just cleaning up crypt.cpp if you like the cleanup.

zach-snell:
Won't fix for now due to my fix breaking standard. Will withdraw PR. Will reopen PR specific to Crypt.cpp cleanup and let maintainers decide if they want that at some point.

Will instead look into alternative fixes.

PR #1862, opened 2023-01-17
Fix for clients that do not write typed text and instead await IRC response (ex. TheLounge)

zach-snell:
Fix for #1861


zach-snell:
I'll take a look at the failed test tomorrow. I did run them locally but will make sure it ran them all.

Issue #1842, opened 2022-11-05
Add ircv3 extended join account into to logging


khaytsus:
ZNC already handles extended join nicely in that it passes it along to clients like weechat etc, that all works fine, but I would love to see it logged in the znc logs.

Of course this is the typical join log from znc:

[22:18:19] *** Joins: somenick (~toor@zncrools.com)

Perhaps something like?

[22:18:19] *** Joins: somenick (~toor@zncrools.com accountname)

And if no account name, just don't add the (space)$accountname bit?

pajamian:
Created PR #1870 to address this.

PR #1870, opened 2023-08-09
Add account to joins for the log module.

pajamian:
This commit adds the account name for identified users to "Joins" lines in logs generated by the log module.  It can get the account name from either the account tag (if the account-tag capability is requested) or the extended-join info (if the extended-join capability is requested).  The current version of ZNC requests both, but this feature will still work if the IRC server only supports one or the other.

### DIFF
--- a/modules/log.cpp
+++ b/modules/log.cpp
@@ -99,7 +99,7 @@ class CLogMod : public CModule {
                 const CString& sMessage) override;
     void OnQuit(const CNick& Nick, const CString& sMessage,
                 const vector<CChan*>& vChans) override;
-    void OnJoin(const CNick& Nick, CChan& Channel) override;
+    void OnJoinMessage(CJoinMessage& Message) override;
     void OnPart(const CNick& Nick, CChan& Channel,
                 const CString& sMessage) override;
     void OnNick(const CNick& OldNick, const CString& sNewNick,
@@ -457,12 +457,26 @@ CModule::EModRet CLogMod::OnSendToIRCMessage(CMessage& Message) {
     return CONTINUE;
 }
 
-void CLogMod::OnJoin(const CNick& Nick, CChan& Channel) {
-    if (NeedJoins()) {
-        PutLog("*** Joins: " + Nick.GetNick() + " (" + Nick.GetIdent() + "@" +
-                   Nick.GetHost() + ")",
-               Channel);
+void CLogMod::OnJoinMessage(CJoinMessage& Message) {
+    if (!NeedJoins())
+        return;
+
+    const CNick& Nick = Message.GetNick();
+    CChan& Channel = *Message.GetChan();
+    CString Account = Message.GetTag("account");
+    const char* s = " ";
+
+    if (Account.empty())
+        Account = Message.GetParam(1);
+
+    if (Account.empty() || Account == "*") {
+        Account = "";
+        s = "";
     }
+
+    PutLog("*** Joins: " + Nick.GetNick() + " (" + Nick.GetIdent() + "@" +
+           Nick.GetHost() + ")" + s + Account,
+           Channel);
 }
 
 void CLogMod::OnPart(const CNick& Nick, CChan& Channel,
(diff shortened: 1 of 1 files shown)


pajamian:
This addresses issue #1842

khaytsus:
I've been running this for a while now and have had no issues, doing exactly what it's supposed to and seems like a very clean solution.

modules/log.cpp
```diff
@@ -457,12 +457,26 @@ CModule::EModRet CLogMod::OnSendToIRCMessage(CMessage& Message) {
...
-        PutLog("*** Joins: " + Nick.GetNick() + " (" + Nick.GetIdent() + "@" +
-                   Nick.GetHost() + ")",
-               Channel);
+void CLogMod::OnJoinMessage(CJoinMessage& Message) {
+    if (!NeedJoins())
+        return;
+
+    const CNick& Nick = Message.GetNick();
+    CChan& Channel = *Message.GetChan();
+    CString Account = Message.GetTag("account");
```
DarthGandalf, line 466: please add `s` prefix to the string for consistency with the rest: sAccount

modules/log.cpp
```diff
@@ -457,12 +457,26 @@ CModule::EModRet CLogMod::OnSendToIRCMessage(CMessage& Message) {
...
+    const char* s = " ";
+
+    if (Account.empty())
+        Account = Message.GetParam(1);
```
DarthGandalf, line 470: Perhaps there should be some kind of `Message.GetAccount()` which handles both the tag and extended-joins. But that can wait until there's some other user of that info, for now doing it here is fine. Add a TODO though

modules/log.cpp
pajamian: Okay, I'll push that out soon.

modules/log.cpp
pajamian: The issue here is that the extended-joins param is only valid for a JOIN message, so if we have a method that attempts to get it from generic messages then we need to explicitly check that it's actually a JOIN message before attempting to get extended-join info.

modules/log.cpp
```diff
@@ -457,12 +457,26 @@ CModule::EModRet CLogMod::OnSendToIRCMessage(CMessage& Message) {
...
+
+    if (Account.empty() || Account == "*") {
+        Account = "";
+        s = "";
     }
+
+    PutLog("*** Joins: " + Nick.GetNick() + " (" + Nick.GetIdent() + "@" +
+           Nick.GetHost() + ")" + s + Account,
```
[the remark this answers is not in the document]
pajamian: Not very obvious, but at least it's there.  I think an overriding concern is to avoid impacting on any existing attempt at parsing the log line as much as possible, so putting it at the end makes sense in that respect.

modules/log.cpp
pajamian: I think I may have misunderstood, I put the space as a separate string to avoid the extra overhead of having to prepend a space onto the Account string.  I can certainly do that but I just felt this was a bit more efficient this way.

DarthGandalf: You have misunderstood. I know why you made `s` variable separately. I meant `sAccount` literally. All CString variables here start with symbol `s`

pajamian: Oh, that's actually what I originally thought.  Okay I'll change that.

modules/log.cpp
DarthGandalf: Yes, this needs some consideration how to do it better, and TODO for now will be enough

modules/log.cpp
pajamian: Done, it builds but I'm still testing.  I don't anticipate any issues with a one letter change to a variable name, though.

DarthGandalf: Of course

pajamian: Tested with the variable name change and still works.

modules/log.cpp
pajamian: TODO added.

Issue #322, opened 2013-04-16
Some messages not buffered if client disconnects without /quit and then times out


LunNova:
If I disconnect from znc due to an issue with my internet connection, some buffer may be lost, leading to confused conversations as someone assumes you saw their reply.

It would be nice to have an option to make znc keep a buffer of what was said between now and the last successful ping, and add that to the buffer if the client disconnects due to missing a ping, or E_CONN_RESET/other similar error.
##

winny-:
Has any headway been made in relation to this issue?

jpnurmi:
This is the solution that I've been happily using for a year: http://wiki.znc.in/Playback. Yes, it needs special support from your client, but that's not difficult to implement.

Mikaela:
> Yes, it needs special support from your client, but that's not difficult to implement.

Is this currently supported by anything else than Communi-\*?

jpnurmi:
> Is this currently supported by anything else than Communi-*?

Textual & Mobile Colloquy

caktux:
The OP's proposed solution would be a very welcomed feature.

PotcFdk:
Assuming I'm not missing any important technical detail that would lead to issues, I agree with @caktux, as this solution would (hopefully) work with all standard clients and not require special plugins.

Mikaela:
Could this be closed based on \*playbck or will this wait for it to be merged to ZNC?

PotcFdk:
I'm not entirely sure if a solution to this issue should require non-standard client implementations and/or (third-party) plugins.

AndrioCelos:
Isn't this what the `clearbufferonmsg` module is for?

glechapelain:
It never checks whether the actual `write(...)' call succeeded :(
I'll see if I can come up w/ sthing!

glechapelain:
o/ o/

So, I **think** the issue is fixed by the commit I made in my fork (Idk if it shows here for others.)

However, only when using the `clientbuffer` module.  I **could** carry on looking for fixing w/o, but that wouldn't be solving a problem I have. 

I'll send a PR for now with that, at least it is something.

glechapelain:
> Isn't this what the `clearbufferonmsg` module is for?

So I guess yes, it is, after educating myself.
(At first I thought `write()` or `SSL_write()` returning <= 0 meant that the message did not make it, and therefore the other way around meant that it did however as it turns out `SSL_write()` may still return a # of byte transmitted even though the connection is down, and for good.)

[...]

PR #1877, opened 2023-09-23
Fix for issue #322: message loss upon spurious disconnect

glechapelain:
When the connection shuts down, we never check for the result of the call to `write()`.

This change does so, and change the logic to only call the module callbacks upon it succeeding.

This change the behaviour/expectation on these callbacks, but I saw them only used in the `clientbuffer` module.
It does not mean they are not used by some peeps at home w/o our knowing.

Also, the change fixes the issue only when the clientbuffer module is used.
But Idk either is the issue even happens without using it.  I had have to make more digging there


src/Client.cpp
```diff
@@ -568,12 +568,14 @@ bool CClient::PutClient(const CMessage& Message) {
     Msg.SetClient(this);
     Msg.SetNetwork(m_pNetwork);
 
-    bool bReturn = false;
-    NETWORKMODULECALL(OnSendToClientMessage(Msg), m_pUser, m_pNetwork, this,
-                      &bReturn);
-    if (bReturn) return false;
+    bool bReturn = PutClientRaw(Msg.ToString());
```
DarthGandalf, line 571: You destroyed the ability of modules to block sending some things to client

glechapelain: Agreed: I think like the real fix is creating a new callback
Sorry I should have said that (and perhaps have my change in a single clean commit.)
I don't really know the process or who looks and when so cheers for taking a peek.

But now I want to give a whirl to the vanilla ZNC too (no clientbuffer.)  From the issue it sounds like it isn't fixed..  

But if the buffer is updated based on response from the client, then it should have always worked ?

I'll look at the code.

PR #1878, opened 2023-09-24
Fix for #322:

glechapelain:
+ Add a OnSentToClientMessage callback, only called when `write-to-socket' succeeds = Change `CClient::PutClient{,Raw}' to return false upon fail of that write.

The rest of the change is in the `clientbuffer' module.

Questions: Is it OK? (the change)?

Observations:
 . If so we could go further and remove `PutClientRaw()' altogether to clean-up.
 . It starts to be a lot of filtering
 . I'm not sure why it isn't used in `single-player'.
   I suppose then we merely clear the buffer upon connection from the client.

This would go along with: https://github.com/CyberShadow/znc-clientbuffer/pull/15


glechapelain:
I'm still missing messages.  Something else might be updating that timestamp ?

glechapelain:
This isn't sound the `write()` to a socket cannot be used to verify that the message got transmitted (I wasn't aware of that.)

PR #1901, opened 2024-02-28
docker: Fix startup-sequence locations

christian-heusel:
In https://github.com/znc/znc-docker/pull/28 the location of the init scripts was moved which broke the logic to copy all the init files once the docker submodule was updated in 5d7aa0f7d44cb0d686881ced5b409c3d0569e7ca. Consequently only `30-build-modules.sh` was available in the build docker image. 
We fix this issue by also including all startup-sequence subfolders.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ RUN apk add --no-cache --virtual build-dependencies \
     && cd / && rm -rf /znc-src
 
 COPY docker/slim/entrypoint.sh /
-COPY docker/*/??-*.sh /startup-sequence/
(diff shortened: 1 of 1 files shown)


christian-heusel:
Before:
```
$ ls docker/*/??-*.sh                                                     
docker/full/30-build-modules.sh
```
After:
```
$ ls -1 docker/*/??-*.sh docker/*/startup-sequence/??-*.sh 
docker/full/30-build-modules.sh
docker/slim/startup-sequence/00-try-sh.sh
docker/slim/startup-sequence/01-options.sh
docker/slim/startup-sequence/20-chown.sh
docker/slim/startup-sequence/99-launch.sh
```

PR #1905, opened 2024-04-07
Added SCRAM-SHA1/256/512 to SASL module

patrick-irc:
I have added support for SCRAM-SHA1/256/512 based on OpenSSL to the SASL module.

Networks for testing:
* Libera (SCRAM-SHA-512)
* Contempt https://www.contempt.chat/sasl (SCRAM-SHA1/256/512)
* irc.alphachat.net (SCRAM-SHA-256)

### DIFF
--- a/modules/sasl.cpp
+++ b/modules/sasl.cpp
@@ -554,6 +554,7 @@ Scram::ScramStatus Scram::ProcessServerFirst(CString& sInput) {
     size_t uClientNonceLen;
     CString sPassword = m_pModule->GetNV("password");
     CString sOutput;
+    bool hasMandatoryExtensions = false;
 
     sInput.Split(",", vsParams, false);
 
@@ -570,10 +571,14 @@ Scram::ScramStatus Scram::ProcessServerFirst(CString& sInput) {
             sSalt = sParam.substr(2);
         } else if (!strncmp(sParam.c_str(), "i=", 2)) {
             uiIterCount = strtoul(sParam.substr(2).c_str(), NULL, 10);
+        } else if (!strncmp(sParam.c_str(), "m=", 2)) {
+            hasMandatoryExtensions = true;
+            break;
         }
     }
 
-    if (sServerNonceB64.empty() || sSalt.empty() || uiIterCount == 0) {
+    if (hasMandatoryExtensions || sServerNonceB64.empty() || sSalt.empty() ||
+        uiIterCount == 0) {
         m_pModule->PutModule(
             m_pModule->t_f("Invalid server-first-message: {1}")(sInput));
         return SCRAM_ERROR;
--- a/modules/sasl.cpp
+++ b/modules/sasl.cpp
@@ -554,7 +554,7 @@ Scram::ScramStatus Scram::ProcessServerFirst(CString& sInput) {
     size_t uClientNonceLen;
     CString sPassword = m_pModule->GetNV("password");
     CString sOutput;
-    bool hasMandatoryExtensions = false;
+    bool hasMandatoryExtensions = false, hasError = false;
 
     sInput.Split(",", vsParams, false);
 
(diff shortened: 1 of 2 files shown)


DarthGandalf (commented):
I haven't looked at the algorithm yet, and only glanced at this code.

Can you add a test?

DarthGandalf:
CI failed

modules/sasl.cpp
```diff
@@ -338,10 +463,256 @@ class CSASLMod : public CModule {
...
+    ulVerifierLen = sVerifier.Base64Decode();
+
+    // ServerKey := HMAC(SaltedPassword, "Server Key")
+    pServerKey = static_cast<unsigned char*>(malloc(m_uDigestSize));
+    HMAC(m_pDigest, m_pSaltedPassword, m_uDigestSize,
+         (unsigned char*)SERVER_KEY, strlen(SERVER_KEY), pServerKey,
+         &uiServerKeyLen);
+
+    // ServerSignature := HMAC(ServerKey, AuthMessage)
+    pServerSignature = static_cast<unsigned char*>(malloc(m_uDigestSize));
```
DarthGandalf, line 666: Please don't use memory manually like this, it's not C. Use std::string/std::vector/std::unique_ptr, etc

patrick-irc:
> CI failed

Why are all those checks failing?

DarthGandalf:
I don't know, check logs

modules/sasl.cpp
```diff
@@ -338,10 +463,256 @@ class CSASLMod : public CModule {
...
+
+    if (sServerNonceB64.empty() || sSalt.empty() || uiIterCount == 0) {
+        m_pModule->PutModule(
+            m_pModule->t_f("Invalid server-first-message: {1}")(sInput));
+        return SCRAM_ERROR;
+    }
+    uClientNonceLen = m_sClientNonceB64.length();
+
+    // The server can append his nonce to the client's nonce
+    if (sServerNonceB64.length() < uClientNonceLen ||
```
aaronmdjones, line 584: This does not verify that the server appended anything.

modules/sasl.cpp
```diff
@@ -338,10 +463,256 @@ class CSASLMod : public CModule {
...
+    for (const CString& sParam : vsParams) {
+        if (!strncmp(sParam.c_str(), "r=", 2)) {
+            sServerNonceB64 = sParam.substr(2);
+        } else if (!strncmp(sParam.c_str(), "s=", 2)) {
+            sSalt = sParam.substr(2);
+        } else if (!strncmp(sParam.c_str(), "i=", 2)) {
+            uiIterCount = strtoul(sParam.substr(2).c_str(), NULL, 10);
+        }
+    }
+
```
aaronmdjones, line 575: This does not properly abort on the presence of `m=`.

aaronmdjones: It also does not verify that the server is not repeating parameters, and it does not abort on the presence of a server error (`e=`).

aaronmdjones:
```
RFC 5802 Section 5.1.: SCRAM Attributes

    [...]

    m: This attribute is reserved for future extensibility.  In this
      version of SCRAM, its presence in a client or a server message
      MUST cause authentication failure when the attribute is parsed by
      the other end.

    r: This attribute specifies a sequence of random printable ASCII
      characters excluding ',' (which forms the nonce used as input to
      the hash function).  No quoting is applied to this string.  As
      described earlier, the client supplies an initial value in its
      first message, and the server augments that value with its own
      nonce in its first response.  It is important that this value be
      different for each authentication (see [RFC4086] for more details
      on how to achieve this).  The client MUST verify that the initial
      part of the nonce used in subsequent messages is the same as the
      nonce it initially specified.

```

```
RFC 5802 Section 7: Formal Syntax (With Errata ID 2651)

   [...]

   printable       = %x21-2B / %x2D-7E
                     ;; Printable ASCII except ",".
                     ;; Note that any "printable" is also
                     ;; a valid "value".

   [...]

   nonce           = "r=" c-nonce [s-nonce]
                     ;; Second part provided by server.

   c-nonce         = 1*(printable)

   s-nonce         = 1*(printable)
```

DarthGandalf:
Feel free to reopen if you address the comments

PR #1712, opened 2020-04-04
Allow '--makepass' to take an optional argument

stephenfin:
When automating deployments of ZNC, it can be helpful to capture a generated `<Password>` sections. Allow users to do this by passing an optional argument to the `--makepass` section.

PR #1917, opened 2024-09-22
Allow `--makepass` to work non-interactively

pmodin:
When automating deployments of ZNC, it can be helpful to capture a generated `<Password>` section. Allow users to do this by defining an environment variable `ZNC_PASS` containing the chosen password.

The password block can be retrieved with the following snippet when the environment variable is set;

`znc --makepass | sed -n '/Pass/,/Pass/p'`

I saw that there was a previous attempt in #1712 upon completion, and I shamelessly reused @stephenfin's commit message somewhat. This is another approach to the same issue.

### DIFF
--- a/include/znc/Utils.h
+++ b/include/znc/Utils.h
@@ -56,6 +56,7 @@ class CUtils {
 	 * @returns Piece of znc.conf with <Pass> block
 	 * */
     static CString AskSaltedHashPassForConfig();
+    static CString ConstructSaltedPass(const CString& sPass, const CString& sSalt);
 
     static CString GetSalt();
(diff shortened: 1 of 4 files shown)


DarthGandalf:
Wouldn't it be any better to use this instead?

```
$ echo -n password | argon2 saltysalty -id -e
$argon2id$v=19$m=4096,t=3,p=1$c2FsdHlzYWx0eQ$uYMsezRK7lr5T5Cs3kNT/CeJ1g44FayYpiFEm1BEkxU
```

And you can configure iterations/parallelism/etc more precisely this way

PR #1927, opened 2025-01-28
Fix parsing the 005 numeric.

SadieCat:
From [draft-hard-irc-isupport-01](https://datatracker.ietf.org/doc/html/draft-hardy-irc-isupport-00):

> Due to the increasingly modular nature of IRC server software, it is possible that the status of features previously advertised to clients in RPL_ISUPPORT can change.  When this happens, a server SHOULD reissue the RPL_ISUPPORT numeric with the relevant parameters that have changed.  **If a feature becomes unavailable, the server MUST prefix the parameter with the dash character ('-') when issuing the updated RPL_ISUPPORT.**

I've avoided touching any of the token checks in ParseISupport as they are either:

1. Not reasonable to ever be removed by a server (e.g. NICKLEN).
2. Are checks for `NAMESX` and `UHNAMES` and the booleans for these are never unset when the equivalent capabilities go away either (this should probably be addressed but its out of scope for this PR imo).


src/IRCSock.cpp
```diff
@@ -1456,7 +1456,14 @@ void CIRCSock::ParseISupport(const CMessage& Message) {
             break;
         }
 
-        m_mISupport[sName] = sValue;
+        bool bAdding = '-' != sName[0];
```
DarthGandalf, line 1459: I think this introduces a vulnerability: try sending `:server 005 user =FOO :are supported`

PR #1928, opened 2025-01-28
Fix sending server passwords with spaces in them.

### LINKED ISSUE #1899: Missing : in PASS command
ZNC sends PASS without a leading colon. This means that for ZNC users to send their passwords with spaces in them, they need to prefix their password with a leading colon. That aspect of IRC shouldn't leak into the password configuration field.

https://github.com/znc/znc/blob/da88ec3df8c17aa6358e2b7f4489635cc2232c7c/src/IRCSock.cpp#L1256

SadieCat:
Passwords should be sent as a `<trailing>` parameter or they may be misinterpreted, e.g. `PASS foo bar` could be interpreted as a password of "foo" on some servers.

Fixes #1899.

### DIFF
--- a/src/IRCSock.cpp
+++ b/src/IRCSock.cpp
@@ -1323,7 +1323,7 @@ void CIRCSock::Connected() {
     PutIRC("CAP LS 302");
 
     if (!sPass.empty()) {
-        PutIRC("PASS " + sPass);
+        PutIRC("PASS :" + sPass);
     }
 
     PutIRC("NICK " + sNick);


src/IRCSock.cpp
```diff
@@ -1323,7 +1323,7 @@ void CIRCSock::Connected() {
     PutIRC("CAP LS 302");
 
     if (!sPass.empty()) {
-        PutIRC("PASS " + sPass);
+        PutIRC("PASS :" + sPass);
```
DarthGandalf, line 1326: This will break broken servers which don't expect PASS to start with `:` and just take the second word as is

Need to add it or not depending on what characters the password has; and easy way to do that is to use CMessage

SadieCat: I think the CString overload of PutIRC will already use CMessage?

DarthGandalf: When CMessage parses a string, it remember whether the trailing arg had an extra colon, and if it had, it puts the extra colon there when stringifying.

DarthGandalf: But when you do SetParam(), that flag is not set

PR #1933, opened 2025-02-21
CMake: use cmake_{push,pop}_check_state()

dacav:
Useful info:
https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ include(use_homebrew)
 include(GNUInstallDirs)
 include(CheckCXXSymbolExists)
 include(copy_csocket)
+include(CMakePushCheckState)
 
 set(CMAKE_THREAD_PREFER_PTHREAD true)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -89,13 +90,12 @@ if(WANT_OPENSSL)
 		# LibreSSL gained that function later too.
 		# TODO: maybe remove this check at some point, and stop supporting old
 		# libssl versions
-		function(check_SSL_SESSION_get0_cipher)
+		cmake_push_check_state(RESET)
 			set(CMAKE_REQUIRED_LIBRARIES ${OPENSSL_LIBRARIES})
```
DarthGandalf, line 94: since this is not a function anymore, please fix indent.

Alternatively, maybe it makes sense to indent the one below instead

DarthGandalf (approved):

PR #1873, opened 2023-08-31
Feature sasl plain

delthas:
This builds upon https://github.com/znc/znc/pull/1573:
- rebased
- comments addressed
- test added
- tested with senpai

Feel free to squash when merging

PR #1936, opened 2025-02-22
SASL Authentication for Clients

### LINKED ISSUE #296: SASL Login to ZNC
It would be nice to see users able to log into their ZNC account via SASL.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/101799-sasl-login-to-znc?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1759&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DarthGandalf:
Based on #1873 which is based on #1573, with heavy modifications

CC @delthas @MrLenin

This is draft PR for now, I need to fix modpython+modperl before merging, and add many tests. But I think the API is ready

Close #296


modules/saslplainauth.cpp
```diff
@@ -0,0 +1,58 @@
...
+    EModRet OnClientSASLAuthenticate(const CString& sMechanism,
+                                     const CString& sMessage) override {
+        if (!sMechanism.Equals("PLAIN")) {
+            return CONTINUE;
+        }
+
+        CString sNullSeparator = std::string("\0", 1);
+        CString sAuthzId = sMessage.Token(0, false, sNullSeparator, true);
+        CString sAuthcId = sMessage.Token(1, false, sNullSeparator, true);
+        CString sPassword = sMessage.Token(2, false, sNullSeparator, true);
```
psychon, line 37: Having no clue about sasl.... what is up with the 3rd or 4th field? Should the last call use `true` instead of `false` so that all of them are just part of the password?

DarthGandalf: From [RFC 4616](https://datatracker.ietf.org/doc/html/rfc4616)

```
   message   = [authzid] UTF8NUL authcid UTF8NUL passwd
   passwd    = 1*SAFE ; MUST accept up to 255 octets
   UTF8NUL   = %x00 ; UTF-8 encoded NUL character

   SAFE      = UTF1 / UTF2 / UTF3 / UTF4
               ;; any UTF-8 encoded Unicode character except NUL
```

Because 4th field doesn't exist, and password cannot contain NUL itself, we can either ignore the extra fields, or reject auth if they are present

DarthGandalf: That said, we probably should replace the extensive usage of Token() throughout the codebase with some kind of Split(), to avoid going through first fields multiple times. And replace many usages of CString with std::string_view. But that's outside of scope of this PR

DarthGandalf:
A caveat: this doesn't currently support specifying the network name and client id. I don't know whether it should

dgw:
> I don't know whether it should

To me, it seems important for a pre-welcome auth method to allow passing in all the important connection details, including network name and client ID.

I see limited utility in connecting without _at least_ specifying the network name up front so ZNC can get to sending joins, buffer, etc.

Client ID is nice to have, but I'd guess it will come along for little to no additional cost if net-name support is added.

DarthGandalf:
> > I don't know whether it should
> 
> To me, it seems important for a pre-welcome auth method to allow passing in all the important connection details, including network name and client ID.
> 
> I see limited utility in connecting without _at least_ specifying the network name up front so ZNC can get to sending joins, buffer, etc.
> 
> Client ID is nice to have, but I'd guess it will come along for little to no additional cost if net-name support is added.

Sure.

My main concern there is that it would be bending the SASL protocol, assigning additional meaning to its fields. This use somewhat reminds me of authzid field (though not exact match), but what clients allow setting authcid and authzid separately?

Re network name specifically, perhaps [bouncer-networks](https://github.com/ircv3/ircv3-specifications/pull/478) cap would solve that. But not client id

delthas:
For specifying the network, soju supports passing a `nick/network` in the SASL username field. Just like for USER (similarly to what ZNC does). Perhaps ZNC could do the same?

dgw:
> My main concern there is that it would be bending the SASL protocol, assigning additional meaning to its fields.

As I see it, ZNC already bends the IRC protocol a bit by parsing the network name & client ID out of the USER and/or PASS—or even just allowing `PASS username:p@ssw0rd` to override the USER field. What makes SASL the red line not to be crossed?

I promise I'm not trying to be difficult 😅, just to understand the thought process. To me, SASL on IRC is just a way to send the same username + password you would via PASS, just pre-registration. If the semantics for SASL-to-ZNC differed from PASS-to-ZNC, I think _that_ would be weirder.

DarthGandalf:
> To me, SASL on IRC is just a way to send the same username + password you would via PASS

If you limit it to PLAIN mechanism, yes. But SASL supports many other mechanisms, including mechanisms which are not conceived yet.

> I promise I'm not trying to be difficult 😅, just to understand the thought process.

Therefore, as devil's advocate, I try to imagine ways how such future SASL mechanisms will break the model here.

E.g. in EXTERNAL mechanism, there is no password to send, and there's no authcid either, only authzid. I plan to add EXTERNAL to the certauth module; and with auth by certificate, again there are questions how to specify network and client id.

Don't worry, difficult is good.

modules/saslplainauth.cpp
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        GetClient()->StartSASLPasswordCheck(sAuthcId, sPassword);
+        return HALTMODS;
+    }
+};
+
+template <>
+void TModInfo<CSASLMechanismPlain>(CModInfo& Info) {
+    Info.SetWikiPage("saslplain");
```
dgw, line 52: Forgotten in af2175390160b1558d844036b270d419b86471e9?

```suggestion
    Info.SetWikiPage("saslplainauth");
```

dgw:
> > To me, SASL on IRC is just a way to send the same username + password you would via PASS
> 
> If you limit it to PLAIN mechanism, yes. But SASL supports many other mechanisms, including mechanisms which are not conceived yet.

That's on me for not being specific enough. This PR only seems to implement PLAIN, so everything I've said about SASL here has been about this mechanism.

(Looking at SCRAM a bit, I think it would be possible to bring the `user@client/network` semantics into that, too, if wanted.)

> I plan to add EXTERNAL to the certauth module; and with auth by certificate, again there are questions how to specify network and client id.

Yes, that mech in particular makes it tricky to specify a network.

If users are expected to generate self-signed certs or otherwise fully control their own EXTERNAL credentials (i.e. not pay-per-cert to a third party issuer), I could see using the CN field to indicate `user@client/network`, and every "profile" would have its own fingerprint. (If there are other EXTERNAL credential types to think about besides TLS client certificates, that's a layer I haven't touched at all.)

DarthGandalf:
All the modern SASL mechanism, including EXTERNAL, have a way to specify authz id, so I've parsed network name from there.

Haven't tested with actual IRC client yet, but works in unit test.

DarthGandalf:
@delthas yes, see the test added for more options

PR #1939, opened 2025-03-16
Find boost in CONFIG mode of CMake

### LINKED ISSUE #1934: CMake upgrade?
The minimum supported cmake version is 3.13, which is quite old by now.
Would you consider bumping it to something more recent?

For a comparison, this page aggregates information about what versions are shipped by distributions: https://pkgs.org/download/cmake

DarthGandalf:
To silence CMake's warning

Fix #1934

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,7 +261,7 @@ endif()
 
 tristate_option(I18N "Native language support (i18n)")
 if(WANT_I18N)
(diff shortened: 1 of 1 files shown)


DarthGandalf:
@dacav boost on cygwin is too old

dacav:
Ah, pity! :)
Oh well, thanks for trying.

DarthGandalf:
OTOH, at this point it's more preferable to disable the i18n support on the broken platform than to hold back other platforms.

Cygwin maintainers can always update the version of boost packaged

PR #1942, opened 2025-04-12
Webadmin: Fix Edit Network when using `DenySetNetwork`

### LINKED ISSUE #1931: [Feature request] Allow editing other settings except the servers and network name
When #1814 was added (per my request) I was expecting it to only disallow users from changing the network name and removing/adding new servers.

Recently I started a free ZNC service and unfortunately, users aren't able also to change the charset, etc.

I'd like to know if it'd be possible to change the behaviour to only disallow users from changing network name and remove/add servers.

Ping to @bashgeek, since it's the person that made the #1814 PR.

Thanks!

bashgeek:
Fixes #1931 

When using `DenySetNetwork` it is currently not possible to edit/save a network on webadmin, it'll fail with `Network name is a required argument`, because the input field for the name is disabled in the form and won't get transmitted in submit. I added a hidden field for this if that's the case.

Also fixed the population of `Ident`, `QuitMsg` and `RealName` on the network form - to have the current values in the disabled inputs.


modules/data/webadmin/tmpl/add_edit_network.tmpl
```diff
@@ -31,8 +31,9 @@
 			<div class="sectionbody">
 				<div class="subsection">
 					<div class="inputlabel"><label for="name"><? FORMAT "Network Name:" ?></label></div>
-					<input id="name" type="text" name="name" value="<? VAR Name ?>" class="half" maxlength="128"
+					<input id="name" type="text" name="name<? IF !NameEdit ?>_disabled<? ENDIF ?>" value="<? VAR Name ?>" class="half" maxlength="128"
 						   title="<? FORMAT "The name of the IRC network." ?>" <? IF !NameEdit ?>disabled<? ENDIF ?> />
+					<? IF !NameEdit ?><input type="hidden" name="name" value="<? VAR Name ?>"><? ENDIF ?>
```
DarthGandalf, line 36: What's the point of adding the hidden field AND renaming the old field?

I think the change you did in .cpp should be enough

bashgeek: The browser will not send disabled fields on submit, so it will be missing in the POST request and with error message `Network name is a required argument`, that's why a hidden field was added to transfer this field. The disabled input field was renamed because of HTML compliance (input name's in the same form ought to be unique).

PR #1945, opened 2025-04-21
Unix sockets

### LINKED ISSUE #1240: support for listening on a unix socket, not a port

DarthGandalf:
Based on @psychon's old #1363 
Fix #1240


frostyfalls:
Is there any way to set the access mode of the socket? I couldn't find any documentation or clues in the diff.

I'm trying to proxy the HTTP traffic, and it only connects if it's writable (ideally 0666), but the socket is created as 0755.

PR #1961, opened 2025-06-21
Make away-notify and chghost only send to client if attached into a c…

RealKindOne:
ZNC sends ```away-notify``` and ```chghost``` events to the client even if you are detached from all channels with the user. 

This pull request does break the ```ZNCTest.AwayNotify``` test. I have not figure out how to exactly fix that?


src/Client.cpp
```diff
@@ -591,6 +591,21 @@ bool CClient::PutClient(const CMessage& Message) {
     switch (Message.GetType()) {
         case CMessage::Type::Away:
             if (!m_bAwayNotify) return false;
+            // Do not send away-message if all shared channels are detached.
```
DarthGandalf, line 594: This is a good idea, but it doesn't belong here: the channel is attached or detached for all clients together. none of this code is specific to the client. So if several clients are connected, you're doing the same check several times.

The check should happen in IRCSock.cpp instead

src/Client.cpp
```diff
@@ -605,6 +620,24 @@ bool CClient::PutClient(const CMessage& Message) {
...
+
+                const vector<CChan*>& vChans = m_pNetwork->GetChans();
+                for (const CChan* pChan : vChans) {
+                    if (!pChan->IsDetached() && pChan->FindNick(sChgHost)) {
+                        bHasSharedChannel = true;
+                        break;
+                    }
+                }
+
+                if (!bHasSharedChannel) return false;
```
DarthGandalf, line 638: This should be a function, called from both places. Let's not duplicate code

src/Client.cpp
```diff
@@ -1669,6 +1702,14 @@ bool CClient::OnTagMessage(CTargetMessage& Message) {
 bool CClient::OnTextMessage(CTextMessage& Message) {
     CString sTargets = Message.GetTarget();
 
+    // Check for empty PRIVMSG
```
DarthGandalf, line 1705: This belongs to a separate commit, this is not related to away-notify nor chghost

test/integration/tests/core.cpp
```diff
@@ -301,6 +301,81 @@ TEST_F(ZNCTest, AwayNotify) {
     client.Write("znc shutdown");
 }
 
+
+TEST_F(ZNCTest, AwayNotifyDetached) {
+    auto znc = Run();
+    auto ircd = ConnectIRCd();
+    auto client = ConnectClient();
```
DarthGandalf, line 308: LoginClient() would simplify some of this code. It doesn't send CAP LS, but that's fine, REQ is enough

The test above already tests that away-notify is not advertised by ZNC too early

test/integration/tests/core.cpp
```diff
@@ -301,6 +301,81 @@ TEST_F(ZNCTest, AwayNotify) {
...
+    client.ReadUntil(":test!test@test JOIN #test");
+    ircd.Write("353 nick = #test :nick!ident@host test!test@test");
+    ircd.Write("366 nick #test :End of /NAMES list.");
+    client.ReadUntil("#test :End of /NAMES");
+    client.Write("znc detach #test");
+    client.ReadUntil(":*status!status@znc.in PRIVMSG nick :Detached 1 channel");
+    ircd.Write(":test!test@test AWAY :going away");
+
+    QString result = QString::fromUtf8(client.ReadRemainder());
+    EXPECT_FALSE(result.contains(":test!test@test AWAY :going away"))
```
DarthGandalf, line 346: EXPECT_THAT(client.ReadReminder().toStdString(), Not(HasSubstr())

DarthGandalf (changes requested):

DarthGandalf:
Thinking about it further... Could server genuinely send AWAY for people who you've recently talked in private even if you share no channels with them? This ZNC-side filtering would break that

RealKindOne:
> Thinking about it further... Could server genuinely send AWAY for people who you've recently talked in private even if you share no channels with them? This ZNC-side filtering would break that

https://ircv3.net/specs/extensions/away-notify.html

>>When this capability is enabled, clients will be sent an AWAY message when a user sharing a channel with them sets, changes or removes their away state

Issue #1964, opened 2025-07-07
extra rpath missing from libraries


th0ma7:
I'm the packager for the SynoCommunity ZNC package
Current PR: https://github.com/SynoCommunity/spksrc/pull/6636

I've been hunting down an issue when, looking in more details at rpath (as I'm using multiple rpath when cross-comliling) - the resulting `bin/znc` binary has my two expected rpaths (using libtree tool to track from my build system which shows libraries as not found as expected, but most importantly shows the actual rpath detail):
```
libtree work-x64-7.1/install/var/packages/znc/target/bin/znc
work-x64-7.1/install/var/packages/znc/target/bin/znc 
├── libz.so.1 [ld.so.conf]
├── libargon2.so.1 [ld.so.conf]
├── libcrypto.so.3 [ld.so.conf]
├── libssl.so.3 [ld.so.conf]
│   └── libcrypto.so.3 [ld.so.conf]
├── libpthread.so.0 [ld.so.conf]
├── librt.so.1 [ld.so.conf]
├── libicuuc.so.77 not found
└── libboost_locale.so.1.86.0 not found
    ┊ Paths considered in this order:
    ┊ 1. rpath:
    ┊    depth 1
    ┊    /var/packages/znc/target/lib
    ┊    /var/packages/python312/target/lib
    ┊ 2. LD_LIBRARY_PATH was not set
    ┊ 3. runpath was not set
    ┊ 4. ld config files:
    ┊    /usr/lib/x86_64-linux-gnu/libfakeroot
    ┊    /usr/local/lib
    ┊    /usr/local/lib/x86_64-linux-gnu
    ┊    /lib/x86_64-linux-gnu
    ┊    /usr/lib/x86_64-linux-gnu
    ┊    /lib32
    ┊    /usr/lib32
    ┊    /libx32
    ┊    /usr/libx32
    ┊ 5. Standard paths:
    ┊    /lib
    ┊    /lib64
    ┊    /usr/lib
    ┊    /usr/lib64
```
Although all the libraries are missing that second rpath:
```
libtree work-x64-7.1/install/var/packages/znc/target/lib/znc/log.so 
work-x64-7.1/install/var/packages/znc/target/lib/znc/log.so 
├── libssl.so.3 [ld.so.conf]
│   └── libcrypto.so.3 [ld.so.conf]
├── libcrypto.so.3 [ld.so.conf]
└── libicuuc.so.77 not found
    ┊ Paths considered in this order:
[...]

PR #1967, opened 2025-07-12
Fix issue with LDFLAGS defined rpath missing in modules

th0ma7:
Fix issue with `LDFLAGS` defined rpath being missing in modules.

Fixes: https://github.com/znc/znc/issues/1964

### DIFF
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -17,6 +17,8 @@
 # This is not recommended, but whatever.
 file(GLOB all_modules LIST_DIRECTORIES FALSE "*")
 
+set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} ${CMAKE_EXE_LINKER_FLAGS}")
+
 function(add_cxx_module mod modpath)
 	znc_add_library("module_${mod}" MODULE "${modpath}")
 	set_target_properties("module_${mod}" PROPERTIES


DarthGandalf:
I don't know what's wrong with LDFLAGS, but can't you pass `CMAKE_MODULE_LINKER_FLAGS` to cmake without modifying CMakeLists.txt?

th0ma7:
@DarthGandalf sorry to bother but I'm not entirely sure I follow... I guess you mean passing a `-DCMAKE_MODULE_LINKER_FLAGS=$(LDFLAGS)` or similar at configure time? ... actually that could be `-DCMAKE_MODULE_LINKER_FLAGS=$(CMAKE_EXE_LINKER_FLAGS)`?  I could easily add something in that train of thought without needing to modify the CMakeLists.txt.

Although that would would only be a fix on my end, not that different than maintaining my patch which is totally fine in the end.  Although, the underneath issue wheras LDFLAGS are not being processed properly for modules while they are totally fine with the znc binary would still be.

th0ma7:
> I suggest you open an issue about it to cmake. Would need a smaller reproducing example though than whole znc

I was thinking the exact same thing, and also test other cmake versions to see how reproducible that is.  Next on my todo list...

th0ma7:
Problem now solved: we've further standardized our environment to uniquely rely on our toolchain file and no longer have `LDFLAGS` set in the environment.  Issue is now gone, closing.

PR #1976, opened 2025-10-02
Add a dark mode to the default webskin

dannycolin:
The light or dark theme is automatically selected based on the user `prefers-color-scheme` preference from the OS or browser. To do so, it uses the following CSS features:

* [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) (available across browsers since ⁨January 2022).
* [light-dark()](available across browsers since ⁨May 2024)


DarthGandalf:
Is it possible to fallback to a color if light-dark() isn't available?

DarthGandalf:
SASL module will need some special treatment

Выделение_744

dannycolin:
We can use the older syntax which consist of two separated blocks. It would looks like:

```css
:root {
  // variables for the light theme

  @media (prefers-color-scheme: dark) {
   // variables for the dark theme
  }
}
```

It's a bit more verbose but it'll fallback to the light colors if a browser doesn't support `prefers-color-scheme`.

dannycolin:
> SASL module will need some special treatment

Ah! Good catch. It looks like jQuery UI components are using their builtin styling even in the current codebase. I'll fix it and update my patch ;).

webskins/_default_/pub/selectize-0.12.1-dark.css
```diff
@@ -0,0 +1,32 @@
+:root {
```
DarthGandalf, line 1: Please add a comment that it's ZNC's custom code, and not upstream.

Renaming this file to e.g. selectize-znc-dark.css would also help

webskins/_default_/pub/_default_.css
```diff
@@ -109,7 +109,8 @@ a:hover {
 button,
 input[type="button"],
 input[type="submit"],
-.sectionbody input[type="button"] {
+.sectionbody input[type="button"],
```
DarthGandalf, line 112: shouldn't the line above cover this?

DarthGandalf:
Thanks!

> I also added styling for the jquery UI components like the one you point to in your screenshot.

The autocomplete menu is still white. Try pressing backspace to remove PLAIN from the to see it in the white autocomplete list

webskins/_default_/pub/selectize-0.12.1-dark.css
dannycolin: Done.

dannycolin:
> Thanks!
> 
> > I also added styling for the jquery UI components like the one you point to in your screenshot.
> 
> The autocomplete menu is still white. Try pressing backspace to remove PLAIN from the field to see it in the white autocomplete list

It should be fixed now. 

*Note:* 
If you remove "external" and "plain" from the input, the dropdown styling is slightly broken. That problem existed prior to this patch. Do you prefer to fix it now or in a follow-up patch?

DarthGandalf:
Follow up is fine

PR #1993, opened 2026-03-10
Enable SSL by default in ZNC configuration

TehPeGaSuS:
It's 2026. With Let's Encrypt, IMHO, there's literally no reason to not use SSL/TLS for server connections.


DarthGandalf:
1. This is not enough to do what you want
2. I see CI is failing

TehPeGaSuS:
I'll revisit this on a later date

PR #2015, opened 2026-04-25
ZNCString: guard Replace/Split against empty-width arguments

### LINKED ISSUE #2009: ZNCString: CString::Replace and CString::Split loop forever on empty separator
Two sibling library-API bugs in `CString`. Same shape as #1994: a missing input guard turns a degenerate argument into UB / unbounded loop. No in-tree caller passes the bad argument today, but the public API is exposed to module authors and to anyone embedding libznc.

## 1) `CString::Replace`: empty `sReplace` causes pointer underflow + OOM

```cpp
// src/ZNCString.cpp
626: unsigned int CString::Replace(CString& sStr, const CString& sReplace,
627:                               const CString& sWith, const CString& sLeft,
628:                               const CString& sRight, bool bRemoveDelims) {
...
656:         } else if (!bInside &&
657:                    strncmp(p, sReplace.c_str(), uReplaceWidth) == 0) {
658:             sStr += sWith;
659:             p += uReplaceWidth - 1;        // <-- when uReplaceWidth==0: p -= 1
660:             uRet++;
...
665:         p++;                                // <-- back to where we started
```

When `sReplace.empty()`, `strncmp(_, _, 0) == 0` always matches, the arithmetic at 659 underflows `size_t` to `SIZE_MAX`, the `p++` at 665 returns to the same byte, and the loop appends `sWith` indefinitely until ASan kills the process for OOM.

PoC (3-second guard prevents the OOM kill):
```cpp
#include <cstdio>
#include <signal.h>
#include <unistd.h>
#include <znc/ZNCString.h>
extern "C" void to(int) { _exit(124); }
int main() {
    signal(SIGALRM, to); alarm(3);
    CString s = "abc";
    CString::Replace(s, /*sReplace=*/"", /*sWith=*/"X");
}
```
```
ZNCString.cpp:659:15: runtime error: addition of unsigned offset to
  0x... overflowed to 0x...
[3s later] killed by SIGALRM
```

## 2) `CString::Split`: empty `sDelim` + `bAllowEmpty=false` infinite-loops

Same pattern, two locations:

```cpp
// src/ZNCString.cpp
854:     if (!bAllowEmpty) {
855:         while (strncasecmp(p, sDelim.c_str(), uDelimLen) == 0) {
856:             p += uDelimLen;                // <-- 0 when sDelim.empty()
857:         }
858:     }
...
893:             if (!bAllowEmpty) {
894:                 while (strncasecmp(p, sDelim.c_str(), uDelimLen) == 0) {
895:                     p += uDelimLen;        // <-- same bug
896:                 }
897:             }
```

`strncasecmp(_, _, 0)` is unconditionally 0; `p` never advances; the loop spins on the CPU forever. No allocation, so this hangs rather than OOMs.

PoC:
```cpp
VCString out;
CString s = "abc";
s.Split(/*sDelim=*/"", out, /*bAllowEmpty=*/false);   // never returns
```

## Reachability

I grepped every caller in `src/` and `modules/` and didn't find one that passes an empty `sReplace` or empty `sDelim`. So this is the same "library-API correctness, not currently reachable" framing as #1994. I'm just bringing the contracts of these two functions in line with what callers expect.

## Fix

```cpp
// in CString::Replace
if (sReplace.empty()) return 0;

// in CString::Split, guard the prefix-skip loop
if (!bAllowEmpty && !sDelim.empty()) {
    while (strncasecmp(p, sDelim.c_str(), uDelimLen) == 0) p += uDelimLen;
}
// and likewise at line 894.
```

Or, more aggressively, bail out at the top of `Split` if `sDelim.empty()`.

Happy to PR with a couple of regression tests in `test/StringTest.cpp` mirroring #1995.

MarkLee131:
Close #2009:
CString::Replace with an empty sReplace underflowed 'p += uReplaceWidth - 1' to SIZE_MAX and then the per-iteration 'p++' brought p back to the same byte, so the function looped forever appending sWith to the output and eventually OOMed.

CString::Split with empty sDelim and bAllowEmpty=false spun in the prefix-skip loops because strncasecmp(p, "", 0) is unconditionally 0 and p += 0 never advances.

No in-tree caller currently passes the bad argument, but the public library API should not be a bear trap for module authors. Same shape as #1994.

### DIFF
--- a/test/StringTest.cpp
+++ b/test/StringTest.cpp
@@ -190,11 +190,11 @@ TEST(StringTest, Split) {
     // element (or zero elements if the input itself is empty).
     VCString vempty;
     EXPECT_EQ(CS("abc").Split("", vempty, false), 1u);
-    EXPECT_THAT(vempty, ElementsAre("abc"));
+    EXPECT_EQ(vempty, VCString({"abc"}));
     EXPECT_EQ(CS("abc").Split("", vempty, true), 1u);
     EXPECT_EQ(vempty, VCString({"abc"}));
     EXPECT_EQ(CS("").Split("", vempty, false), 0u);
-    EXPECT_THAT(vempty, IsEmpty());
+    EXPECT_TRUE(vempty.empty());
 }
 
 TEST(StringTest, NamedFormat) {
--- a/test/StringTest.cpp
+++ b/test/StringTest.cpp
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <znc/ZNCString.h>
+
+using ::testing::ElementsAre;
+using ::testing::IsEmpty;
 
 class EscapeTest : public ::testing::Test {
   protected:
@@ -190,11 +194,11 @@ TEST(StringTest, Split) {
     // element (or zero elements if the input itself is empty).
     VCString vempty;
     EXPECT_EQ(CS("abc").Split("", vempty, false), 1u);
-    EXPECT_EQ(vempty, VCString({"abc"}));
+    EXPECT_THAT(vempty, ElementsAre("abc"));
     EXPECT_EQ(CS("abc").Split("", vempty, true), 1u);
-    EXPECT_EQ(vempty, VCString({"abc"}));
+    EXPECT_THAT(vempty, ElementsAre("abc"));
     EXPECT_EQ(CS("").Split("", vempty, false), 0u);
-    EXPECT_TRUE(vempty.empty());
+    EXPECT_THAT(vempty, IsEmpty());
 }
 
 TEST(StringTest, NamedFormat) {
--- a/test/StringTest.cpp
+++ b/test/StringTest.cpp
@@ -190,7 +190,7 @@ TEST(StringTest, Split) {
     // element (or zero elements if the input itself is empty).
     VCString vempty;
     EXPECT_EQ(CS("abc").Split("", vempty, false), 1u);
-    EXPECT_EQ(vempty, VCString({"abc"}));
+    EXPECT_THAT(vempty, ElementsAre("abc"));
     EXPECT_EQ(CS("abc").Split("", vempty, true), 1u);
     EXPECT_EQ(vempty, VCString({"abc"}));
     EXPECT_EQ(CS("").Split("", vempty, false), 0u);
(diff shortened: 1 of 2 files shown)


DarthGandalf:
All these prs need some unit tests

test/StringTest.cpp
```diff
@@ -172,6 +184,17 @@ TEST(StringTest, Split) {
...
+    // Empty delimiter must not spin in the prefix-skip loop (#2009).
+    // With nothing to split on, the whole input is returned as a single
+    // element (or zero elements if the input itself is empty).
+    VCString vempty;
+    EXPECT_EQ(CS("abc").Split("", vempty, false), 1u);
+    EXPECT_EQ(vempty, VCString({"abc"}));
+    EXPECT_EQ(CS("abc").Split("", vempty, true), 1u);
+    EXPECT_EQ(vempty, VCString({"abc"}));
+    EXPECT_EQ(CS("").Split("", vempty, false), 0u);
+    EXPECT_TRUE(vempty.empty());
```
DarthGandalf:
```suggestion
    EXPECT_THAT(vempty, IsEmpty());
```

test/StringTest.cpp
```diff
@@ -172,6 +184,17 @@ TEST(StringTest, Split) {
 
     CS("a=x&c=d&a=b").URLSplit(mresult);
     EXPECT_EQ(mexpected, mresult) << "URLSplit";
+
+    // Empty delimiter must not spin in the prefix-skip loop (#2009).
+    // With nothing to split on, the whole input is returned as a single
+    // element (or zero elements if the input itself is empty).
+    VCString vempty;
+    EXPECT_EQ(CS("abc").Split("", vempty, false), 1u);
+    EXPECT_EQ(vempty, VCString({"abc"}));
```
DarthGandalf:
```suggestion
    EXPECT_THAT(vempty, ElementsAre("abc"));
```

DarthGandalf (approved):

DarthGandalf:
have you noticed failing CI?

MarkLee131:
Sorry, missed the CI mail. Just pushed a fix: those EXPECT_THAT lines came from your own suggestion, but StringTest.cpp doesn't include gmock, so the file wouldn't compile on any cpp job. Swapped them for EXPECT_EQ against a VCString and EXPECT_TRUE(vempty.empty()) to keep the file gtest-only. Want me to pull gmock in instead and put the matchers back?

DarthGandalf:
Um, sure. I thought it's obvious enough that using gmock requires including gmock :)

DarthGandalf:
The problem with EXPECT_TRUE is that the error message it gives only says "not true", while EXPECT_THAT outputs the actual list of values

MarkLee131:
yeah fair, I overcorrected. should've just pulled in gmock instead of swapping the matchers. point about the failure diagnostics is right too. put the EXPECT_THAT lines back with #include <gmock/gmock.h> and the using ::testing::* in 80699c41, same shape as MessageTest.cpp.

PR #2017, opened 2026-04-25
User: use constant-time compare for MD5/SHA256/plain password paths

### LINKED ISSUE #2011: User: legacy MD5/SHA256 CheckPass uses non-constant-time compare
For `HASH_MD5` and `HASH_SHA256`, `CheckPass` ends up doing `m_sPass.Equals(salted_hash)` which routes through `strcmp` (`CString::StrCmp` at `src/ZNCString.cpp:148-150`). That's not a constant-time compare.

```cpp
// src/User.cpp
1057: bool CUser::CheckPass(const CString& sPass) {
...
1064:     switch (m_eHashType) {
1065:         case HASH_MD5:
1066:             bResult = m_sPass.Equals(CUtils::SaltedMD5Hash(sPass, m_sPassSalt));
...
1069:         case HASH_SHA256:
1070:             bResult = m_sPass.Equals(CUtils::SaltedSHA256Hash(sPass, m_sPassSalt));
...
1077:         case HASH_ARGON2ID:
1078:             return argon2id_verify(...) == ARGON2_OK;     // constant-time
...
1085:         case HASH_NONE:
1086:             return (sPass == m_sPass);                    // also not CT
```

The Argon2id branch is fine; only the legacy hashes are affected.

The leak via remote timing is the stored hash digest, not the password itself. Recovering 64 hex chars of a SHA256 hash via remote timing requires a lot of measurements and ZNC has fail2ban available for rate-limiting, so this is more "code quality / future-proofing" than an exploitable vulnerability. Filing because every other piece of crypto-adjacent code in the project uses constant-time comparisons.

## Fix

```cpp
case HASH_SHA256: {
    CString expected = CUtils::SaltedSHA256Hash(sPass, m_sPassSalt);
    bResult = expected.length() == m_sPass.length() &&
              CRYPTO_memcmp(expected.data(), m_sPass.data(),
                            m_sPass.length()) == 0;
#if ZNC_HAVE_ARGON
    bUpgrade = true;
#endif
    break;
}
```

Same idea for `HASH_MD5`. `CRYPTO_memcmp` is in OpenSSL and ZNC already links against OpenSSL when `WANT_OPENSSL=ON`. For builds without OpenSSL, a two-line fallback that XORs all bytes and ORs the result works.

Could PR if you'd like.

MarkLee131:
Close #2011:
CString::Equals falls through to strcmp, which short-circuits on the first differing byte. That leaks the length of the common prefix between the stored hash (or plain password) and the attacker's guess via response timing. Argon2id already uses argon2id_verify which is constant-time.

Add a small ConstantTimeEquals helper and use it for the legacy MD5, SHA256 and plain HASH_NONE branches. No new dependency: the helper is ~10 lines and works on builds without OpenSSL.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -271,6 +271,23 @@ CString CUtils::SaltedHash(const CString& sPass, const CString& sSalt) {
 #endif
 }
 
+bool CUtils::ConstantTimeEquals(const CString& a, const CString& b) {
+    // Length is leaked, but for the cases this is used in (fixed-size
+    // hex hashes for MD5 / SHA256) the lengths are constant. Plain-text
+    // mode does leak length, but plain-text passwords are deprecated and
+    // discouraged in znc.conf.
+    if (a.length() != b.length()) {
+        return false;
+    }
+    unsigned char acc = 0;
+    const unsigned char* pa = reinterpret_cast<const unsigned char*>(a.data());
+    const unsigned char* pb = reinterpret_cast<const unsigned char*>(b.data());
+    for (size_t i = 0; i < a.length(); ++i) {
+        acc |= static_cast<unsigned char>(pa[i] ^ pb[i]);
+    }
+    return acc == 0;
+}
+
 CString CUtils::GetPass(const CString& sPrompt) {
 #ifdef HAVE_TCSETATTR
     // Disable echo
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -28,6 +28,7 @@
 #ifdef HAVE_LIBSSL
 #include <openssl/ssl.h>
 #include <openssl/bn.h>
+#include <openssl/crypto.h>
 #include <openssl/rsa.h>
 #if (OPENSSL_VERSION_NUMBER < 0x10100000L) || (defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x20700000L))
 #define X509_getm_notBefore X509_get_notBefore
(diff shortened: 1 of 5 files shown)


src/Utils.cpp
```diff
@@ -271,6 +271,23 @@ CString CUtils::SaltedHash(const CString& sPass, const CString& sSalt) {
...
+    // mode does leak length, but plain-text passwords are deprecated and
+    // discouraged in znc.conf.
+    if (a.length() != b.length()) {
+        return false;
+    }
+    unsigned char acc = 0;
+    const unsigned char* pa = reinterpret_cast<const unsigned char*>(a.data());
+    const unsigned char* pb = reinterpret_cast<const unsigned char*>(b.data());
+    for (size_t i = 0; i < a.length(); ++i) {
+        acc |= static_cast<unsigned char>(pa[i] ^ pb[i]);
```
DarthGandalf, line 295: could compiler recognize this and "optimize" it back to strcmp()?

psychon: Random drive-by comment: Yes, they can.

At least I do not know about anything that would forbid compilers from doing so. (Even if they do not do it currently)

https://stackoverflow.com/a/44700409/436275 (random Google result) has a similar opinion:

> TL;DR: Use assembly.
> [...]
> the LLVM optimizer will wreak havoc with your algorithm, short-circuiting it, eliminating unnecessary reads, now or in the future,

MarkLee131: Thanks for pointing it out. 
 Was thinking of changing it to use CRYPTO_memcmp (<openssl/crypto.h>) under HAVE_LIBSSL, since OpenSSL is already pulled in for the SHA256 path. For builds without OpenSSL, fall back to the same byte loop but with the accumulator and pointers marked volatile, plus a comment noting it's best-effort on that path so MD5 and plain still get covered.

Does that seem reasonable before I push a change?

DarthGandalf: yep, seems reasonable

MarkLee131: Done, pushed it.

PR #2018, opened 2026-04-25
HTTPSock: emit standard hardening response headers

### LINKED ISSUE #2012: HTTPSock: emit standard hardening response headers (XFO, nosniff, Cache-Control)
`PrintHeader` doesn't set `Cache-Control`, `X-Frame-Options`, `X-Content-Type-Options`, or `Referrer-Policy` for any response. The static file path adds `Cache-Control: public` (which is also wrong direction for authenticated content) but everything else goes out bare.

```cpp
// src/HTTPSock.cpp
716: bool CHTTPSock::PrintHeader(off_t uContentLength, const CString& sContentType,
                                 unsigned int uStatusId, const CString& sStatusMsg) {
...
734:     Write("HTTP/" + ... + "\r\n");
735:     Write("Date: " + GetDate() + "\r\n");
736:     Write("Server: " + CZNC::GetTag(false) + "\r\n");
737:     if (uContentLength > 0) Write("Content-Length: ...");
741:     Write("Content-Type: " + m_sContentType + "\r\n");
742:     for (Set-Cookie: ...) // SameSite=Strict; HttpOnly; Secure-when-SSL
749:     for (auto& it : m_msHeaders) ...
753:     Write("Connection: Close\r\n");
755:     Write("\r\n");
```

Concrete consequences:

- No `X-Frame-Options` / `frame-ancestors` → webadmin pages framable. CSRF is enforced server-side (`WebModules.cpp:661-669`), but the admin's browser still carries the cookie + token, so clickjacking that triggers a state-changing form submit will pass the token check.
- No `X-Content-Type-Options: nosniff` → endpoints that serve `text/plain` (`samplewebapi`, any module that does `WebSock.Write(text)` with `Content-Type: text/plain`) are MIME-sniffed by old browsers and can yield reflected XSS.
- No `Cache-Control` on authenticated HTML → previously rendered admin pages can be reread from browser history on a shared workstation.

## Fix

In `PrintHeader`, before the trailing `\r\n`:

```cpp
Write("X-Frame-Options: SAMEORIGIN\r\n");
Write("X-Content-Type-Options: nosniff\r\n");
Write("Referrer-Policy: same-origin\r\n");
if (uStatusId != 304 && !m_sContentType.StartsWith("image/") &&
    !m_sContentType.StartsWith("font/") && !m_sContentType.StartsWith("text/css")) {
    Write("Cache-Control: no-store, no-cache, must-revalidate, max-age=0\r\n");
    Write("Pragma: no-cache\r\n");
}
```

Also at line 378, change `Cache-Control: public` for static files to `Cache-Control: private, max-age=...` so shared HTTP caches don't fan out skin/avatar resources across users.

Happy to PR. Easy to test with `curl -i`.

MarkLee131:
Close #2012:
Add X-Frame-Options: SAMEORIGIN, X-Content-Type-Options: nosniff and Referrer-Policy: same-origin to every response so webadmin and module pages are framed/sniff-protected by default. Add no-store Cache-Control on dynamic responses so shared workstations can't replay authenticated pages from browser history. Static asset paths keep the ETag revalidation flow but switch from public to private so shared caches don't fan skin/avatar resources across users.

### DIFF
--- a/include/znc/HTTPSock.h
+++ b/include/znc/HTTPSock.h
@@ -56,6 +56,11 @@ class CHTTPSock : public CSocket {
      * inputs that could split the response into a separate header or body
      * (RFC 7230 disallows obs-fold; treat any bare CR/LF as invalid). */
     static bool IsValidHeaderField(const CString& s);
+    /** Suppress one of the default security/cache hardening headers for
+     * the next response. Use this when the caller does not want a value
+     * sent at all (a different value can instead be supplied via
+     * AddHeader). Must be called before PrintHeader. */
+    void OmitHardeningHeader(const CString& sName);
     void SetContentType(const CString& sContentType);
 
     bool PrintNotFound();
@@ -121,6 +126,11 @@ class CHTTPSock : public CSocket {
     void WriteFileUncompressed(CFile& File);
     void WriteFileGzipped(CFile& File);
 
+    /** Write the security/cache hardening header lines for the current
+     * response. Skips any name that the caller has already populated via
+     * AddHeader, or has explicitly omitted via OmitHardeningHeader. */
+    void WriteHardeningHeaders(unsigned int uStatusId);
+
   protected:
     void PrintPage(const CString& sPage);
     void Init();
@@ -142,6 +152,7 @@ class CHTTPSock : public CSocket {
     std::map<CString, VCString> m_msvsPOSTParams;
     std::map<CString, VCString> m_msvsGETParams;
     MCString m_msHeaders;
+    SCString m_ssOmitHardening;
     bool m_bHTTP10Client;
     CString m_sIfNoneMatch;
     bool m_bAcceptGzip;
--- a/src/HTTPSock.cpp
+++ b/src/HTTPSock.cpp
@@ -793,7 +793,7 @@ void CHTTPSock::WriteHardeningHeaders(unsigned int uStatusId) {
     // entirely via OmitHardeningHeader, before PrintHeader runs.
     writeIfWanted("X-Frame-Options", "SAMEORIGIN");
     writeIfWanted("X-Content-Type-Options", "nosniff");
-    writeIfWanted("Referrer-Policy", "same-origin");
+    writeIfWanted("Referrer-Policy", "no-referrer");
 
     // Don't cache authenticated/dynamic responses. Skip for 304 and for
     // static asset MIME types that the ETag/Last-Modified path handles
@@ -804,9 +804,7 @@ void CHTTPSock::WriteHardeningHeaders(unsigned int uStatusId) {
         m_sContentType.StartsWith("text/css") ||
         m_sContentType.StartsWith("application/javascript");
     if (!bStaticLike) {
-        writeIfWanted("Cache-Control",
-                      "no-store, no-cache, must-revalidate, max-age=0");
-        writeIfWanted("Pragma", "no-cache");
+        writeIfWanted("Cache-Control", "no-store");
     }
 }
 
--- a/src/HTTPSock.cpp
+++ b/src/HTTPSock.cpp
@@ -740,6 +740,8 @@ bool CHTTPSock::PrintHeader(off_t uContentLength, const CString& sContentType,
     }
     Write("Content-Type: " + m_sContentType + "\r\n");
 
+    WriteHardeningHeaders(uStatusId);
+
     for (const auto& it : m_msResponseCookies) {
         Write("Set-Cookie: " + it.first.Escape_n(CString::EURL) + "=" +
               it.second.Escape_n(CString::EURL) + "; HttpOnly; path=/;" +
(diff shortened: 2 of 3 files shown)


test/HTTPSockTest.cpp
```diff
@@ -0,0 +1,91 @@
...
+namespace {
+
+bool ContainsLineStartingWith(const VCString& vs, const CString& sPrefix) {
+    for (const CString& s : vs) {
+        if (s.StartsWith(sPrefix)) return true;
+    }
+    return false;
+}
+
+bool ContainsLine(const VCString& vs, const CString& sLine) {
```
DarthGandalf: gmock has such matchers already: Contains("foo"), Contains(StartsWith("bar"))

include/znc/HTTPSock.h
```diff
@@ -52,6 +52,13 @@ class CHTTPSock : public CSocket {
                      unsigned int uStatusId = 200,
                      const CString& sStatusMsg = "OK");
     void AddHeader(const CString& sName, const CString& sValue);
+    /** Build the security/cache hardening header lines a response should
+     * carry. Names already present in `msCallerHeaders` are skipped, so
+     * callers can opt out via AddHeader before PrintHeader is invoked.
+     * Each returned line is `Name: value` with no trailing CRLF. */
+    static VCString HardeningHeaders(unsigned int uStatusId,
```
DarthGandalf: I don't like API of this: the temporary vector.

Also this shouldn't really be exposed in header, as users aren't supposed to call this function. For test, can you construct the CHTTPSock object and collect via overridden Write() function how Write() was called? Then this method could be private and call Write directly instead of returning vector

include/znc/HTTPSock.h
```diff
@@ -52,6 +52,13 @@ class CHTTPSock : public CSocket {
                      unsigned int uStatusId = 200,
                      const CString& sStatusMsg = "OK");
     void AddHeader(const CString& sName, const CString& sValue);
+    /** Build the security/cache hardening header lines a response should
+     * carry. Names already present in `msCallerHeaders` are skipped, so
+     * callers can opt out via AddHeader before PrintHeader is invoked.
```
DarthGandalf: That doesn't work: if I want to remove one of hardening headers (for whatever reason), what to call AddHeader with?

test/HTTPSockTest.cpp
MarkLee131: Switched to Contains(StartsWith(...)) and dropped the hand-rolled ContainsLine/ContainsLineStartingWith helpers. Pushed in fe3e70bb.

include/znc/HTTPSock.h
MarkLee131: Done. The helper is now a private WriteHardeningHeaders that calls Write() directly from PrintHeader, no more temporary vector and nothing exposed in the public header. The tests construct a tiny CHTTPSock subclass with an overridden Write() that captures each call into a vector, then drive it through PrintHeader and match with gmock.

include/znc/HTTPSock.h
MarkLee131: Fair point, AddHeader had no way to actually drop one. Added OmitHardeningHeader(name) so callers can suppress a default outright (separate from overriding the value via AddHeader). Test for it in HardeningHeadersOmittedByCaller. 
Happy to switch to a sentinel like AddHeader(name, "") if you'd rather not grow the API, but that changes existing AddHeader semantics for empty values, so I went with the explicit method.

test/HTTPSockTest.cpp
```diff
@@ -37,3 +43,114 @@ TEST(HTTPSockTest, IsValidHeaderField) {
...
+        m_vsLines.push_back(sData);
+        return true;
+    }
+
+    VCString m_vsLines;
+};
+
+class HTTPSockHeadersTest : public ::testing::Test {
+  protected:
+    HTTPSockHeadersTest() { CZNC::CreateInstance(); }
```
DarthGandalf, line 73: What is this for?

MarkLee131: `PrintHeader` writes the `Server:` line via `CZNC::GetTag()`, which dereferences the singleton in `Get()`. Without `CreateInstance()` the test segfaults on the first `PrintHeader` call...
Same pattern as `BufferTest`/`UserTest`/`NetworkTest`.

src/HTTPSock.cpp
```diff
@@ -776,6 +778,38 @@ void CHTTPSock::AddHeader(const CString& sName, const CString& sValue) {
...
+        if (m_msHeaders.find(sName) != m_msHeaders.end()) return;
+        if (m_ssOmitHardening.find(sName) != m_ssOmitHardening.end()) return;
+        Write(sName + ": " + sValue + "\r\n");
+    };
+
+    // Always-on defaults: callers can override via AddHeader, or skip
+    // entirely via OmitHardeningHeader, before PrintHeader runs.
+    writeIfWanted("X-Frame-Options", "SAMEORIGIN");
+    writeIfWanted("X-Content-Type-Options", "nosniff");
+    writeIfWanted("Referrer-Policy", "same-origin");
```
DarthGandalf: Wouldn't `no-referrer` be better?

src/HTTPSock.cpp
```diff
@@ -776,6 +778,38 @@ void CHTTPSock::AddHeader(const CString& sName, const CString& sValue) {
...
+    // static asset MIME types that the ETag/Last-Modified path handles
+    // explicitly via PrintFile.
+    const bool bStaticLike =
+        uStatusId == 304 || m_sContentType.StartsWith("image/") ||
+        m_sContentType.StartsWith("font/") ||
+        m_sContentType.StartsWith("text/css") ||
+        m_sContentType.StartsWith("application/javascript");
+    if (!bStaticLike) {
+        writeIfWanted("Cache-Control",
+                      "no-store, no-cache, must-revalidate, max-age=0");
```
DarthGandalf: according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#preventing_storing no-store should be enough?

src/HTTPSock.cpp
```diff
@@ -776,6 +778,38 @@ void CHTTPSock::AddHeader(const CString& sName, const CString& sValue) {
...
+        writeIfWanted("Pragma", "no-cache");
```
DarthGandalf: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma says deprecated

src/HTTPSock.cpp
MarkLee131: Yeah, you're right. Switched to no-referrer in .

src/HTTPSock.cpp
MarkLee131: Yep, simplified to just Cache-Control: no-store in .

src/HTTPSock.cpp
MarkLee131: Dropped Pragma in .

PR #2021, opened 2026-05-25
fileutils: open Copy() destination at the source's mode

dxbjavid:
`CFile::Copy` opens the destination via `open(O_WRONLY|O_CREAT|O_TRUNC)` (default mode 0644 in `FileUtils.h`) and only `Chmod()`s it down to the source's mode after the write completes. When the source is 0600 (znc.conf backup in `BackupConfigOnce` at src/znc.cpp:986, module `.registry` copies at src/Modules.cpp:283 and src/User.cpp:1280), another local user can `open()` the destination O_RDONLY during the window and keep the fd across the final chmod, reading the credentials once writes finish.

Stat the source up front and pass its mode to `Open()` so the file is created restricted, falling back to 0600 if the stat fails. The trailing `Chmod()` is kept since `O_CREAT|O_TRUNC` doesn't update the mode of an existing destination in the overwrite path.

### DIFF
--- a/src/FileUtils.cpp
+++ b/src/FileUtils.cpp
@@ -265,7 +265,11 @@ bool CFile::Copy(const CString& sOldFileName, const CString& sNewFileName,
         iMode = st.st_mode & 07777;
     }
 
-    if (!NewFile.Open(O_WRONLY | O_CREAT | O_TRUNC, iMode)) {
+    // Force owner read+write while copying so the write still works for a
+    // source that lacks them (e.g. r-xr-xr-x); the trailing Chmod() puts
+    // the exact source mode back. This only ever adds owner bits, so the
+    // group/other bits stay as restrictive as the source the whole time.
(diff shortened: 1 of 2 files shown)


dxbjavid:
Thanks for catching this. I will add a regression test for the restricted-permission copy path so the new open-with-source-mode behavior is covered and the Codecov warning goes away.

DarthGandalf:
Does this work if the original file is `r-xr-xr-x`? Note missing `w` in owner's

dxbjavid:
Good catch. Opening it straight at the source mode does work for a fresh destination (the create keeps a writable fd even at r-xr-xr-x), but it's fragile: the overwrite path then can't reopen a destination that has no owner write, so the copy fails there. I've switched it to force owner read+write while copying and let the trailing chmod put the exact source mode back, which only ever adds owner bits so the group/other bits stay as restrictive as the source the whole time. Added a regression test covering the r-xr-xr-x case plus a plain 0600 one.

PR #2022, opened 2026-06-06
Enable SSL by default in ZNC configuration when connecting to servers

TehPeGaSuS:
Change the default value of `bSSL` (server) from `false` to `true` in `WriteNewConfig()`, so that `--makeconf` suggests connecting to servers over SSL/TLS out of the box.

Users connecting to modern IRC networks (Libera, OFTC, etc.) should be using SSL on port 6697. Defaulting to plaintext on 6667 is a poor security default for new installs. This brings `--makeconf` in line with what the Libera path already does (which hard-codes `bSSL = true` when the user picks that network).

Users on networks that don't support SSL can still opt out during the interactive prompt.

### DIFF
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -791,7 +791,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
         }
 
         CString sHost, sPass, sHint;
-        bool bSSL = false;
+        bool bSSL = true;
         unsigned int uServerPort = 0;
 
         if (sNetwork.Equals("libera")) {
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -1,4 +1,4 @@
-        /*
+/*
  * Copyright (C) 2004-2026 ZNC, see the NOTICE file for details.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/znc.cpp
+++ b/src/znc.cpp
@@ -1,4 +1,4 @@
-/*
+        /*
  * Copyright (C) 2004-2026 ZNC, see the NOTICE file for details.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -791,14 +791,15 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
         }
 
         CString sHost, sPass, sHint;
-        bool bSSL = true;
+        #ifdef HAVE_LIBSSL
+            bool bSSL = true;
+        #else
+            bool bSSL = false;
+        #endif
         unsigned int uServerPort = 0;
 
         if (sNetwork.Equals("libera")) {
             sHost = "irc.libera.chat";
-#ifdef HAVE_LIBSSL
-            bSSL = true;
-#endif
         } else {
             sHint = "host only";
         }
--- a/test/integration/framework/znctest.cpp
+++ b/test/integration/framework/znctest.cpp
@@ -47,7 +47,7 @@ void WriteConfig(QString path) {
     p.ReadUntil("Name [libera]");           p.Write("test");
     p.ReadUntil("Server host (host only)"); p.Write("unix:" + path.toUtf8() + "/inttest.ircd");
     p.ReadUntil("Server uses SSL?");        p.Write();
-    p.ReadUntil("6667");                    p.Write();
+    p.ReadUntil("6697");                    p.Write();
     p.ReadUntil("password");                p.Write();
     p.ReadUntil("channels");                p.Write();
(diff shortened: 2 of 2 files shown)


TehPeGaSuS:
During config:
```
$ ./znc --makeconf
[ ** ] 
[ ** ] -- Global settings --
[ ** ] 
[ ?? ] Listen on port (1025 to 65534): 1111
[ ?? ] Listen using SSL (yes/no) [no]: 
[ ?? ] Listen using both IPv4 and IPv6 (yes/no) [yes]: 
[ .. ] Verifying the listener...
[ ** ] Unable to locate pem file: [/home/icka/.znc/znc.pem], creating it
[ .. ] Writing Pem file [/home/icka/.znc/znc.pem]...
[ ** ] Enabled global modules [corecaps, saslplainauth, webadmin]
[ ** ] 
[ ** ] -- Admin user settings --
[ ** ] 
[ ?? ] Username (alphanumeric): Test
[ ?? ] Enter password: 
[ ?? ] Confirm password: 
[ ?? ] Nick [Test]: 
[ ?? ] Alternate nick [Test_]: 
[ ?? ] Ident [Test]: 
[ ?? ] Real name (optional): 
[ ?? ] Bind host (optional): 
[ ** ] Enabled user modules [chansaver, controlpanel]
[ ** ] 
[ ?? ] Set up a network? (yes/no) [yes]: 
[ ** ] 
[ ** ] -- Network settings --
[ ** ] 
[ ?? ] Name [libera]: ptirc
[ ?? ] Server host (host only): irc.ptirc.org
[ ?? ] Server uses SSL? (yes/no) [yes]: 
[ ?? ] Server port (1 to 65535) [6697]: 
[ ?? ] Server password (probably empty): 
[ ?? ] Initial channels: #test
[ ** ] Enabled network modules [simple_away]
[ ** ] 
[ .. ] Writing config [/home/icka/.znc/configs/znc.conf]...
[ ** ] 
[ ** ] To connect to this ZNC you need to connect to it as your IRC server
[ ** ] using the port that you supplied.  You have to supply your login info
[ ** ] as the IRC server password like this: user/network:pass.
[ ** ] 
[ ** ] Try something like this in your IRC client...
[ ** ] /server <znc_server_ip> 1111 Test/ptirc:<pass>
[ ** ] 
[ ** ] To manage settings, users and networks, point your web browser to
[ ** ] http://<znc_server_ip>:1111/
[ ** ] 
[ ?? ] Launch ZNC now? (yes/no) [yes]: no
```
Inside znc.conf:
```
$ cat ~/.znc/configs/znc.conf 
// WARNING
//
// Do NOT edit this file while ZNC is running!
// Use webadmin or *controlpanel instead.
//
// Altering this file by hand will forfeit all support.
//
// But if you feel risky, you might want to read help on /znc saveconfig and /znc rehash.
// Also check https://wiki.znc.in/Configuration

Version = 1.11.x
<Listener l>
	Port = 1111
	IPv4 = true
	IPv6 = true
	SSL = false
</Listener>
LoadModule = corecaps
LoadModule = saslplainauth
LoadModule = webadmin

<User Test>
<Pass password>
	Method = Argon2id
	Hash = $argon2id$v=19$m=6144,t=6,p=1$T1FpNjlQMW1OSW5rR2l1bjNVTz8$KRM3S7kT7FFSddu2/jDLnIENOXbTGsf4p6Mrn00RGPI
</Pass>
	Admin      = true
	Nick       = Test
	AltNick    = Test_
	Ident      = Test
	LoadModule = chansaver
	LoadModule = controlpanel

	<Network ptirc>
		LoadModule = simple_away
		Server     = irc.ptirc.org +6697 

		<Chan #test>
		</Chan>
	</Network>
</User>
```

src/znc.cpp
DarthGandalf: This is noop now

src/znc.cpp
```diff
@@ -791,7 +791,7 @@ bool CZNC::WriteNewConfig(const CString& sConfigFile) {
         }
 
         CString sHost, sPass, sHint;
-        bool bSSL = false;
+        bool bSSL = true;
```
DarthGandalf: Please try to compile ZNC without libssl support and see what happens if you apply this patch.

Maybe it would be useful to have libssl-less setup in CI, but I didn't bother with it yet

DarthGandalf (changes requested):

src/znc.cpp
```diff
@@ -1,4 +1,4 @@
-/*
```
DarthGandalf, line 1: what happened here?

TehPeGaSuS: Fixed in the latest commit. I don't know how that slipped in

PR #2024, opened 2026-06-16
Fix build error with OpenSSL 4

raviksharma:
Fixes the issues with build error

More details here https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138356

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -129,7 +129,8 @@ void CUtils::GenerateCert(FILE* pOut) {
     sEmailAddr += "@";
     sEmailAddr += sHostName;
 
-    X509_NAME* pName = X509_get_subject_name(pCert.get());
+    X509_NAME* pName = X509_NAME_new();
+    if (!pName) return;
     X509_NAME_add_entry_by_txt(pName, "OU", MBSTRING_ASC,
                                (unsigned char*)pLogName, -1, -1, 0);
     X509_NAME_add_entry_by_txt(pName, "CN", MBSTRING_ASC,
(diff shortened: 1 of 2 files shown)


src/Utils.cpp
```diff
@@ -129,15 +129,18 @@ void CUtils::GenerateCert(FILE* pOut) {
...
+    X509_NAME* pName = X509_NAME_new();
+    if (!pName) return;
     X509_NAME_add_entry_by_txt(pName, "OU", MBSTRING_ASC,
                                (unsigned char*)pLogName, -1, -1, 0);
     X509_NAME_add_entry_by_txt(pName, "CN", MBSTRING_ASC,
                                (unsigned char*)sHostName.c_str(), -1, -1, 0);
     X509_NAME_add_entry_by_txt(pName, "emailAddress", MBSTRING_ASC,
                                (unsigned char*)sEmailAddr.c_str(), -1, -1, 0);
 
+    X509_set_subject_name(pCert.get(), pName);
```
psychon, line 141: Drive-by-comment: This change seems unrelated (or at least I do not understand it). Why is this new function call required to fix a build error?

(Also: The commit message does not say anything about the build error. The PR description is slightly better by linking to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138356 but that issue also "does not say much".)

DarthGandalf: When merging it's easy enough to select "squash" option which takes the PR description, so this is not a big deal

DarthGandalf (commented):
Does this change bump the minimum supported openssl version?

src/Utils.cpp
raviksharma: Previously, X509_get_subject_name(pCert.get()) was used to set the values directly, now X509_NAME_new is used to create the pName and X509_set_subject_name is required to set the pName in pCert.

raviksharma:
> Does this change bump the minimum supported openssl version?

The minimum version required is 1.1.1 https://docs.openssl.org/1.1.1/man3/X509_dup/

PR #2026, opened 2026-06-30
Fix build with LibreSSL >= 2.7.0 by using native ASN1_STRING_get0_data

fred0r:
LibreSSL >= 2.7.0 has its own ASN1_STRING_get0_data() that accepts const ASN1_STRING*, unlike the ASN1_STRING_data() fallback used in the #else branch. The const-qualifier loss error was introduced by commit 94bcf919 which made common_name_asn1 const for OpenSSL 4 compatibility.


DarthGandalf:
How old is 2.7.0?

Does #2025 fixes it for you?

fred0r:
yes - that fixes it for me on FreeBSD and Linux.
will close this PR.

Issue #2035, opened 2026-08-17
schat - Cannot chat with self-signed certs.


RealKindOne:
Git bisect shows commit 4442af724f061a5b9baf7e80c644f41b027a65a2 - "Validate TLS certificate of IRC server." caused this regression.

How should znc handle this?

znc -D log from commit 4442af724f061a5b9baf7e80c644f41b027a65a2:
```
[2026-08-16 03:34:23.215107] TDNS: SCHAT::(s)KindOne_, connecting to [10.10.1.2] using bindhost [10.10.1.2]
[2026-08-16 03:34:23.265110] (KindTwo/EFNet) ZNC -> CLI [:(s)KindOne_!(s)KindOne_@10.10.1.2 PRIVMSG KindTwo :*** Connected.]
[2026-08-16 03:34:23.282111] SCHAT::(s)KindTwo: No cert
[2026-08-16 03:34:23.282111] SCHAT::(s)KindTwo == SockError(Anonymous SSL cert is not allowed, Unknown error 12569)
[2026-08-16 03:34:23.282111] (KindOne/EFnet) ZNC -> CLI [:(s)KindTwo!(s)KindTwo@10.10.1.2 PRIVMSG KindOne_ :*** Connected.]
[2026-08-16 03:34:23.282111] (KindOne/EFnet) ZNC -> CLI [:(s)KindTwo!(s)KindTwo@10.10.1.2 PRIVMSG KindOne_ :*** Disconnected.]
[2026-08-16 03:34:23.293112] SSLVerifyHost: checking 10.10.1.2
[2026-08-16 03:34:23.293112] SSLVerifyHost: Found CN Example
[2026-08-16 03:34:23.293112] SSLVerifyHost: host doesn't match
[2026-08-16 03:34:23.293112] SCHAT::(s)KindOne_: Bad cert
[2026-08-16 03:34:23.293112] SCHAT::(s)KindOne_ == SockError(Invalid SSL certificate: hostname doesn't match, self signed certificate, Unknown error 12569)
[2026-08-16 03:34:23.293112] (KindTwo/EFNet) ZNC -> CLI [:(s)KindOne_!(s)KindOne_@10.10.1.2 PRIVMSG KindTwo :*** Disconnected.]
```

---

znc -D log from a commit before "Validate TLS certificate of IRC server." without the issue:
```
[2026-08-16 03:01:44.758090] (KindOne/EFnet) CLI -> ZNC [PRIVMSG *schat :chat KindTwo]
[...]

Issue #815, opened 2015-01-14
awaystore replay does not work


bullfinsh:
Hi! 

13:26:51 <bullfinch> messages
13:26:51 -_awaystore(znc@znc.in)- Welcome Back!
13:26:51 -_awaystore(znc@znc.in)- You have 1 messages!
13:26:52 <bullfinch> show
13:26:52 <_awaystore> bullfinsh!~yeah@booomshakkalakka
13:26:52 <_awaystore>     0) [2015-01-14 10:47:16] pai
13:26:52 <*awaystore> #--- End Messages
13:26:55 <bullfinch> replay

the last command opens a new window with the following content:
13:26:55 -!- Irssi: Starting query in 1272 with 
13:26:55 <> 

My guess is that the new window should be from nick bullfinsh and so that I could just respond directly in that window. 

It would help my setup a whole lot. :)

Oh. 
Im running last git version; ZNC - 1.5.0-git-537-230a39f - http://znc.in
##

bullfinsh:
if i write haba:baba:saba to the awaylog and replay it I get the following message from Mr. baba

new window that pops up:
13:57:15 -!- Irssi: Starting query in 1272 with baba
13:57:15 <baba> saba


output from show in *awaystore:
 [2015-01-14 13:54:04] haba:baba:saba


	void ReplayCommand(const CString& sCommand) {
		CString nick = GetClient()->GetNick();
		for (u_int a = 0; a < m_vMessages.size(); a++) {
			CString sWhom = m_vMessages[a].Token(1, false, ":");
			CString sMessage = m_vMessages[a].Token(2, true, ":");
			PutUser(":" + sWhom + " PRIVMSG " + nick + " :" + sMessage);
		}
	}


I think that the bug might be in the way PutUser creates the stings.

PR #2038, opened 2026-08-23
Modernize various modules with newer On...Message events.

RealKindOne:
This pull request makes various modules use newer ```On...Message()``` events.

Crypt module required a small fix for the test to pass.

Not updated due to issues:
autocycle - Issue with OnKickMessage() if the op kicks themself with only me in channel? -- need to look into that more.
awaystore - #815 - And other issues
schat - #2035

### DIFF
--- a/modules/autoattach.cpp
+++ b/modules/autoattach.cpp
@@ -209,20 +209,24 @@ class CChanAttach : public CModule {
         }
     }
 
-    EModRet OnChanNotice(CNick& Nick, CChan& Channel,
-                         CString& sMessage) override {
-        TryAttach(Nick, Channel, sMessage);
+    EModRet OnChanNoticeMessage(CNoticeMessage& Message) override {
+        CString sText = Message.GetText();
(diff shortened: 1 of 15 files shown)


modules/autoattach.cpp
```diff
@@ -209,20 +209,24 @@ class CChanAttach : public CModule {
         }
     }
 
-    EModRet OnChanNotice(CNick& Nick, CChan& Channel,
-                         CString& sMessage) override {
-        TryAttach(Nick, Channel, sMessage);
+    EModRet OnChanNoticeMessage(CNoticeMessage& Message) override {
+        CString sText = Message.GetText();
+        TryAttach(Message.GetNick(), *Message.GetChan(), sText);
+        Message.SetText(sText);
```
DarthGandalf: not necessary: TryAttach doesn't change the message. Not sure why it accepts & instead of const&

RealKindOne: Fixed.

PR #2044, opened 2026-09-22
Fix null socket handling in CSockManager::Connect

pagzlol:
Companion to #2043, same pattern in the TCP connect path.

`CSockManager::Connect()` declares `CZNCSock* pcSock = nullptr` as a default argument (`include/znc/Socket.h:160`), but the body cannot actually handle null:

```cpp
m_InFlightDnsSockets[pcSock] = false;
if (pcSock) {
    pcSock->SetHostToVerifySSL(sHostname);
}
```

The null check guards exactly one line. `pcSock` is inserted into `m_InFlightDnsSockets` as a map key *before* it is checked, and immediately afterwards it is passed on unconditionally — to `task->pcSock` with threaded DNS, or straight to `FinishConnect()` without — both of which use it. So a null argument inserts a null key into the in-flight map and then crashes a few lines later anyway.

This resolves it the same way `CSocketManager::Listen()` in Csocket already does:

```cpp
if( !pcSock )
    pcSock = GetSockObj( "", 0 );
```

As with #2043, all in-tree callers pass a socket, so ZNC does not hit this today — the issue is that the signature offers a default that is not usable.


DarthGandalf:
> So a null argument inserts a null key into the in-flight map and then crashes a few lines later anyway.

Have you reproduced the crash?

I agree that this code looks strange and should be fixed, but I don't see a problem passing task->pcSock (which is nullptr) to FinishConnect

pagzlol:
No I haven't reproduced it that line was wrong and I should have traced the call before hand.

youre right about `FinishConnect`. a null there just gets a socket created for it and no crash on that path

There is still a null dereference on the error path rather than the o In `SetTDNSThreadFinished()`, if DNS resolution throws

```cpp
} catch (const CString& s) {
    DEBUG(task->sSockName << ", dns resolving error: " << s);
    task->pcSock->SetSockName(task->sSockName);
    task->pcSock->SockError(-1, s);
    delete task->pcSock;
}
```

`task->pcSock` is used with no null check, so with `HAVE_THREADED_DNS` and an unresolvable hostname a null `pcSock` would crash there,  have traced this rather than reproduced it.

happy to go whichever way you prefer, keep as is or drop it and remove the default from the signature

DarthGandalf:
Don't worry too much about this code - I have a WIP refactor of sockets and DNS, but that's quite slow, for several years already.

For short term, probably this can be merged as is, I'll think some more about it

Thanks for PR

src/Socket.cpp
```diff
@@ -396,10 +396,11 @@ CSockManager::~CSockManager() {}
 void CSockManager::Connect(const CString& sHostname, u_short iPort,
                            const CString& sSockName, int iTimeout, bool bSSL,
                            const CString& sBindHost, CZNCSock* pcSock) {
-    m_InFlightDnsSockets[pcSock] = false;
-    if (pcSock) {
-        pcSock->SetHostToVerifySSL(sHostname);
+    if (!pcSock) {
+        pcSock = GetSockObj("", 0);
```
DarthGandalf, line 400: shouldn't this be sHostname, iPort instead?

pagzlol:
Good catch and fixed

